PR #45, opened 2021-01-26
Add front display

linknum23:
Added driver for AmpliPi OLED display. Uses a queue for handling multiple sources of display update events. Display updates are currently added to the queue from two sources:
- Thread `__gather_stats()` that pulls system stats (IP, CPU %, CPU temp, RAM usage, uptime) once per second
- Function `set_volumes()` that can be called to immediately set the levels of the 6x volume bars

Please review thread-safety: `__gather_stats()` and `set_volumes()` both add to the queue and can happen asynchronous to each other. Not sure if additional protection is needed here.

### DIFF
--- a/amplipi/oled.py
+++ b/amplipi/oled.py
@@ -22,8 +22,8 @@ from luma.oled.device import ssd1309
 from PIL import ImageFont
 
 # +--------------------------------------------------+
-# ¦               AmpliPI OLED class                 ¦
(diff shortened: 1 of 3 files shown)


linknum23:
Reminder: current test setup with display makes volume update lag

linknum23:
Closing this. @Lohrer will give us a new PR with the fancy new color display.

PR #57, opened 2021-02-01
Fresh start

linknum23:
We are trying to document and make our development process a little easier here. The goal is that outsiders that want to play with AmpliPi's code will have the tools available to get started with what they want to work on.

### DIFF
--- a/config/update_config.bash
+++ b/config/update_config.bash
@@ -3,6 +3,8 @@
 # Update AmpliPi's configuration on a raspberry pi.
 # This script should install and configure everything necessary
 
+# TODO: many of the checks here depend on apt, if something is manually installed this script will not detect it yet. Please fix.
+
 # get directory that the script exists in
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
--- a/config/update_config.bash
+++ b/config/update_config.bash
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Update AmpliPi's configuration on a raspberry pi.
+# This script should install and configure everything necessary
+
 # get directory that the script exists in
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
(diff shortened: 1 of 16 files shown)


linknum23:
@cornercase I tried to document how to develop with amplipi from an outsiders perspective in [Developing.md](https://github.com/micro-nova/AmpliPi/blob/fresh_start/DEVELOPING.md). Does this help you? Please let me know if there is anything unclear or something doesn't do what you expected.

cornercase:
That looks great, I’ll give the setup a shot on my Pi 3 and let you know
how it goes.

On February 1, 2021 at 2:24:27 PM, Lincoln Lorenz (notifications@github.com)
wrote:

@cornercase <https://github.com/cornercase> I tried to document how to
develop with amplipi from an outsiders perspective in Developing.md
<https://github.com/micro-nova/AmpliPi/blob/fresh_start/DEVELOPING.md>.
Does this help you? Please let me know if there is anything unclear or
something doesn't do what you expected.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/micro-nova/AmpliPi/pull/57#issuecomment-771133695>, or
unsubscribe

linknum23:
Re-based on top of master to update to the latest documentation.

@micronova-jb can you add something that installs the shairport-metadata-reader if it doesn't exist? I think that's the last thing missing from bringing this completely up to date with all of our stream integrations. It should probably be based on our fork since we expect to make changes in the future.

config/update_config.bash
```diff
@@ -29,6 +29,27 @@ if [ 0 -eq "${sp_installed}" ]; then
   sudo systemctl disable shairport-sync.service
 fi
 
+# rough configuration of shairport-sync-metadata-reader
+git_installed=$(sudo apt list --installed 2> /dev/null | grep git/ -c)
```
linknum23, line 33: why is there a '/' here?

linknum23: This should also check for the package **build-essential** that checks for make and gcc and it looks like it also needs the package **autoconf**

linknum23 (commented):
After adding those packages, check that this still works on a fresh install.

config/update_config.bash
micronova-jb: Searching for 'git/' actually provides the info that we want - searching for 'git' without the slash ends up showing nine results vs. the one for git/. I'll add checks for build-essential and autoconf

micronova-jb: Should stm32flash be added to update_config.bash? It isn't necessary for normal operation, but it is required for flashing the preamp firmware.

linknum23:
:+1: Jeremy I think you can approve this now and merge it in!

PR #61, opened 2021-02-11
Updates to all scripts involved in Shairport Metadata

micronova-jb:
-streams.py: Pulls new data points out of currentSong, checks whether the picture is empty or not, then forwards the picture directory/name
-shairport_metadata.bash: cleans out photo repositories from previous instances, runs the metadata reader from the image directory so picture data is properly saved in the right place
sp_meta.py: Checks for the name and size of the image - this info gets saved in currentSong
script.js: Now makes use of the shairport cover art

### DIFF
--- a/config/shairport_metadata.bash
+++ b/config/shairport_metadata.bash
@@ -1,3 +1,8 @@
 #!/bin/bash
+# Clear out any previous album cover images, then navigate to the proper directory #
+rm -r -f /home/pi/web/static/imgs/srcs/$1/
+mkdir -p /home/pi/web/static/imgs/srcs/$1/
+cd /home/pi/web/static/imgs/srcs/$1/
+
 # Start the metadata service with argument $1 being the source number #
 cat /home/pi/config/srcs/$1/shairport-sync-metadata | shairport-sync-metadata-reader | /home/pi/scripts/sp_meta.py $1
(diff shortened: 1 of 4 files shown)


config/shairport_metadata.bash
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
+# Clear out any previous album cover images, then navigate to the proper directory #
```
linknum23, line 2: add a guard here before the rm -rf to validate the $1 arg, maybe something like this:
```
src="$1"
iif ((src <= 0 || src > 4)); then
  exit 1;
fi

Comment (linknum23) [reply config/shairport_metadata.bash:2]: This way we cant delete random stuff even if this script is called in a weird way

File: scripts/sp_meta.py [hunk 2]
```diff
@@ -89,5 +89,15 @@ def s_info(inp):
         f = open(cs_loc, 'w')
         f.write(data)
         f.close()
+    elif field == 'Image length':
```
linknum23, line 92: Add note about our fork?

linknum23 (changes requested):

linknum23 (approved):

linknum23:
@micronova-jb can you rebase this on master?

PR #74, opened 2021-02-23
Add preset interface to automate system configuration

linknum23:
Adds the following endpoints
- POST presets/#/load - Loads a preset, returns state after preset load completes
- GET presets - Get the list of available presets
- POST preset - Add a preset. Preset syntax is name, id, commands, and state where state has the same syntax as config file.
- PATCH presets/# - Update a preset


amplipi/ctrl.py
```diff
@@ -608,8 +625,168 @@ def get_stations(self, id, stream_index=None):
...
+    """
+    To avoid any issues with audio coming out of the wrong speakers, we will need to carefully load a preset configuration. Below is an idea of how a preset configuration could be loaded to avoid any weirdness. We are also considering adding a "Last config" preset that allows us to easily revert the configuration changes.
+
+    1. Grab system modification mutex to avoid accidental changes (requests during this time return some error)
+    1. Save current configuration as "Last config" preset
+    1. Mute any effected zones
+    1. Execute changes to source, zone, group each in increasing order
+    1. Unmute effected zones that were not muted
+    1. Execute any stream commands
+    1. Release system mutex, future requests are successful after this
```
micronova-jb, line 715: Is this supposed to be a list of steps? Why are they all '1.' ?

docs/PRESETS.md
```diff
@@ -0,0 +1,25 @@
...
+1. Grab system modification mutex to avoid accidental changes (requests during this time return some error)
+1. Save current configuration as "Last config" preset
+1. Mute any effected zones
+1. Execute any stream changes (configuration then commands)
+1. Execute changes source by source in increasing order
+1. Execute change zone by zone in increasing order
+1. Execute changes group by group in increasing order
+1. Unmute effected zones that were not muted
+1. Force web client refresh to fixup website
+1. Release system mutex, future requests are successful after this
```
micronova-jb, line 16: Same question here

docs/amplipi_api.yaml
```diff
@@ -947,4 +1037,56 @@ components:
...
+        stream:
+          type: integer
+          description: stream id
+      required:
+        - cmd
+        - stream
+    Preset:
+      title: Preset
+      type: object
+      description: A partial controller configuration that can be loaded on demand. In addition to most of the configurration found in Status, this can contain commands as well that configure the state of different streaming services. At the moment configuration of streams is only supported as commands.
```
micronova-jb, line 1066: Spelling

micronova-jb (approved):
Made a few comments, but other than that it looks really good! I look forward to expanding on the pytest stuff.

amplipi/ctrl.py
linknum23: These were copied from a markdown document, should probably number them so they are easier to read

linknum23:
I still have to fix this mute bug I found, yesterday. @micronova-jb could you play with presets using the open api site to test each of the examples on a live amplipi? I want to make sure the documentation both makes sense and is consistent ( and the commands work! ).

PR #81, opened 2021-03-08
Add internetradio stream type using VLC

kjk2010:
Adds a new stream type of 'internetradio' that utilizes VLC to stream any internet radio URL. AmpliPi launches config/vlc.py to start the stream, and vlc.py also keeps track of metadata changes.


linknum23:
This looks great! We are testing it on our setup right now. I had to fix a couple things with the vlc params but everything is looking good. The KEXP station does not appear to support the current playing song, was that your experience?

kjk2010:
On my Raspberry Pi, I get this for the current song for KEXP: 'KEXP.ORG 90.3FM - where the music matters'. For Artist, it's 'No Artist'. If it's not showing up on your test, I wonder if there is an issue with the conversion of the metadata. The metadata comes in from the stream as 'title', then gets saved in the currentSong file as 'song'.
Here is another internet radio source that does have artist and song: http://stream.revma.ihrhls.com/zc5247/hls.m3u8

linknum23:
Ooops sorry, I meant artist. I updated my comment above. The radio station name comes through perfectly.

config/vlc.py
```diff
@@ -0,0 +1,5660 @@
+#! /usr/bin/python
```
linknum23, line 1: It looks like this library can be installed with pip via:  `pip install python-vlc`. Could we install that and just add the modified lines under the `if __name__ == '__main__':` to a simpler script? This would make it a lot easier to keep up to date with the latest vlc version.

linknum23 (changes requested):
This is looking great. We've been playing internet radio all morning over here with this new stream type.
Check out my comments and let us know what you think. We should be able to merge this in real soon.

linknum23:
> On my Raspberry Pi, I get this for the current song for KEXP: 'KEXP.ORG 90.3FM - where the music matters'. For Artist, it's 'No Artist'. If it's not showing up on your test, I wonder if there is an issue with the conversion of the metadata. The metadata comes in from the stream as 'title', then gets saved in the currentSong file as 'song'.
> Here is another internet radio source that does have artist and song: http://stream.revma.ihrhls.com/zc5247/hls.m3u8

I tested this radio station and the artist and song name worked as expected.

linknum23:
@kjk2010 let me know if you would like me to make the changes requested

amplipi/streams.py
```diff
@@ -522,3 +529,122 @@ def __str__(self):
...
+
+    if self.mock:
+      print('{} connected to {}'.format(self.name, src))
+      self.state = 'connected'
+      self.src = src
+      return
+
+    # Start audio via VLC.py
+    inetradio_args = ['python3 /home/pi/config/vlc.py {} {} {}'.format(self.url, src, utils.output_device(src))]
+    self.proc = subprocess.Popen(args=inetradio_args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, preexec_fn=os.setpgrp)
```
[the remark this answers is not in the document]
kjk2010: I had issues with Python finding vlc or it's components when running shell=False. I could have been missing a setting. Running it without a shell would make things easier, as it wouldn't need to kill it's process via PGID.

config/vlc.py
kjk2010: That sounds like a very reasonable option.

kjk2010:
@linknum23 feel free to make the changes

kjk2010:
Thanks for the changes and suggestions you made. I cleaned up vlc.py as you suggested, but had to rename it to runvlc.py as the file name was conflicting with the 'vlc' module. I also updated streams.py so that it doesn't launch runvlc.py in a shell.

linknum23 (approved):

linknum23 (approved):

PR #88, opened 2021-03-14
Local fallback on CDN failure for js and css

kmhallen:
I couldn't find a solution to Google fonts in local style.css, but everything else has a local fallback.

References:
- CSS
  - https://stackoverflow.com/a/52223674/15394192
- jQuerry
  - https://stackoverflow.com/a/5531821
- jQuerry UI
  - https://stackoverflow.com/a/5014094
- Bootstrap
  - https://stackoverflow.com/a/25759833


linknum23:
Awesome. I'll add in the Google fonts soon and we'll call this good. Thanks @kmhallen!

PR #93, opened 2021-03-18
Change shairport album art handling

micronova-jb:
Album cover art from shairport has been moved from 'web/static' to 'web/generated' which makes more sense. This also paves the way for a future change - these images should be stored on some partition outside of the eMMC so that we aren't dipping too greedily into the number of read/write cycles


linknum23:
I'm waiting to merge the restructuring before adding this.

linknum23 (approved):

PR #97, opened 2021-03-29
Add RAM disks

### LINKED ISSUE #76: Use a RAM disk for volatile configuration
AmpliPi uses an mmc on board the Pi Cm3+. To reduce the wear on the mmc, we should configure any volatile writes to use a RAM disk. Almost all of the stream config should be moved to this.
- [x] RAM directory creation at startup
- [x] Add pertinent config and metadata files to RAM
- [x] Figure out how much space is needed for the partition (100 MB or smaller)

micronova-jb:
Read/write cycles saved on the SD card
DLNA metadata switched to proper directory (now part of RAM)

This will need more testing once other branches are merged. The most important part of this PR is 'use_ram.py'
Closes #76

### DIFF
--- a/config/use_ram.py
+++ b/config/use_ram.py
@@ -12,8 +12,8 @@
 
 import os
 
-web_directory = '/home/pi/web/generated'
-src_directory = '/home/pi/config/srcs'
+web_directory = '/home/pi/amplipi-dev/web/generated'
+src_directory = '/home/pi/amplipi-dev/config/srcs'
 
 os.system('mkdir -p {}'.format(web_directory))
 os.system('sudo mount -osize=100m tmpfs -t tmpfs {}'.format(web_directory))


amplipi/streams.py
```diff
@@ -511,7 +511,7 @@ def disconnect(self):
     self.src = None
 
   def info(self):
-    loc = '/home/pi/config/dlna/{}/currentSong'.format(self.src)
+    loc = '/home/pi/config/srcs/{}/currentSong'.format(self.src)
```
linknum23, line 514: Can you rebase this PR on some of the recent restructuring to get utils.get_dir('config') and use that to remove the absolute paths?

micronova-jb:
-This needs to be updated: this needs to end up in the proper place before a restart/run
-Folders need to be updated for the amplipi-dev change

micronova-jb:
The branch is basically the same as "master" as of June 28th, 2021. Leave this branch open for the final changes:
- [ ] Figure out the 'final' directory structure of config/srcs and web/generated
- [ ] Integrate this python script into startup (involves copying to the right spot and linking to init.d. Reboot required after this)

PR #98, opened 2021-04-02
Fix internetradio metadata

kjk2010:
Some stations use 'nowplaying' metadata instead of the separate 'artist' and 'title' metadata. This adds a check for the 'nowplaying' metadata and splits it into 'artist' and 'title' for AmpliPi.

### DIFF
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -622,7 +622,7 @@ class InternetRadio:
         d['artist'] = data['artist']
         d['song'] = data['song']
         d['img_url'] = self.logo
-        d['album'] = ""
+        d['station'] = data['station']
 
         return(d)
     except Exception:
--- a/docs/amplipi_api.yaml
+++ b/docs/amplipi_api.yaml
@@ -558,6 +558,18 @@ paths:
                   type: internetradio
                   url: http://ice2.somafm.com/groovesalad-16-aac
                   logo: https://somafm.com/img3/groovesalad-200.jpg
+              Add Beatles Internet Radio Station:
+                value:
+                  name: Beatles Radio
+                  type: internetradio
+                  url: http://www.beatlesradio.com:8000/stream/1/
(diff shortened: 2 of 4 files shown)


linknum23:
Awesome. Just testing this right now. I am getting the artist and track info now on SomaFM radio stations like Groove Salad. Thanks for figuring that out! Were there any other notable radio stations that you tested this with? I can add them as examples to the Create New Stream documentation to make them easy to test with

kjk2010:
I tested with:
      "name": "Classical KING FM 98.1",
      "url": "http://classicalking.streamguys1.com/king-fm-aac-iheart",
      "logo": "https://i.iheart.com/v3/re/assets/images/7bcfd87a-de3e-47d0-b896-be0ed38c9d74.png"

and

      "name": "Beatles Radio",
      "url": "http://www.beatlesradio.com:8000/stream/1/",
      "logo": "http://www.beatlesradio.com/content/images/thumbs/0000587.gif"

linknum23:
What do you think about showing the radio station title/tagline underneath the artist and song title? It seems to be passed in Metadata.Title field when the Metadata.NowPlaying is also populated. I'm trying to figure out how useful it is.

kjk2010:
I would expect most stations to include their name in their logo, which would make displaying the station name from the 'title' metadata a little redundant, but it could easily be passed into the 'Album' field on the GUI.

linknum23:
I'm still working on updates today, but I will integrate this in the next couple of days. I agree with you on the icon being redundant. maybe it makes sense to only show the station info only if the station icon is not specified/available.

kjk2010:
I've updated runvlc.py to pass the station title in the album field, in case it's ever useful to show on the front end.

linknum23:
I cleaned up the code a little bit and added a verbose flag for debugging in the future. I also changed the name from 'album' to 'station' since that's what it really is.
Merging this in. Thanks again!

linknum23:
@kjk2010 Check your email. You should have gotten something from micro-nova.com.

PR #112, opened 2021-04-27
Change editorconfig to cover every file by default

Lohrer:
The only exception I know of now that is important is Markdown's end-of-line spaces.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,24 +3,18 @@
 # top-most EditorConfig file
 root = true
 
-# newline ending every file
+# Every file
 [*]
-insert_final_newline = true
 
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -3,24 +3,22 @@
...
 indent_size = 2
+indent_style = space
+
+# Ensure a newline is at the end of every file
+insert_final_newline = true
+
+# Remove any extraneous whitespace
+trim_trailing_whitespace = true
+
+# Trailing whitespace in markdown is meaningful
```
linknum23, line 22: Can we remove them anyway?

linknum23 (changes requested):

PR #113, opened 2021-04-27
Add preamp functionality for new boards

micronova-jb:
FW update to include new power board features. Will update more as all of the boards are replaced


micronova-jb:
Need to look into timeouts on the I2C loops - things like "while(I2C_GetFlagStatus(I2C1, I2C_FLAG_ADDR) == RESET);"

We should probably break out of these after a period of time

micronova-jb:
Also, once things are finalized with version number generation, be sure to build as Release and fully test the release FW

micronova-jb:
Use the following command to get the seven nibble hash for a commit. Dirty can be whatever you choose, and it shows up at the end of a hash string if there are pending changes
`git describe --always --match "not a tag" --dirty="*"`

There also may be some way to finesse `git diff --exit-code` to more smoothly automate the process. This seemingly returns nothing on a clean branch while returning the changes on a dirty branch. Between these two commands, some sort of automation for the preamp firmware can be discussed

fw/preamp/src/front_panel.c
```diff
@@ -25,6 +25,7 @@
 
 bool audio_power_on = false;
 
+// Simple counter used for time-sensitive operations
 void delay(int a) {
```
Lohrer, line 29: In general this is not a good idea if a consistent delay is needed. The time needed to execute this can vary based on flash read times, pipelining, and probably other things I don't know about in this architecture. This is simple enough it's _probably_ fairly consistent, but I'd have to look at the generated assembly and then see what the M0 architecture is doing. Adding instruction barriers might make things more consistent too. BUT, why dig into all of that when there's a timer designed for _exactly_ this function! Ok fine, it's designed for real-time systems like FreeRTOS but it's a very simple timer that'd be great for just "delay x us/ms/s/whatever"

Lohrer (commented):
I assume you don't see my comments until I "finish" my review. I'm not done but wanted you to see comments as I go.

fw/preamp/src/front_panel.c
micronova-jb: What is this magical timer?

fw/preamp/src/main.c
```diff
@@ -134,6 +127,30 @@ void init_i2c(int preamp_addr)
 	//I2C_InitStructure1.I2C_Timing = 0x10805E89; // From 8-Mhz HSI clock with 100ns rise, 10ns fall -> 100 KHz
```
Lohrer, line 127: Where did this number come from and why is it set for only one of the two I2C's?

fw/preamp/src/front_panel.c
Lohrer: // Assumes the CPU's HCLK is 8 MHz from HSI,
// so this function is not robust AT ALL to changes in HCLK
// Max of 2^24, or ~16.7s with 8 MHz / 8
void delay_us(uint32_t us) {
  // Disable SysTick and its interrupts and set the clock source to HCLK / 8
  SysTick->CTRL &= ~(SysTick_CTRL_ENABLE_Msk | SysTick_CTRL_TICKINT_Msk | SysTick_CTRL_CLKSOURCE_Msk);
  SysTick->LOAD = nus;                        // Set the time to count down from
  SysTick->VAL = 0;                           // Clear the counter
  SysTick->CTRL |= SysTick_CTRL_ENABLE_Msk;   // Start countdown (and read CTRL which resets COUNTFLAG)

  // Keep reading CTRL until either SysTick is disabled or has counted to 0
  uint32_t ctrl;
  do {
    ctrl = SysTick->CTRL;
  } while((ctrl & SysTick_CTRL_ENABLE_Msk) && !(temp & SysTick_CTRL_COUNTFLAG_Msk));

  SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;  // Disable the counter
  SysTick->VAL = 0X00;                        // Clear the counter
}

References:
https://developer.arm.com/documentation/dui0497/a/cortex-m0-peripherals/optional-system-timer--systick/systick-control-and-status-register
https://www.st.com/en/microcontrollers-microprocessors/stm32f030r8.html#documentation (RM0360 and DS9773)

Lohrer:
You can shrink the flash space used by debug builds to 2/3 the size by setting -Og. In Project->Properties C/C++ Build->Settings, then Tool Setting tab (make sure Debug config is selected), under MCU GCC Compiler->Optimization set the `Optimization Level` to `Optimize for Debug (-Og). This setting performs any optimizations that won't interfere with proper debugging.

I'd commit the changes to fw/preamp/.cproject but I'm afraid Eclipse will start some Linux/Windows fight if I do that...

The Release build is set to -O3 which is great. If we ever run out of space we can drop down to -O2 or even go sideways to -Os.

amplipi/rt.py
```diff
@@ -63,6 +64,15 @@ def __init__(self, mock=False):
       self.bus = None
       print('Mocking preamp connection')
     else:
+      # Reset preamp board before establishing a communication channel
+      GPIO.setmode(GPIO.BCM)
+      GPIO.setup(4, GPIO.OUT)
+      GPIO.output(4, 1)
```
Lohrer, line 70: A couple things here.
- Do we really need to set NRST high before pulling low? There is an internal 40k pull-up in the STM32 so it should already be high.
  - NRST becomes an output
- These times all seem super long.
  - 100ms for resetting seems excessive, 1ms should be more than enough. The datasheet isn't very clear on how long is required but seems to imply >300ns would work.
  - Is the 1 second mentioned the time it takes for all 6 units to restart? Even assuming they reset in sequence that's 167ms per unit, if they are taking that long to be ready to talk to something is wrong.
- Is there a way to check for the serial connection to be ready? Try sending a test message or something? Then we don't have to figure out some max wait time, we just keep retrying at a fast interval until it's good-to-go. Do we ever use the UART's RX (to the Pi)? If not we could even use that as GPIO to signal ready.

linknum23 (approved):

PR #130, opened 2021-06-06
Bootstrapping script upgrades

Lohrer:
- Default to desktop Raspberry Pi OS image, but allow lite version with --cli-only flag
- Remove `diskpath` option and instead use `/dev/disk/by-id/usb-RPi-MSD*` to identify the Pi.
  TODO: This needs further testing with a second Pi.
- Replace simple delays with checks to verify next step is ready.
  - Check for USB VID:PID before trying to write image
  - Check for the boot and root partitions before trying to mount them
- I keep accidentally setting the hostname to some random fat-fingered parameter setting, so I made it `--hostname HOSTNAME` instead of just `HOSTNAME`. My sanity has vastly improved.

### DIFF
--- a/scripts/bootstrap_pi
+++ b/scripts/bootstrap_pi
@@ -6,10 +6,11 @@ helptext="Usage: bootstrap_pi diskpath [hostname] [--debug] [-h|--help]
 Bootstrap a Raspberry Pi Compute Module for first-time AmpliPi setup.
 
   diskpath:   The path to the mounted raspberry pi disk. Should be /dev/sdX
-  hostname:   The hostname to set on the pi, so once this script successfully completes
-              connect via {hostname}.local. Default is amplipi
+  hostname:   The hostname to set on the pi, so once this script successfully
+              completes connect via {hostname}.local. Default is amplipi
+  --cli-only: Use the lite version of Raspberry Pi OS which does not have a
+              desktop environment
   --debug:    Will print all commands to be run and disable cleanup on exit
-  --desktop:  Use the desktop version of Raspberry Pi OS instead of the lite version
   -h, --help: Print this help text.
 "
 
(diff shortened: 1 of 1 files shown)


Lohrer:
@linknum23, could you check what shows up under `/dev/disk/by-id/` for a 32GB compute module if one's on hand to check?

1. Turn on with service cable plugged in
2. Run `sudo ./rpiboot`
3. `ls /dev/disk/by-id/usb-RPi-MSD*`
4. If the above shows nothing, just dump `ls /dev/disk/by-id/`

Looking at how UDEV names things here, it appears the name I'm seeing for mine, usb-RPi-MSD-\_0001\_1b95b568-0:0, comes from $env{ID_BUS}-$env{ID_VENDOR}\_$env{ID_MODEL}\_$env{ID_SERIAL_SHORT}-$env{ID_INSTANCE}.
ID_BUS will always be `usb`, I expect ID_VENDOR will always be `RPi-MSD-` but this should be verified. The ID_MODEL may or may not always be the same but I figured `usb-RPi-MSD` is enough to know it's the thing we're looking for.

linknum23 (approved):
Nice changes, looking awesome. The readability of this really improved.

linknum23:
Just getting back to this, I plugged in the Pi on my desk and verified the eMMC was 32GB, looks like the id is 0001_85f5fc90.
Here is the output from `ls /dev/disk/by-id/usb-RPi-MSD*`:
`/dev/disk/by-id/usb-RPi-MSD-_0001_85f5fc90-0:0  /dev/disk/by-id/usb-RPi-MSD-_0001_85f5fc90-0:0-part1  /dev/disk/by-id/usb-RPi-MSD-_0001_85f5fc90-0:0-part2
`

Lohrer:
Sweet! This should be good to merge then. The only question is what happens if multiple are plugged in, but I don't even know if `rpi_boot` can handle that.

PR #135, opened 2021-06-15
Add PA

linknum23:
Add PA-like announcement system

- [ ] Play announcement from:
    - [x] URL
    - [ ] Llive stream input
    - [ ] File
- [x] Default to all zones, 50% volume
- [x] Specify zones and groups
- [x] Specify Volumes
- [ ] Announcement in progress notification


linknum23:
Merging this in since the underlying functionality is needed for #158. Although it is not fully featured yet it is quite usable. The additional functionality we hoped to include will be tracked in #163.

PR #148, opened 2021-07-28
Update schematics and hardware info for developer units

Lohrer:


linknum23 (commented):
Looks good. Just add relative links to the pdfs in the table, and add a notes section to explain that 0A and 1A are effectively the same board.

linknum23 (approved):

gorski123 (commented):
Updated controller board sch and led board sch in schematics branch

PR #149, opened 2021-07-29
Add zeroconf/mDNS advertisement to simplify network discovery

linknum23:
Implements #152

### DIFF
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -686,7 +686,7 @@ def create_app(mock_ctrl=None, mock_streams=None, config_file=None, delay_saves=
   return app
 
 def advertise_service(port):
-  """ Advertise the AmpliPi api via zerconf, can be verified with 'avahi-browse -ar' """
+  """ Advertise the AmpliPi api via zeroconf, can be verified with 'avahi-browse -ar' """
   hostname = f'{gethostname()}.local'
   url = f'http://{hostname}'
   if port != 80:
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -38,6 +38,7 @@ from functools import lru_cache
 import asyncio
 import json
 import yaml
+from time import sleep
 
 from PIL import Image # For custom album art size
 
@@ -51,6 +52,7 @@ from starlette.responses import FileResponse
 from sse_starlette.sse import EventSourceResponse
 
 # mdns service advertisement
+import netifaces as ni
 from socket import gethostname, inet_aton
 from zeroconf import IPVersion, ServiceInfo, Zeroconf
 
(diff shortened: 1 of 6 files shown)


amplipi/app.py
```diff
@@ -681,6 +685,35 @@ def create_app(mock_ctrl=None, mock_streams=None, config_file=None, delay_saves=
...
+def advertise_service(port):
+  """ Advertise the AmpliPi api via zerconf, can be verified with 'avahi-browse -ar' """
+  hostname = f'{gethostname()}.local'
+  url = f'http://{hostname}'
+  if port != 80:
+    url += f':{port}'
+  info = ServiceInfo(
+    "_http._tcp.local.",
+    "amplipi-api._http._tcp.local.", # this is named AmpliPi-api to distinguish from the common Spotify/Airport name of AmpliPi
+    addresses=[inet_aton("127.0.0.1")],
```
Lohrer, line 697: I'm seeing the "address" field advertised as 127.0.0.1 even when on another computer. I assume this should be an appropriate IP for each interface? E.g. I'd expect to see 127.0.0.1 on `lo` but not on `eth0`, nor on another computer at all.

amplipi/app.py
```diff
@@ -681,6 +685,35 @@ def create_app(mock_ctrl=None, mock_streams=None, config_file=None, delay_saves=
...
+      'path': '/api/',
+      # extra info - for interfacing
+      'name': 'AmpliPi',
+      "vendor": 'MicroNova',
+      'version': utils.detect_version(),
+      # extra info - for user
+      'web_app': url,
+      'documentation': f'{url}/doc'
+    },
+    server=f'{hostname}.', # Every other service has a trailing '.' here. If it is missing Zeroconf's example/browser.py fails to process this.
```
Lohrer, line 710: See the definition of an SRV record for why a trailing dot is required for the service and target https://en.wikipedia.org/wiki/SRV_record

amplipi/app.py
```diff
@@ -681,6 +685,35 @@ def create_app(mock_ctrl=None, mock_streams=None, config_file=None, delay_saves=
   get_ctrl().reinit(settings, change_notifier=notify_on_change)
   return app
 
+def advertise_service(port):
+  """ Advertise the AmpliPi api via zerconf, can be verified with 'avahi-browse -ar' """
```
Lohrer, line 689:
```suggestion
  """ Advertise the AmpliPi api via zeroconf, can be verified with 'avahi-browse -ar' """
```

Lohrer:
Since I'm not super familiar with zeroconf I'm not sure if this is weird or not, but here is something that might be an issue:

1. Start debug webserver. Running `avahi-browse -ar` shows amplipi-api (plus other services on my network).
2. Boot another computer. Running `avahi-browse -ar` on this fresh computer shows the other services, but no amplipi-api.
3. Kill then start debug webserver again on first computer. Now on second computer amplipi-api service shows up.

linknum23:
I think the issue is that the process running discovery is closing after advertising, lemme check into this.

amplipi/app.py
linknum23: Hmm, I'll check into this. I wonder what is advertised if that field is not populated.

linknum23:
I tried adding local but for some reason none of this works on my home network. @Lohrer can you test again when you get a chance?

PR #173, opened 2021-09-08
Add logic back to keep track of expected pandora state

linknum23:


amplipi/streams.py
```diff
@@ -451,11 +452,13 @@ def send_cmd(self, cmd):
...
       if cmd in supported_cmds:
-        if cmd in cmd_states:
-          self.state = cmd_states[cmd]
         with open(self.ctrl, 'w') as file:
           file.write(supported_cmds[cmd])
           file.flush()
+        if cmd in cmd_states:
+          self.state = cmd_states[cmd]
+        else:
+          self.state = 'playing' # all other commands start playback even from paused
```
micronova-jb, line 461: Just checked on an actual AmpliPi - this is almost completely true! 'love' is the only command that doesn't work in this way. Perhaps this else should change to "elif cmd is not 'love':" or something like that? Other than that, everything looks good

micronova-jb (changes requested):

micronova-jb (approved):
New changes look good!

PR #175, opened 2021-09-08
Support expansion unit programming

Lohrer:
Mostly changes to support expansion programming, but also various other firmware-related fixes.

- Reset expander during self-initialization instead of after, saving 1ms at startup
- Add register to control expander NRST and BOOT0, and allow UART passthrough
- Start hw.py as an upgrade to rt.py. For now its main purpose it to perform firmware upgrades of all detected preamps.
- Added UART1 (Pi<->preamp UART) auto-baud detection. Whatever baud is detected on UART1 is set on UART2 so that all UARTs in the system will talk at the same rate.
- Set I2C2 (internal I2C) to 100 kHz, it was improperly set to 16.7 kHz previously.
- Started `tests/test_fw.py` which will be used to add basic firmware testing. For now just adding flashing and verifying I2C responds.


micronova-jb (approved):

linknum23 (approved):
Looks great!

PR #212, opened 2021-11-10
Fan PWM and Linear Voltage Control

Lohrer:
Power Board Rev 3's will upgrade to PWM fan control, while the new Power Board Rev 4's will utilize their new digital potentiometer to do linear voltage fan control.

The internal I2C bus speed is increased to 400kHz and three new I2C registers were added: PI_TEMP, FAN_DUTY and FAN_VOLTS.


hw/tests/preamp.py
```diff
@@ -40,17 +40,26 @@ def print_status(p: amplipi.rt._Preamps, u: int):
...
 
   # Fans
-  override, fans_on, ctrl, ovr_tmp, failed = p.read_fan_status(u)
+  ctrl, fans_on, ovr_tmp, failed = p.read_fan_status(u)
+  fan_duty = preamps.read_fan_duty(args.u)
   print('  Fan Status')
-  fan_ctrl = 'MAX6644' if ctrl == False else 'ON_OFF'
-  fans_forced = ' (forced)' if override == True else ''
-  print(f'    Fans: On={fans_on}{fans_forced}, Failed={failed}, Control={fan_ctrl}')
+  if ctrl == 0:
```
linknum23, line 52: Might as well make an enum for the different control modes

amplipi/rt.py
```diff
@@ -293,31 +297,44 @@ def read_power_status(self, preamp: int = 1) -> Tuple[Union[bool, None],
...
 
   def read_fan_status(self, preamp: int = 1) -> Tuple[Union[bool, None],
-    Union[bool, None], Union[bool, None], Union[bool, None], Union[bool, None]]:
+    Union[int, None], Union[bool, None], Union[bool, None]]:
     """ Read the status of the fans
 
       Returns:
-        override:  True if the fans have been forced on
+        ctrl       Fan control method currently in use:
+                   0 = MAX6644, 1 = PWM, 2 = Linear, 3 = Forced
```
linknum23, line 311: Enum

fw/preamp/src/fans.c
```diff
@@ -0,0 +1,237 @@
...
+ *    amp_temp: Temperature of the amplifier heatsinks
+ *    psu_temp: Temperature of the high-voltage PSU
+ *    pi_temp:  Temperature of the Raspberry Pi
+ *    force:    Force fans on 100%
+ *    linear:   Digital potentiometer for linear voltage control is available
+ * All temps are in Q7.8 fixed-point format.
+ *
+ * Returns the current fan state.
+ */
+FanState* updateFans(int16_t amp_temp, int16_t psu_temp, int16_t rpi_temp,
```
linknum23, line 140: It looks like we can dynamically switch fan control methods since the control method detection is coupled with the control. Does this make sense? Should detection be separate and/or only done at startu

fw/preamp/src/fans.c
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  return &state;
+}
+
+/* Determines the GPIO expander's FAN_ON pin state.
+ * Call every millisecond tick.
+ *
+ * Inputs
+ *    duty_f8: Fan duty cycle in the range [0,1] in UQ1.7 format
```
linknum23, line 229: duty
```suggestion
 *    duty_f7: Fan duty cycle in the range [0,1] in UQ1.7 format
```

fw/preamp/src/fans.c
linknum23: tldr just make a comment explaining its dynamic

linknum23 (approved):
Looks great. Thanks for updating the fw docs (and fixing the centering).

fw/preamp/src/fans.c
Lohrer: Good point, I'm going to get this one in the next PR though since I've already touched a bunch of code removing the `state` variable.

PR #216, opened 2021-11-17
Make the web updater user friendly

linknum23:
The web updater needs a little polishing
- [x] Make github release files work as updates
- [x] Add update files for older GH release?
- [x] Make update site look like the web app
- [x] Simplify updates to use the latest release on GH
- [x] Add simple update verification
- [x] Fix Manual updates


linknum23:
Guys I want to get this updater in develop soon so we can get some good field testing in.

Please check out the 3 possible update paths:
- Older releases
- Latest version (if you are at the latest version use older to move back)
- Custom (just download a .tar.gz release from GH)

@klay2000 Can you check out the updater site and see if everything makes sense? You can run the updater from the terminal with `./scripts/run_debug_updater` the site is hosted at [http://localhost:5001/update](http://localhost:5001/update).

scripts/configure.py
```diff
@@ -463,6 +463,16 @@ def _configure_authbind() -> List[Task]:
 def _enable_linger(user: str) -> List[Task]:
   return [Task(f'Enable linger for {user} user', f'sudo loginctl enable-linger {user}'.split()).run()]
 
+def _copy_old_config(dest_dir: str) -> None:
+  # try to copy the config of the current runming amplipi service into base_dir/house.json
```
Lohrer, line 467:
```suggestion
  # try to copy the config of the current running amplipi service into base_dir/house.json
```

scripts/deploy
```diff
@@ -104,21 +102,23 @@ if ! $mock_ctrl; then
...
+if $dev; then
+  old_version=$(poetry version -s)
+  developer_version="$(git describe --always --dirty)-$(git rev-parse --abbrev-ref HEAD)"
+  # TODO: write extra git info to the pyproject.toml file since poetry complains about non-standard versions
+  #poetry version $developer_version
 fi
 # package release
 poetry build
 
+if $dev; then
```
Lohrer, line 114: What's going on here?

Lohrer (changes requested):
I don't see any documentation added/changed on how to update, should there be anything?

amplipi/updater/static/js/update-ui.js
```diff
@@ -71,9 +74,13 @@ function ui_reboot_app() {
       setTimeout(ui_check_after_reboot, 3000)
         // TODO: on fail -> show info on how to recover
     } else {
-      ui_add_log('Error restarting update server: ' + response, 'danger')
+      ui_add_log('Error restarting update server: ' + response, 'danger');
+      ui_show_failure();
```
klay2000, line 78: In this kind of place doing
```suggestion
      throw response;
```
makes more sense to me, but it doesn't really matter I suppose.

klay2000 (approved):
Suggested a change, but otherwise looks good!

amplipi/updater/static/js/update-ui.js
klay2000: Ok so the suggested changes feature gives less context of the surrounding code than I expected, but throwing the response there would get caught by another catch that would then log the same thing.

linknum23:
> I don't see any documentation added/changed on how to update, should there be anything?

Umm definitely

scripts/deploy
linknum23: Removed the weird stuff. Added new issue for integrating support for displaying git info on the app

linknum23: #224

PR #228, opened 2021-12-10
Add multiple zone selection

klay2000:
PR to make zone selector a multiselector.


web/templates/index.html.j2
```diff
@@ -323,6 +328,18 @@
   <script src='https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/js/bootstrap.bundle.min.js'></script>
   <!-- Bootstrap JS local fallback -->
   <script>(typeof($.fn.modal) === 'undefined') && document.write('<script src="static/js/libs/bootstrap/4.5.2/bootstrap.bundle.min.js"><\/script>')</script>
+  <!-- UUID CDN -->
```
linknum23, line 331: remove UUID and qs, below and keep these lines

linknum23 (changes requested):

web/templates/index.html.j2
```diff
@@ -283,15 +283,20 @@
...
-              {% for g, name in unused_groups[src].items() %}
-              <option value="g{{ g }}">G - {{ name }}</option>
-              {% endfor %}
-              {% for z, name in unused_zones[src].items() %}
-              <option value="z{{ z }}">Z - {{ name }}</option>
-              {% endfor %}
-            </select>
+            <div class="input-group">
+              <select id="s{{ src }}-add-input" class="multiselect custom-select" multiple="multiple" title="Select Group or Zone">
+                <optgroup label="Groups">
```
linknum23, line 288: Can you hide the group and zone labels if there aren't any?

linknum23 (changes requested):

web/templates/index.html.j2
linknum23: lines 349-358 should be removed, can't suggest that with GH

Lohrer:
Just checked this out. It's a bit awkward since the action that actually performs the zone update is "close the zone dropdown" AKA click anywhere that's not the zone selection dropdown. Some kind of "Add" or "OK" button would be nice, but it does technically work as-is.

It also doesn't seem to work on the Android app at all, but I like that there's an OK button there so it makes more intuitive sense.

linknum23:
> It also doesn't seem to work on the Android app at all, but I like that there's an OK button there so it makes more intuitive sense.

Just tested this in the app and it works as expected.

linknum23 (approved):

Lohrer:
> We can think about adding a button in the future, both Klayton and I were having trouble getting a button to look good CSS-wise.
> Just tested this in the app and it works as expected.

Yeah after testing again I don't get the button that I saw before and things work as expected. I must have tested too soon or maybe grabbed an old commit somehow, I'm not sure.

PR #232, opened 2021-12-16
Add back button to updater

linknum23:
This allows users to easily get back to their AmpliPi app even if they don't have access to a back button. This should make updates easier in the Android App, where back exits the app.

### DIFF
--- a/amplipi/updater/static/css/styles.css
+++ b/amplipi/updater/static/css/styles.css
@@ -3,9 +3,12 @@
  */
 body {
   padding-bottom: 2rem;
-  padding-top: 4rem;
+  padding-top: 1rem;
   color: white;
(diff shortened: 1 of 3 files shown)


klay2000:
looks good to me

Lohrer (approved):
Looks good! Just a note that opening the button's link in a new tab takes you to port 5001 instead of 80 for some reason. Probably doesn't matter since that's not the use-case being solved here.

linknum23:
> Looks good! Just a note that opening the button's link in a new tab takes you to port 5001 instead of 80 for some reason. Probably doesn't matter since that's not the use-case being solved here.

Yeah I don't see a way around that unless we want to add a redirect to port 80 in the updater.

PR #233, opened 2021-12-16
Add refresh on zone display mismatch

linknum23:
Make zones display update when they are moved from one source to another.

To test:
1. Open two instances of the AmpliPi app ( on your phone and also on your browser?)
2. Add a zone to a different source in app 1
3. Verify that app 2 updates to display the new zone

### DIFF
--- a/web/static/js/script.js
+++ b/web/static/js/script.js
@@ -63,7 +63,14 @@ function debounce(ms, fun){
 function refresh() {
   get();
 }
-current_src = 's0';
+current_src = 's0'; // this needs to be set base on the URL path
+for (var i = 0; i < 4; i++) {
+  if (window.location.href.endsWith(`/${i}`)) {
+    current_src = `s${i}`;
+    break;
+  }
+}
+
 $(document).ready(function(){
   // Some things are not part of the automatic tab-content switching
   // hide things related to the old src and show things related to the new one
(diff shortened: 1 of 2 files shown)


Lohrer:
There's some buggery going on here and I can't find the pattern to it. In general, you get kicked over to source 0 half the time there's a zone update. Also, if you go back on say amplipi.local/0 to amplipi.local/3, it kicks you back to amplipi.local/0 after a second.

Lohrer:
Oh also it doesn't work on the mobile app.

linknum23:
> Oh also it doesn't work on the mobile app.

It should now with the new zeroconf interfaces

> There's some buggery going on here and I can't find the pattern to it. In general, you get kicked over to source 0 half the time there's a zone update. Also, if you go back on say amplipi.local/0 to amplipi.local/3, it kicks you back to amplipi.local/0 after a second.

Try it out now, seemed more stable in my limited testing

web/static/js/script.js
```diff
@@ -63,6 +63,13 @@ function debounce(ms, fun){
 function refresh() {
   get();
 }
+current_src = 's0'; // this needs to be set base on the URL path
```
Lohrer, line 66:
```suggestion
current_src = 's0'; // this needs to be set based on the URL path
```

Lohrer (approved):
Works much nicer now, both on web and Android!

PR #235, opened 2021-12-16
Mute preouts and auto-standby in firmware

Lohrer:
Also implements the final major refactoring (at least for a while of course) which integrates the audio controls and as much as possible groups functionality by physical device.


linknum23:
Is this still supposed to be a draft?

fw/preamp/src/audio.c
```diff
@@ -48,104 +45,120 @@ const I2CReg zone_right_[NUM_ZONES] = {
...
+  // Mute pins are active-low
+  return !readPin(zone_mute_[zone]);
 }
 
 // Connect a Zone to a Source
 void setZoneSource(size_t zone, size_t src) {
   // Mute the zone during the switch to avoid an audible pop
-  bool was_muted = !isOn(zone);
+  bool was_muted = muted(zone);
   mute(zone, !was_muted);
```
linknum23, line 162: This code doesn't do what the comment says, try this
```suggestion
  mute(zone, true);
```

fw/preamp/src/audio.c
```diff
@@ -196,3 +202,32 @@ InputType getSourceAD(size_t src) {
...
+void initAudio() {
+  // Initialize each zone's audio state (does not write to volume control ICs)
+  for (size_t zone = 0; zone < NUM_ZONES; zone++) {
+    mute(zone, true);
+    setZoneSource(zone, DEFAULT_SOURCE);
+  }
+
+  // Initialize each source's analog/digital state
+  for (size_t src = 0; src < NUM_SRCS; src++) {
+    setSourceAD(src, IT_DIGITAL);
```
linknum23, line 215:
```suggestion
    // Set all sources to Digitial initially since the digital input is by default nothing.
    // This avoids any unwanted playback.
    // Also, expanders need to be set to digital to get passed through audio (this is also expander fw).
    setSourceAD(src, IT_DIGITAL);
```

linknum23 (changes requested):
parital review, need another pass to check out other changes

linknum23 (approved):
Looks good. Only thing I would recommend is putting it though the tester to make sure nothing in pins.h got messed up.

PR #236, opened 2021-12-21
Add support for zeroconf advertising on various interfaces (not just eth0)

linknum23:
This will enable users to use a USB Wifi Dongle and also make it easier for testing with the debug webserver running on a linux-based computer.


Lohrer (approved):
Looks good to me, and didn't break Android finding AmpliPi so seems to work for me! I did notice a fair number of double-quoted strings that could have been single-quoted in app.py, if you care about that.

PR #240, opened 2021-12-27
I2C Arbitration Lost Fix

Lohrer:
If the preamp's micro is reset in the middle of sending an I2C message, the bus is left in a bad state with the addressed slave expecting the rest of the message. The old method that attempted to fix this was rough and only partially fixed the problem. Now, At every startup the micro will send clock pulses on SCL until enough 1's are read on the SDA line to consider the bus inactive.  The old method failed to initialize the bus once every ~150 times, the new method has successfully initialized the bus hundreds of thousands of times in a row with no errors.

This PR also currently includes two other minor changes:
- Initialize the ADC's Setup Byte. Previously the reset value of the ADC's Setup Byte was being relied on. But if the ADC receives a write of all ones (as is the case if the ADC was addressed in write mode, then the micro resets, then 9 clocks are sent with SDA high) then the Setup Byte is written to all ones. This switches the reference from VDD (3.3V) to a 2.048V internal reference so all the temps and voltages start reading as much higher than they really are.
- Since STANDBY is now handled by logic in firmware, slightly modify the STANDBY register to act as an AMP_DISABLE register. It will default to 0x3F, but if any of the lower 6 bits are written low the corresponding zone amp will be disabled. A disabled amp means it won't count as on when determining standby. This could be useful to save power if preout-only zones are used. I haven't added any software support for this setting yet. If old software communicates to this register thinking it's still the standby register, it will write zeros expecting to enter standby but the new firmware will disable the amps, which achieves the same result. So this change is backwards-compatible.

### DIFF
--- a/fw/preamp/src/audio.c
+++ b/fw/preamp/src/audio.c
@@ -56,6 +56,9 @@ uint8_t vol_req_[NUM_ZONES] = {VOL_MUTE};
 // The TDA7448 volume controller always reports 0x00 on read
 uint8_t vol_[NUM_ZONES] = {0};
 
+// If only preouts are used, the amplifier for a zone can be disabled
+bool amp_en_[NUM_ZONES] = {};
+
 // Convert a requested dB to the corresponding volume IC register value.
 static inline uint8_t dB2VolReg(uint8_t db) {
   /* The volume IC has a discontinuity in its register value to attenuation
@@ -131,14 +134,25 @@ bool inStandby() {
   return false;
 }
 
-// Returns true if any zone is unmuted
-static bool anyZoneOn() {
+// Returns true if any zone is unmuted and has amps enabled
+static bool anyZoneAmpOn() {
   for (uint8_t zone = 0; zone < NUM_ZONES; zone++) {
-    if (!muted(zone)) {
+    if (!muted(zone) && amp_en_[zone]) {
       return true;
     }
   }
   return false;
+}
+
+void enZoneAmp(size_t zone, bool en) {
+  amp_en_[zone] = en;
+
+  // If no zones are on, standby
+  standby(!anyZoneAmpOn());
+}
+
+bool zoneAmpEnabled(size_t zone) {
+  return amp_en_[zone];
 }
 
 // Mute the specified zone
@@ -147,7 +161,7 @@ void mute(size_t zone, bool mute) {
   writePin(zone_mute_[zone], !mute);
 
   // If no zones are on, standby
-  standby(!anyZoneOn());
+  standby(!anyZoneAmpOn());
 }
 
 bool muted(size_t zone) {
@@ -206,6 +220,7 @@ InputType getSourceAD(size_t src) {
 void initAudio() {
   // Initialize each zone's audio state (does not write to volume control ICs)
   for (size_t zone = 0; zone < NUM_ZONES; zone++) {
+    enZoneAmp(zone, true);
     mute(zone, true);
     setZoneSource(zone, DEFAULT_SOURCE);
   }
--- a/fw/preamp/src/audio.c
+++ b/fw/preamp/src/audio.c
@@ -56,7 +56,13 @@ uint8_t vol_req_[NUM_ZONES] = {VOL_MUTE};
 // The TDA7448 volume controller always reports 0x00 on read
 uint8_t vol_[NUM_ZONES] = {0};
 
-// If only preouts are used, the amplifier for a zone can be disabled
+// If any zone uses only the preout, the amp can be 'disabled' which will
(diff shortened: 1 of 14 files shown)


fw/preamp/src/i2c.c
```diff
@@ -75,6 +75,51 @@ void initI2C2() {
   I2C_Cmd(I2C2, ENABLE);
 }
 
+void deinitI2C2() {
+  /* Disable I2C2 and release GPIO control */
+
+  // Ensure I2C clock is enabled
+  RCC_APB1PeriphClockCmd(RCC_APB1Periph_I2C2, ENABLE);
```
linknum23, line 82: why does the clock need to be enabled before i2c is disabled?

fw/preamp/src/i2c.c
```diff
@@ -75,6 +75,51 @@ void initI2C2() {
...
+
+uint32_t writeByteI2C2(I2CDev dev, uint8_t val) {
+  // Wait if I2C2 is busy
+  while (I2C2->ISR & I2C_ISR_BUSY) {}
+
+  // Setup to send send start, addr, subaddr
+  I2C_TransferHandling(I2C2, dev, 1, I2C_AutoEnd_Mode,
+                       I2C_Generate_Start_Write);
+
+  // Wait for transmit interrupted flag or an error
```
linknum23, line 96:
```suggestion
  // Wait for transmit interrupted flag or an error
  // TODO: Add a timeout condition
```

fw/preamp/src/i2c.c
```diff
@@ -75,6 +75,51 @@ void initI2C2() {
...
+      I2C2->ICR = I2C_ICR_ARLOCF;
+      return I2C_ISR_ARLO;
+    }
+    isr = I2C2->ISR;
+  } while (!(isr & I2C_ISR_TXIS));
+
+  // Send data
+  I2C_SendData(I2C2, val);
+
+  // Wait for stop flag to be sent and then clear it
```
linknum23, line 117:
```suggestion
  // Wait for stop flag to be sent and then clear it
  // TODO: handle timeout?
```

fw/preamp/src/int_i2c.c
```diff
@@ -38,121 +39,72 @@ const I2CReg pwr_io_gpio_ = {0x42, 0x09};
...
+ */
+void quiesceI2C() {
+  const uint32_t HALF_PERIOD = 3;  // 4 us period = 166.7 kHz I2C clock
+  // Ensure the I2C peripheral is disabled and pins are set as GPIO
+  // Pins will be configured to HI-Z (pulled up externally)
+  deinitI2C2();
+  configI2C2PinsAsGPIO();
+
+  const uint32_t NUM_CONSECUTIVE_ONES = 9;
+  // Require NUM_CONSECUTIVE_ONES on I2C's SDA before proceeding.
```
linknum23, line 65: Why?

linknum23 (approved):
Looks good

fw/preamp/src/audio.c
```diff
@@ -131,23 +134,34 @@ bool inStandby() {
   return false;
 }
 
-// Returns true if any zone is unmuted
-static bool anyZoneOn() {
+// Returns true if any zone is unmuted and has amps enabled
+static bool anyZoneAmpOn() {
   for (uint8_t zone = 0; zone < NUM_ZONES; zone++) {
-    if (!muted(zone)) {
+    if (!muted(zone) && amp_en_[zone]) {
```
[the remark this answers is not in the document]
Lohrer: Uh it would be if there wasn't a bug in the logic... Good catch! Turns out the amps weren't actually muting, just going into standby.

fw/preamp/src/i2c.c
Lohrer: This is the peripheral clock, can't write to any of the peripheral's registers if it's disabled. `I2C_Cmd(I2C2, DISABLE);` writes to an I2C register to disable it. Technically right now `deinitI2C2()` isn't even needed since it's being called before I2C is initialized (hence why the peripheral's clock isn't enabled), but there was a thought to further robustify things in the future and if a `I2C_ISR_ARLO` is ever thrown call the `quiesceI2C()` function. This would de-init the bus, send extra clocks to try and resolve the situation, then re-init the bus.

linknum23: Some part of this needs to be in a comment in there, but you already know that :)

Lohrer:
I realized when I slowed down the clock in quiesceI2C() the real issue was I forgot to add a small delay before attempting communication with the preamp. I sped it up again but I'll test overnight again to be sure there's not some other issue that pops up from going slightly faster.

PR #246, opened 2022-01-03
Improved firmware programming

Lohrer:
Previously firmware programming, by default, attempted to program the number of units that were able to be communicated with before programming. This method is flawed if any of the preamps are in a broken or completely unprogrammed state. So the new method implemented here always attempts to program one more unit until a unit fails to program. That way newly plugged-in units or previously unprogrammed ones are always at least attempted to be programmed.

### DIFF
--- a/amplipi/hw.py
+++ b/amplipi/hw.py
@@ -323,11 +323,14 @@ class Preamps:
     present = preamp.available()
     print(f"{self.unit_num_to_name(unit)}'s old firmware version: {preamp.read_version() if present else 'unknown'}")
 
+    # For now the firmware can only pass through 9600 baud to expanders
+    baud = 9600 if unit > 0 else baud
+
     # Reset into bootloader mode and check if bootloader can be communicated to
     self.reset(unit = unit, bootloader = True)
     if not present:
       present = subprocess.run([f'stm32flash -b {baud} {PI_SERIAL_PORT}'],
-                                shell=True, check=False).returncode == 0
+                                shell=True, check=False, stdout=subprocess.DEVNULL).returncode == 0
     if not present:
       # Unit not found, give up
       # TODO: retry?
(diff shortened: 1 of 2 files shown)


amplipi/hw.py
```diff
@@ -290,56 +299,118 @@ def set_i2c_address(self, baud: int = 9600) -> bool:
...
-        print(f"Can't communicate with unit {unit}, stopping programming")
-
-      if not success:
-        break
-    return success
+    # Force attempting to program at least 1 unit
+    if num_units is not None and num_units <= 0:
+      num_units = 1
+    previous_unit_count = len(self)
+    program_count = self.MAX_UNITS if num_units is None else num_units
```
linknum23, line 379: Just attempting to clean up and consolidate the logic here
```suggestion
    if num_units is None:
      # try to program as many units as is feasible by default
      # this is an attempt to never leave any units in a bad firmware state
      program_count = self.MAX_UNITS
    elif num_units <= 0:
      # TODO: since we were instructed to do nothing, why not just do nothing and return an error?
      return False
    else:
      program_count = num_units
    previous_unit_count = len(self)
```

amplipi/hw.py
```diff
@@ -290,56 +299,118 @@ def set_i2c_address(self, baud: int = 9600) -> bool:
...
+    # Attempt programming until program_count, but stop if an error occurs
+    unit = 0
+    success = True
+    while success and unit < program_count:
+      print()
+      success = self.program(filepath, unit)
+      if success:
+        unit += 1
+
+    # Success considerations:
```
linknum23, line 394:
```suggestion
    # Since we are programming until failure, we need to have a different view of success
    # Success considerations:
```

linknum23 (changes requested):

PR #250, opened 2022-01-14
Added schematics for final kickstarter orders

Lohrer:
Latest hardware batch that's currently shipping.


hw/sch/hw_releases.md
```diff
@@ -60,15 +60,13 @@ that acts as a pass-through for audio and control signals.
 Up to 5 Expansion Units can be added to one Controller.
```
linknum23, line 60: Shouldn't these be called zone expanders?

linknum23 (approved):
I would suggest changing the expansion units to zone expanders, but I am not sure how good we have been about maintaining naming consistency for this.

hw/sch/hw_releases.md
Lohrer: You know, I totally had "expansion unit" as the official name in my head but you're right!

gorski123 (approved):

PR #252, opened 2022-01-15
Add zone min and max volumes

Lohrer:


amplipi/rt.py
```diff
@@ -523,15 +523,15 @@ def update_zone_vol(self, zone, vol):
 
       Args:
         zone: zone to adjust vol
-        vol: int in range[MIN_VOL, 0]
+        vol: int in range[MIN_VOL_DB, 0]
```
linknum23, line 526: Can this be an F-string?

amplipi/utils.py
```diff
@@ -88,6 +89,14 @@ def clamp(xval, xmin, xmax):
   """ Clamp and value between min and max """
   return max(xmin, min(xval, xmax))
 
+def round_sf(x: float, sf: int = 3):
```
linknum23, line 92: used?

amplipi/models.py
```diff
@@ -342,6 +384,7 @@ class GroupUpdate(BaseUpdate):
   zones: Optional[List[int]] = fields.Zones
   mute: Optional[bool] = fields.GroupMute
   vol_delta: Optional[int] = fields.GroupVolume
+  vol_delta_f: Optional[float] = fields.GroupVolumeF
```
linknum23, line 387: lets call this `vol_f` with the intent of deprecating `vol_delta` to `vol`

tests/test_rest.py
```diff
@@ -759,3 +772,211 @@ def on_service_state_change(zeroconf: Zeroconf, service_type: str, name: str, st
   # check advertisememts
   assert AMPLIPI_ZC_NAME in services_advertised
   assert services_advertised[AMPLIPI_ZC_NAME].port == 9898
+
```
linknum23, line 775: These tests are really valuable! I know they are pretty verbose but I don't know how much is gained by refactoring.

linknum23 (changes requested):

amplipi/rt.py
Lohrer: Not without a decorator, plus I'd argue that would mask how to properly use this method.

linknum23 (approved):

PR #254, opened 2022-01-18
Server Sent Events

linknum23:
- [x] Add partial change concept
- [x] Broadcast SSE Events
- [x] Plumb full state changes to make the web migration and testing simple at first
- [ ] Add stream info polling loop that is able to push changes
- [ ] Add SSE receiver/change detector for test framework
- [ ] Remove old save handler and integrate saving into the partial change framework
- [ ] Update web interface to handle partial changes from SSE

### DIFF
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -153,13 +153,13 @@ def reset(ctrl: Api = Depends(get_ctrl)) -> models.Status:
   ctrl.reinit(settings=ctrl._settings, change_notifier=notify_on_change)
   return ctrl.get_state()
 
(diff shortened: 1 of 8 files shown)


rtertiaer:
@linknum23  if you intend to ship this eventually, can you re-target this PR to `main`? Github automagically closed it when I deleted `develop` :woozy_face: sorry!

linknum23:
Oh this one deserves a conversation. Let's talk about when something like SSE makes sense. 

There's no way this rebases without 100 conflicts.

PR #277, opened 2022-01-26
Fix Fastapi deps

linknum23:


Lohrer (approved):
Looks great! I ran into the python3.6 issue and tried to allow changing the python version but ran into issues with building the wheel for rpi.gpio, so I settled on adding a warning for python < 3.7.

PR #278, opened 2022-01-28
Fix Volume Throttling

linknum23:
Make volume throttling simpler and less asynchronous  
- This fixes a weird bug we were seeing on mobile where many many volume changes were getting incorrectly throttled.

### DIFF
--- a/web/static/js/script.js
+++ b/web/static/js/script.js
@@ -449,7 +449,7 @@ function initVolControl(ctrl) {
       }
       last_req_stamp = cur_stamp; // only update on a successful request (avoids constant rejection of a stream of user requests)
     } else {
-      console.log('volume adjustment rejected, last request made < 50ms ago')
+      console.debug('volume adjustment rejected, last request made < 50ms ago');
     }
   }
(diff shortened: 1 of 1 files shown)


web/static/js/script.js
```diff
@@ -445,17 +447,15 @@ function initVolControl(ctrl) {
...
         console.log('volume control ' + ctrl.id + ' not bound to any zone or group');
       }
-      req_throttled = true;
-      setTimeout(() => {
-        req_throttled = false
-      }, VOL_REQ_THROTTLE_MS);
-
+      last_req_stamp = cur_stamp; // only update on a successful request (avoids constant rejection of a stream of user requests)
+    } else {
+      console.log('volume adjustment rejected, last request made < 50ms ago')
```
Lohrer, line 452: Was this intended to be left in? It tends to flood the console when dragging the volume sliders. Also, I'm no Javascript expert but if left in should it have a semicolon like the edit below adding one?

Lohrer (approved):

PR #282, opened 2022-02-01
Add internet radio search

kjk2010:
This PR adds a station search when adding a new Internet Radio station stream to AmpliPi. It uses the pyradios library (https://github.com/andreztz/pyradios) for the Radio Browser API (https://api.radio-browser.info/).


amplipi/app.py
```diff
@@ -409,6 +412,14 @@ def announce(announcement: models.Announcement, ctrl: Api = Depends(get_ctrl)) -
   """ Make an announcement """
   return code_response(ctrl, ctrl.announce(announcement))
 
+# Internet radio catalog search
```
linknum23, line 415: I'm trying to figure out whether or not it makes more sense use a javascript library on the web app. This one looks like it would be pretty easy to integrate: https://github.com/nepodev/radio-browser.

Looking at moving to a js solution looks like it has some pros and cons:

Pros:
- removes an additional redirect for the station lookup
- keeps the API a little simpler

Cons:
- Doesn't support low powered interfaces 
- Additional js module to download


My instincts are to use the js solution assuming it is easy to integrate. What do you think?

linknum23 (commented):
Interface is simple and great. If we end up going with pyradios I would suggest adding it to the requirements.txt.
Also, I can make those tabs look a little more AmpliPi like with a little css magic.

linknum23:
We should be able to make the changes to integrate this soon, in the next couple of weeks. Thanks so much for the initial work here @kjk2010.

linknum23:
I added some fixes to a new but very similar PR to avoid overwriting your branch. See #302

PR #289, opened 2022-02-10
Set random default password for pi user

Lohrer:
During an update, will change the password to a new random one if 'raspberry' is still the password.

### DIFF
--- a/scripts/set_pass
+++ b/scripts/set_pass
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# These give serial numbers, but serials are relatively predictable
+# so let's not use them.
+#cat /proc/cpuinfo | sed -n 's/Serial[[:space:]]*:[[:space:]]0*//p'
+#cat /sys/block/mmcblk0/device/cid
+
+confdir="$HOME/.config/amplipi"
+
+# First verify the current user is 'pi'
+if [[ $USER != pi ]]; then
+  echo "ERROR: Not running as pi user."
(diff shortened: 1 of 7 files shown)


scripts/set_pass
```diff
@@ -0,0 +1,73 @@
...
+  esac
+  shift
+done
+
+# Next check if a password has already been generated
+#if [[ -f $confdir/default_password.txt &&  ]]; then
+#fi
+
+if [ -z "$pass" ]; then
+  # Generate a random passphrase and save to file.
```
linknum23, line 51: Would it make sense to make a function for each of these different password options?

scripts/set_pass
```diff
@@ -0,0 +1,73 @@
...
+# First verify the current user is 'pi'
+#if [[ $USER != pi ]]; then
+#  echo "ERROR: Not running as pi user."
+#  exit 1
+#fi
+
+HELP="Set a password on AmpliPi\n
+usage: set_pass [PASSWORD]\n
+If no PASSWORD is given, generate a random passphrase,\n
+otherwise set the pi user's password to PASSWORD.\n"
```
linknum23, line 35: Add note about this password being stored in plaintext for display?

linknum23 (approved):
Looks good. You may want to clean up the change password script a little more just since several things are commented out.

linknum23:
Changes look great. Lets merge this in so we can test it internally.

PR #296, opened 2022-02-22
Add list of supported commands to source info

linknum23:
This should make it easy to discover different commands supported by different stream interfaces. At the moment only pandora and spotify are supported.

Needs more testing. @klay2000 can you test this out?


linknum23:
@klay2000 Can you test this and make sure the play next buttons still work well on spotify and pandora streams?

Lohrer (approved):
I was hoping to get a chance to run this on a real amplipi but haven't. The code looks good to me though.

linknum23:
Lets just test this further on develop.

PR #299, opened 2022-02-25
Fix missed vol_delta_f -> vol_f group volume changes

linknum23:


Lohrer (approved):
Looks good thanks for fixing! I don't see any other missed locations and it fixes both problems I noticed.

PR #302, opened 2022-03-05
Add inetradio search

linknum23:
This PR adds a station search when adding a new Internet Radio station stream to AmpliPi, from @kjk2010 with some minor tweaks.

Left this in draft mode as the ui still needs a little bit of css.

### DIFF
--- a/web/static/js/settings.js
+++ b/web/static/js/settings.js
@@ -331,7 +331,7 @@ $(function() {
     console.log('Searching for station by name: ' + $("#inetradio-search-name-txt").val());
     $.ajax({
       type: "GET",
-      url: '/api/inetradio/search/name/' + $("#inetradio-search-name-txt").val(),
+      url: 'https://de1.api.radio-browser.info/json/stations/byname/' + $("#inetradio-search-name-txt").val(),
       contentType: "application/json",
       success: function(data) {
         $("#inetradio-searchNameResults").html("<h3>Search Results</h3>");
--- a/web/static/js/settings.js
+++ b/web/static/js/settings.js
@@ -353,7 +353,7 @@ $(function() {
             '<a href="#" class="btn btn-success btn-sm float-right addStation" data-stationname="' +
             value.name.toString().replace('"', '\\"') +
             '" data-stationurl="' +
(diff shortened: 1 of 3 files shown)


web/static/js/settings.js
```diff
@@ -99,14 +112,39 @@ $(function() {
...
+            </div>
+            <div class="tab-pane fade" id="inetradio-manual" role="tabpanel" aria-labelledby="inetradio-manual-tab">
+              <div class="form-group">
+                <label for="url">Station Audio URL</label>
+                <input type="text" class="form-control" name="url" id="url" aria-describedby="urlHelp" data-required="true">
+                <small id="urlHelp" class="form-text text-muted">Audio URL must be supported by <a href="https://www.videolan.org/" target="_blank">VLC</a>.</small>
+              </div>
+              <div class="form-group">
+                <label for="logo">Station Logo</label>
+                <input type="text" class="form-control" name="logo">
```
Lohrer, line 144: needs `id="logo"`, although both the label's for= and the input's id= should probably be a bit more unique like `inetradio-logo`.

web/static/js/settings.js
```diff
@@ -264,13 +302,86 @@ $(function() {
...
+          );
+        });
+        if (!numResults) { $('#inetradio-search_name_results').append("No stations found."); }
+      }
+    });
+  });
+
+  /* Add an inetradio station from search screen */
+  $(document).on('click', '.addStation', function () {
+    var name = $(this).data('stationname');
```
Lohrer, line 369: This and all the associated hiding of the stream name setting means Internet Radio stations added by the search functionality don't allow custom names until after they are already added. Is this intentional? I'd think it would be nice to populate the name field with a selected radio station's name but allow editing before adding.

Lohrer (changes requested):
I love the ability to search for radio stations!

web/static/js/settings.js
linknum23: I agree. This is going to need a little redesign to make it all work like that though.

linknum23:
I mad the search populate the fields removing the tabs. Still needs t a little more tlc and fixes for the id issues mentioned above. Getting closer though!

PR #310, opened 2022-04-04
Make announcements have the some vol/vol_f api as everything else

linknum23:


Lohrer:
Is there a way we can add a test for the announcement volume? This issue should have caused something to fail. Otherwise looks good!

linknum23:
I'll add a test and rebase on top of it.

PR #316, opened 2022-04-15
Make password generation optional when deploying

Lohrer:
A default password will be generated and set when performing an update, if a default password doesn't exist at ~/.config/amplipi/default_password.txt

If Deploying or Configuring no password changes will be made unless a flag is specifically set.

Also programming the firmware wasn't an option when using the deploy script so I added that here along with the password flag.


linknum23 (approved):
Looks 👍

PR #317, opened 2022-04-19
Kernel updates

linknum23:


scripts/configure.py
```diff
@@ -751,7 +753,7 @@ def failed():
   parser.add_argument('--web','--webserver', action='store_true', default=False,
     help="Install and configure webserver")
   parser.add_argument('--restart-updater', action='store_true', default=False,
-    help="""Stop the updater if it is running and start the updated one. \
+    help="""Restart AmpliPis OS and the updater itself. \
```
Lohrer, line 756:
```suggestion
    help="""Restart AmpliPi's OS and the updater itself. \
```
What do you think of adding '--restart' to this?

scripts/configure.py
```diff
@@ -715,10 +718,9 @@ def failed():
     if failed():
       return False
   if not web and restart_updater: # if web and restart_updater are True this restart happens in the _update_web function
```
Lohrer, line 720: Since restart_updater is False by default, does this mean it won't restart during an update? Only from command-line usage of configure.py?

linknum23: Updater restart needs to be called after the update finishes. So it is called separately by the web updater

scripts/configure.py
linknum23: This gets tricky. The web updater and the configure.py could get out of sync.

linknum23: We could make restart and restart updater aliases?

scripts/configure.py
linknum23: The weird case here is that --restart-updater only reboots AmpliPi if it is called by itself if called with --web it just restarts both of the services.

PR #321, opened 2022-04-28
0.1.8 prerelease

linknum23:


linknum23:
@Lohrer Can you try out the latest test release?  I think it is ready :)

PR #325, opened 2022-05-10
Reboot and Shutdown

### LINKED ISSUE #284: Feature: Add restart button
AmpliPi needs a simpler way to be restarted without having physical access to the device. Add a button in the settings menu to perform a full restart of the controller.

This will require a few things:

- [x] Add updater restart endpoint
- [x] Add button with spinner
- [x] Add background poll waiting for amplipi up (checking every 2 seconds)

linknum23:
Adds reboot and shutdown to the configuration page. 
In addition spinny wheels, done check boxes, and error icons were added to better reflect current state.

Closes #284

### DIFF
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -115,6 +115,7 @@ def ungrouped_zones(ctrl: Api, src: int) -> List[models.Zone]:
   return [zones[z] for z in ungrouped_zones_ if z is not None and not zones[z].disabled]
 
 # add a default controller (this is overriden below in create_app)
+# TODO: this get_ctrl Singleton needs to be removed and the API converted to be instantiated by a class with ctrl state
 @lru_cache(1) # Api controller should only be instantiated once (we clear the cache with get_ctr.cache_clear() after settings object is configured)
 def get_ctrl() -> Api:
   """ Get the controller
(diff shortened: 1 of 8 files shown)


linknum23:
Spend a decent chunk of time making AmpliPi behave as expected and reset the firmware during shutdown. This has been an undocumented issue for a long time.

Lohrer:
- I think shutdown is potentially ok as-is, but it might be nice to add text saying "shutdown sucessfully requested" or something to that effect to avoid the uncertainty of what the green check mark means. Also maybe we should wait for the API to fail to respond? Or does it stop responding immediately after the shutdown request?
- Probably not important enough to add now but what do you think of a confirmation? Reboot isn't too bad but an accidental remote shutdown could be annoying since turning it back on requires being physically present.

Lohrer (commented):
Shutting down or rebooting doesn't save any recent (<5 second or whatever the timer is) config changes. Is this something worth fixing?

I also changed the mouseover text on the reset/reboot/shutdown buttons to text so it can be seen on mobile. Hopefully it looks ok.

linknum23:
It might make sense to do a preemtpive save before starting the reboot or shutdown though

linknum23:
@Lohrer can you try this right now, it is deployed to amplipi10

linknum23:
The only thing not added is a confirmation box since it is a bit of work to implement.

Lohrer (approved):

PR #345, opened 2022-06-23
Fix front-panel volume display

Lohrer:
The front-panel display didn't get properly updated with the floating-point volume changes in the API. As well as fixing that issue, this PR makes is so that if 3 or more expanders are connected (>18 zones) the first 18 zone volumes are still displayed.

### DIFF
--- a/amplipi/display/display.py
+++ b/amplipi/display/display.py
@@ -260,12 +260,11 @@ def draw_volume_bars(draw, font, small_font, zones: List[models.Zone], x=0, y=0,
   n = len(zones)
   if n == 0: # No zone info from AmpliPi server
     pass
-  elif n <= 6: # Draw horizonal bars
+  elif n <= 6: # Draw horizonal bars and full zone names
     wb = int(width / 2)                   # Each bar's full width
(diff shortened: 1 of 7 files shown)


amplipi/display/display.py
```diff
@@ -295,15 +294,14 @@ def draw_volume_bars(draw, font, small_font, zones: List[models.Zone], x=0, y=0,
                 anchor='mm', font=small_font, fill=Color.WHITE.value)
 
       # Draw background of volume bar
-      draw.rectangle(((xb, y, xb+wb, yt)), fill=Color.LIGHTGRAY.value)
+      draw.rectangle((xb, yb, xb+wb, y), fill=Color.LIGHTGRAY.value)
 
       # Draw volume bar
-      if zones[i].vol > models.MIN_VOL:
+      if zones[i].vol_f > models.MIN_VOL:
```
linknum23, line 300: we may want to differentiate _MIN_VOL_ into _MIN_VOL_DB_ and _MIN_VOL_F_

linknum23 (approved):

Issue #119, opened 2021-05-17
Feature: Use RAM disk for stream configuration and logs


micronova-jb:
Update "use_ram.py" to diskify the proper directories. Test on multiple reboots/deletions

Ex: /home/pi/web/generated  --> /home/pi/amplipi-dev/web/generated

Lohrer:
Fixed in #360

PR #360, opened 2022-08-08
Use ram disk for config and web

Lohrer:
Reduce disk usage by moving the config/srcs and web/generated into ram. See #119. Implemented here as part of the configure.py script.

Currently this will delete any other amplipi-created tmpfs entries in fstab. Is there any reason to keep multiple sets of AmpliPi tmpfs locations?


linknum23 (approved):
Looks good! To further reduce sd thrash we are going to have to add some extra logic for saving the config file so only configuration is saved.

linknum23:
Something like what we do in some of the tests:

```python
def prune_state(state: amplipi.models.Status):
  """ Prune generated fields from system state to make comparable """
  dstate = state.dict(exclude_none=True)
  for field in dstate['sources']:
    field.pop('info')
  dstate.pop('info')

  # default config only adds float volumes, dB volumes are calculated
  for field in dstate['zones']:
    field.pop('vol')
  for field in dstate['groups']:
    field.pop('vol_delta')
  return dstate
```

Lohrer:
Just a note that I should have commented on: `web/generated` is no longer checked in to git (via the README). It is now generated if needed when creating the tmpfs. For cases where AmpliPi is ran as a test setup not on a Pi, `web/generated` is also created if needed in `app.py`.

The reason for this is that `git status` will show `web/generated/README.md` is missing after creating the tmpfs. The README was only used to force git to create the `web/generated` directory. So to avoid the annoyance of constantly having git show a modified file, I moved the README up a directory and simply generate `web/generated` as needed.

PR #363, opened 2022-08-11
No streams

### LINKED ISSUE #361: Bug: Unable to create streams when no streams exist
When attempting to add a new stream I'm consistently getting the following 404 error regardless of which type of stream I try to add:

```
{
  "detail": "create stream failed: max() arg is an empty sequence"
}
```

From some quick looking through the code, it looks like it might be coming from an error when building the stream object from create_stream here: https://github.com/micro-nova/AmpliPi/blob/main/amplipi/ctrl.py#L663 but I have not set up a full debug environment to verify.

linknum23:
Fixes #361

### DIFF
--- a/amplipi/ctrl.py
+++ b/amplipi/ctrl.py
@@ -397,11 +397,11 @@ class Api:
       Returns:
         'None' on success, otherwise error (dict)
     """
-    idx, src = utils.find(self.status.sources, sid)
-    if idx is not None and src is not None:
-      name, _ = utils.updated_val(update.name, src.name)
-      input_, input_updated = utils.updated_val(update.input, src.input)
-      try:
+    try:
+      idx, src = utils.find(self.status.sources, sid)
+      if idx is not None and src is not None:
+        name, _ = utils.updated_val(update.name, src.name)
+        input_, input_updated = utils.updated_val(update.input, src.input)
         # update the name
         src.name = str(name)
         if input_updated or force_update:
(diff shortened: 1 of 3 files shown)


amplipi/ctrl.py
```diff
@@ -662,7 +665,7 @@ def delete_group(self, gid: int) -> ApiResponse:
       return ApiResponse.error('delete group failed: {} does not exist'.format(gid))
 
   def _new_stream_id(self):
-    stream: Optional[models.Stream] = max(self.status.streams, key=lambda stream: stream.id)
+    stream: Optional[models.Stream] = max(self.status.streams, key=lambda stream: stream.id, default=None)
```
linknum23, line 668: This is the actual fix!

amplipi/ctrl.py
```diff
@@ -421,6 +421,9 @@ def set_source(self, sid: int, update: models.SourceUpdate, force_update: bool =
               other_src.input = ''
             stream.disconnect()
             stream.connect(idx)
+          elif src.input and 'stream=' in src.input:
```
linknum23, line 424: Secondary fix, We may want to try to actuate this code somewhere?

linknum23:
Minor change to make sure that we can handle adding a stream when no other streams exist. The bugfix was a single line of code. 

Main changes are to the test infrastructure to handle testing for this issue.

amplipi/ctrl.py
```diff
@@ -421,6 +421,9 @@ def set_source(self, sid: int, update: models.SourceUpdate, force_update: bool =
...
+            # TODO: should this stream id validation happen in the Source model?
+            raise Exception(f'StreamID specified by "{src.input}" not found')
```
linknum23, line 426: Secondary fix, We may want to try to actuate this code somewhere?

linknum23:
Alright fixed one more bug. I think this is good to go now!

linknum23:
Nevermind, looks like I unerthed another related bug

Lohrer:
I'm not sure what the new bug is but I played around with no streams and it seems fixed!

linknum23:
Dare I say this is actually fixed now?

Lohrer (approved):
Rebase on develop please!

PR #369, opened 2022-08-25
Only log to RAM

linknum23:
Significantly reduce logging to disk and make AmpliPi log almost exclusively to RAM. 

Some other additions:
- upgrade apt packages
- add text log view

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # AmpliPi Software Releases
 
+## 0.1.X Upcoming
+* Web App
+  * Add link for text log
+* API
+* Streams
+* Storage
+  * Only log to RAM, drastically reducing disk writes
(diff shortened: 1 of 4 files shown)


linknum23:
Stuff to test:

install: this with the deploy script

verify success,
- no logs being generated in /var/log/messages and /var/log/syslog
- apt in a good state and packages up to date
- ryslog daemon still running `systemctl status rsyslog` and `systemd status systemd-journald.service`
- check log link at the bottom of the amplipi app, should show a text log

Lohrer (approved):
The logs are still written to /tmp when clicking on the Logs link on the web UI, but I think that's reasonable since they're only written to disk if a request is made. With this PR the only thing from AmpliPi that's still writing to disk is the `house.json` config writes. Fix the little type then looks great!

PR #371, opened 2022-09-02
Power board tester support for high voltage units

Lohrer:
Added support for the 8- and 12-channel ADCs on the high-voltage AmpliPi units.


linknum23 (commented):
Looks good at a quick glance!

linknum23 (approved):

PR #374, opened 2022-09-08
Fix cmedia bringup in configure script

Lohrer:
At the very least the `f"sudo .{env['base_dir']}/scripts/reload_cmedia".split()` line needs to remove the `.` so that the path is absolute. The reload_cmeda script wasn't being found. But since I was looking at it, the reload_cmedia script isn't even necessary, so I cleaned up that section of the configure script.


linknum23 (approved):
Looks good. I assume this was tested on a fresh system. I do not have easy access to one,

PR #366, opened 2022-08-17
Use Spotifyd instead of vollibrespot to get latest librespot changes

klay2000:

PR #378, opened 2022-09-28
Add airplay2 build of shairport-sync for testing

linknum23:
Steps to add support for a single fully featured airplay2 stream
- [x] Add shairport-sync binary built for aiplay2
- [x] Add nqptp dependency (needs to be installed from source for now)
- [x] Test single airplay 2 stream with additional airplay 1 streams
- [ ] Add support for single airplay 2 stream, with the second stream either being rejected or converted to airplay 1 (depends on above)
- [x] Add metadata using MPRIS support from #366 (old metadata is broken with ap2 and already needed to be upgraded to support user control)


linknum23:
At the moment we have one working Airplay 2 Stream, without metadata support. Supporting a single airplay 2 stream is looking very promising. 

NOTE: Multiple Airplay 2 streams will require a significant effort and some experimentation to integrate docker audio streams into our ALSA configuration and is not part of this PR.

linknum23:
One option to support multiple Airplay 2 instances sans docker may be to use MVLAN in bridged mode, spinning up one virtual LAN per airplay 2 instance (or just having 4 default ones instantiated at boot time). Need to investigate how feasible this is and if multiple NQPTP servers can be run like this.

linknum23:
We should probably wrap this PR soon. I think there are a couple of routes to finish this, since we can only support 1 airplay2 stream right now:
- Add checking to only allow a single airplay2 instance at a time (unable to run additional airplay1 instances)
- Downgrade additional running airplay2 instances to airplay1 (we would probably use an additional flag to determine which streams want to be airplay2)
- Wait until we can support multiple aitplay2 instances

zetlen:
> We should probably wrap this PR soon. I think there are a couple of routes to finish this, since we can only support 1 airplay2 stream right now:
> 
> - Add checking to only allow a single airplay2 instance at a time (unable to run additional airplay1 instances)
> 
> - Downgrade additional running airplay2 instances to airplay1 (we would probably use an additional flag to determine which streams want to be airplay2)
> 
> - Wait until we can support multiple aitplay2 instances

Hey, I know this is an old comment and you may have already made a decision, but I strongly endorse releasing option 1 for now, and backlogging support for options 2 and 3. There's a huge advantage to having even one airplay 2 stream, for home theater users and for people trying to explain to family and friends why playing from iPhones is "all weird and slow". Thanks in advance, and thanks a million times for all your work up to this point.

linknum23:
We are actually working on a simple version of 2 right now with the plans of getting a prerelease out soon.

linknum23:
Closing this since this work is being finalized in [airplay1and2](https://github.com/micro-nova/AmpliPi/tree/airplay1and2)

PR #379, opened 2022-09-28
Enable support for m3u and other playlist files

### LINKED ISSUE #255: Feature: Add m3u file support to internet radio stream
Although vlc supports m3u file playback our stream implementation does not. What do we need to change?

AmpliPi's vlc support is pretty simple. Here's what you need to reproduce the problem on an AmpliPi:

``` bash
cd amplipi-dev
source venv/bin/activate
python streams/runvlc.py http://radiox.de/media/radiox.m3u --verbose
```

runvlc.py is a standalone script that we use to interface with vlc and get metadata back. If we are lucky adding some small changes there may add support for this and other formats.

linknum23:
closes #255

### DIFF
--- a/streams/runvlc.py
+++ b/streams/runvlc.py
@@ -79,9 +79,13 @@ def update_info(info) -> bool:
     log('Error: %s' % sys.exc_info()[1])
     return False
 
+# try to start url with special handling to detect failures
+# this allows us to verify the stream during creation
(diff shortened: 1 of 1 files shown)


linknum23:
In testing m3u files. We have found that we already support a subset of m3u* files. We really need to investigate this issue further before applying any sort of patch.

PR #390, opened 2022-10-17
Fix volume sliders load time behaviour

linknum23:
Volume sliders would either show 0 or 100% until background refresh and very laggy


linknum23:
@klay2000 can you test this an make sure it works as expected?

klay2000 (approved):
Works as expected.

PR #391, opened 2022-10-18
Config validation and loading

linknum23:
Minor changes to robustify config and stream serialization
- Minimize stream serialization, this only happens on a stream change now.
- Use serialization  changes to simplify the config init code as well
- Add recovery logic when stream used is not available
- Upgrade the config validation tests to catch any config loading errors

~~These changes were originally made to fix a bug we were seeing for configuration load problems,. Oddly we have not been able to reproduce the configuration load problems. That said this code simplifies state changes enough that we think it still makes sense to include it.~~ 
The bug was due to loading an unsupported stream that was connected to a source, This was fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # AmpliPi Software Releases
 
 ## 0.1.X Upcoming
+* Streams
+  * Robustify config loading to handle missing streams
+  * Simplify stream serialization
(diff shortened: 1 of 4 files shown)


linknum23:
Alright this is actually ready for review.

Lohrer (approved):

PR #395, opened 2022-10-21
Add LMS client

linknum23:
Add a client for LMS (Originally: Logitech Media Server) without meta data support. The typical user of this uses AmpliPi as a secondary component to the LMS ecosystem so metadata support is not needed.

We will plan on adding metadata support, controls, and volume following (if possible) in a separate PR in the future. 

Setting up the LMS server can be done on the Pi but we leave that as a exercise to the user. This simplifies our configuration space for now.


amplipi/streams.py
```diff
@@ -25,14 +25,16 @@
...
+from ctypes import c_ulong # unsigned ints for MAC generation
+from typing import Union, Optional
 import threading
 
 # Used by InternetRadio and Spotify
 import json
 import signal
 import ast
 import socket
+import hashlib # md5 for string -> MAC generation
```
klay2000, line 37: do you still need to import this?

linknum23: yes, its still used to generate the MAC from string. Hashlib appears to be part of the standard library tho.

klay2000 (approved):
Tested on Amplipi13, works as expected.

PR #399, opened 2022-11-08
Satisfy mypy checks for unspecified optional and untyped functions

linknum23:


linknum23:
I noticed the airplay2 branch was failing. This should fix the unrelated ci issues.

klay2000 (approved):

PR #404, opened 2022-11-15
Force usb soundcard volume and mute settings in configure script

Lohrer:
Fixes issue where aux in got muted.

### DIFF
--- a/amplipi/utils.py
+++ b/amplipi/utils.py
@@ -31,7 +31,7 @@ from typing import Dict, Iterable, List, Optional, Set, Tuple, TypeVar, Union
 
 import pkg_resources # version
 
-import amplipi.models as models
+from amplipi import models
 
 # pylint: disable=bare-except
 
@@ -48,11 +48,11 @@ def parse_int(i, options):
   """ Parse an integer into one of the given options """
   if int(i) in options:
     return int(i)
-  raise ValueError('{} is not in [{}]'.format(i, options))
+  raise ValueError(f'{i} is not in [{options}]')
 
 def error(msg):
   """ wrap the error message specified by msg into an error """
-  print('Error: {}'.format(msg))
+  print(f'Error: {msg}')
   return {'error': msg}
 
 VT = TypeVar("VT")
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -4,6 +4,7 @@ dasbus
 deepdiff
 fastapi
 fastapi_utils
+httpx
```
linknum23, line 7: What is this needed for? shlex?

linknum23 (commented):
I need to find a box to test this on, otherwise everything looks good!

Lohrer:
> I need to find a box to test this on, otherwise everything looks good!

You can use amplipi197 on my desk if someone hasn't stolen it!

linknum23:
Test case:
1. Add physical loopback from zone 1 preout to aux in
2. Play something out source 1 (which has zone 1 connected)
3. Connect an aux in stream on source 2 using the file player
4. Connect zone 2 to source 2 (un-muted) 
5. Verify audio is output
6. Mute the capture using the alsa tui
7.  Install this update using deploy
8.  After reboot set the zone 1 and 2  volumes, audio should play out the speaker connected to zone 2

linknum23:
It appears that the reboot is "muting" the input capture, pulse audio?

linknum23 (changes requested):
Changes need to persist after reboot

amplipi/utils.py
```diff
@@ -133,6 +134,19 @@ def available_outputs():
...
   return outputs
 
+def configure_inputs():
+  """ The IEC598 and Aux inputs are being muted/misconfigured during system startup
+
+    Enable the capture capability for Aux and IEC958 and set capture volume to 100% (unmuted) """
+  if is_amplipi():
+    # setup usb soundcard input volumes and unmute
+    try:
+      subprocess.run(shlex.split(r'amixer -D hw:cmedia8chint set Speaker 100% unmute'), check=True) # is this required??
```
Lohrer, line 144: I don't think I've ever seen the outputs get set to anything other than 100% by default, but I have seen them muted. Not sure if that was a fluke. We'll always want this state though, so I see no reason not to have this here to prevent any future problems.

Also, any reason this is a raw string?

linknum23: The sequence after the % was being highlighted in vscode, so I figured it would avoid any weird escape sequence caused by the %

linknum23:
Need to wait on mahdees feedback on this.

linknum23:
This works on the test setup

linknum23:
This was merged into the prerelease hot

PR #409, opened 2022-12-01
Peak detect

jonahshader:
When RCA is selected as a source, the green play icon on the source tab reflects the active state of that RCA.


amplipi/audiodetector/audiodetector.c
```diff
@@ -0,0 +1,354 @@
...
+#define SPI_CLOCK_HZ        3600000
+#define NUM_CHANNELS        8
+#define SAMPLES_PER_SECOND  10
+#define WINDOW_SIZE_SECONDS 3
+#define MICRO               1000000
+#define CH_ACTIVE_THR       10
+#define DEFAULT_BASELINE    550
+#define WINDOW_SIZE         (SAMPLES_PER_SECOND * WINDOW_SIZE_SECONDS)
+
+// #define DUMP_CSV
```
jonahshader, line 49: @linknum23 what are you thoughts on using #define for debugging in this way?

linknum23:
Alright I merged in #410, that should fix the unrelated CI issuses. Can you rebase on origin/develop when you get a chance?

amplipi/ctrl.py
```diff
@@ -437,22 +437,38 @@ def get_stream(self, src: Optional[models.Source] = None, sid: Optional[int] = N
       return self.streams.get(idx, None)
     return None
 
-  def _update_src_info(self, src):
+  def _get_rca_info(self, src: models.Source) -> models.SourceInfo:
+      # RCA, name mimics the steam's formatting
+    src_info = models.SourceInfo(img_url='static/imgs/rca_inputs.svg', name=f'{src.name} - rca', state='stopped')
+    playing = False
+    status_file = 'rca_status'
```
linknum23, line 444: we should probably move this file into a memory-backed path

amplipi/audiodetector/audiodetector.c
```diff
@@ -0,0 +1,354 @@
...
+    uint16_t ch3l[WINDOW_SIZE];
+    uint16_t ch3r[WINDOW_SIZE];
+    uint16_t ch2l[WINDOW_SIZE];
+    uint16_t ch2r[WINDOW_SIZE];
+    uint16_t ch1l[WINDOW_SIZE];
+    uint16_t ch1r[WINDOW_SIZE];
+    uint16_t ch0l[WINDOW_SIZE];
+    uint16_t ch0r[WINDOW_SIZE];
+  };
+  uint16_t vals[NUM_CHANNELS * WINDOW_SIZE];
```
linknum23, line 79: This can be made into a 2D array to make the compiler do the indexing.

amplipi/ctrl.py
linknum23: config/srcs/rcas might be a useful path

linknum23 (changes requested):

linknum23:
I played with some changes to the C code a bit with 2-d array aliasing in https://github.com/micro-nova/AmpliPi/tree/peak-detect-updates.

linknum23:
Also it looks like the input is toggling between playing and paused. We may need to make the window a little longer or add some other hysteresis.

amplipi/audiodetector/audiodetector.c
```diff
@@ -0,0 +1,354 @@
...
+  if (!success) {
+    fprintf(stderr, "Failed to create/open status file\n");
+    return EXIT_FAILURE;
+  }
+
+  AdcData baseline;  // Minimum values ever recorded (of window max)
+  for (size_t i = 0; i < sizeof(AdcData) / sizeof(baseline.vals[0]); i++)
+    baseline.vals[i] = DEFAULT_BASELINE;
+  AdcWindow window;
+  memset(&window, 0xFF, sizeof(AdcWindow));
```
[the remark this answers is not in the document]
jonahshader: max value is safest for the initial state of the window. we check if the max value across the window is above/below baseline + some offset. I guess memset was used instead of a for loop because the whole thing can just be initialized to 0xFF.

amplipi/audiodetector/audiodetector.c
```diff
@@ -0,0 +1,354 @@
...
+    data->vals[i] = (rx[i] & 0x300) + ((rx[i] >> 16) & 0xFF);  // rx[i] >> 8;
+  }
+
+  close(spi_fd);
+  return true;
+}
+
+void update_window(AdcWindow *window, AdcData *raw_data) {
+  // window is implemented like a circular queue
+  static size_t window_index = 0;
```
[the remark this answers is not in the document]
jonahshader: no one else needs window_index since we are treating each time step in the window indiscriminately (just taking the max across it). its only used on write. do you want it bundled with AdcWindow for the sake of separating state from logic?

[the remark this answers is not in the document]
linknum23: yeah, it just feels weird having coupled state in 2 places like that.

linknum23 (approved):

PR #414, opened 2022-12-14
Make RCA inputs stream-like

linknum23:
The RCA inputs are a special case right now. Lets just make them a special stream instead so that they fit in the normal paradigm. This should make our model a little simpler.

To make the upgrade process easy we also added a migration step that takes the source's name that was previously used/displayed as the RCA inputs name as the name for the corresponding stream.

This also adds the ability to disable a stream, which is most useful for not showing unused RCA inputs as an option in drop downs. The disable flag is purely cosmetic a disabled stream can still be played on a source. This would be useful for streams that are used in automations but not useful to the user. Whether or not a stream is disabled if it is playing it will be shown in a drop down. 

This is a step toward being able to remove the concept of sources.


amplipi/models.py
```diff
@@ -474,6 +474,8 @@ class Stream(Base):
   client_id: Optional[str] = Field(description='Plexamp client_id, becomes "identifier" in server.json')
   token: Optional[str] = Field(description='Plexamp token for server.json')
   server: Optional[str] = Field(description='Server url')
+  index: Optional[int] = Field(description='RCA index')
+  disbaled: Optional[bool] = Field(description="Soft disable use of this stream. It won't be shown as a selectable option")
```
Lohrer, line 478:
```suggestion
  disabled: Optional[bool] = Field(description="Soft disable use of this stream. It won't be shown as a selectable option")
```

Lohrer: I'm going to assume this means either disabling wasn't tested or this field is useless?

Lohrer (changes requested):

amplipi/models.py
linknum23: LoL it worked before! I think unused fields are just unvalidated?

linknum23:
This should be ready.

Lohrer (approved):

Issue #335, opened 2022-06-02
Migrate to 64-bit Raspberry Pi OS


Lohrer:
We need to migrate to a new OS at some  point to get plexamp support back, while we are doing this there are several other things to consider.
- [ ] test with PlexAmp support
- [ ] investigate using pulse audio or working around it
  - [ ] check audio pipes for dynamic stream switching
  - [ ] try using some streams through docker
- [ ] better update setup

Lohrer:
Tested running:
```
sudo apt update
sudo apt upgrade
sudo apt full-upgrade
sudo rpi-update
```
Audio playback (spotify to a single zone was tested) still works after reboot.

Then upgraded to a 64-bit kernel by putting `arm_64bit=1` in `/boot/config.txt`. Stuff still works! So now there's a 64-bit kernel and 32-bit system, and everything is up-to-date.

Finally upgraded to the latest Raspberry Pi OS (from Debian Bullseye), first changed buster->bullseye in `/etc/apt/sources.list` and `/etc/apt/sources.list.d/raspi.list`. Then:
```
sudo apt update
sudo apt upgrade
```
At this point I tried to run `sudo apt full-upgrade` but ran into the error
- libc6-dev : Breaks: libgcc-8-dev (< 8.4.0-2~) but 8.3.0-6+rpi1 is to be installed

Attempting to resolve:
```
sudo apt remove libgcc-8-dev
sudo apt full-upgrade
sudo apt autoremove
sudo apt autoclean
```
At this point AmpliPi stopped working... And the config script doesn't recognize that we're on an armv7 system. I think Python is confused because it reports the platform as aarch64 even though `dpkg --print-architecture` says it's armhf. I temporarily forced the config script to accept that we are on an arm system and ran `configure.py --os-deps --python-deps`. Now everything is happy and AmpliPi works again! I suspect the switch from Python3.7->Python3.9 messed up the virtual environment.

TODO: The system is still 32-bit.

stoppablemurph:
Forgive me if I sound dumb for asking, but what exactly is left to do here? I see "The system is still 32-bit", but I'm not sure what all would need to be done to resolve this.

linknum23:
The CM3 module has a 64-bit pi on it that we are running a 32-bit OS on.

There are 2 mains tasks to do a 64-bit upgrade:
- [ ] Test the AmpliPi software and of the existing 
[...]

PR #418, opened 2022-12-20
Add basic 64-bit support

linknum23:
This makes the installer work on a recent (bullseye) 64-bit pi


linknum23:
This only fulfills part of #335 as it does not provide any migration. However this will allow us to start testing.

Before this is merged we will need to update the docs to explain what is necessary for a 64-bit setup. @Lohrer what modifications were needed for the bootstrapping script?

Lohrer:
For Bullseye the biggest problem is the configure script fails to identify the platform as ARM. The current check is:
```python
if 'armv7l' in lplatform and 'debian' in lplatform
```

Buster x86 returned a string similar to `linux-5.10.103-v7+-armv7l-with-debian-10.13` for the platform name.
Bullseye x64 returns a string similar to `linux-5.15.76-v8+-aarch64-with-glibc2.31`

So there are two changes there, first 32-bit `armv7l` vs 64-bit `aarch64`. Second, `with-glibc` instead of `with-debian`.

The `bootstrap` branch also has a few tweaks and some work in progress on getting locale setup properly. We should also remove `piwiz` or otherwise disable it since it's sends an audible message until you complete the wizard in the GUI.

rtertiaer:
@linknum23 can you reopen this? Github automagically closed it when I deleted `develop` :woozy_face: sorry!

linknum23:
Looking at this again it is very like;y that these changes either no longer make sense or have been already incorporated. The only thing useful may be the builds of some 64-bit binaries.

PR #429, opened 2023-01-05
Bootstrap changes

Lohrer:
Lower GPU memory usage in config.txt, possibly fix boot config.txt copying and use UTC as timezone

### DIFF
--- a/config/boot_config.txt
+++ b/config/boot_config.txt
@@ -71,3 +71,6 @@ dtoverlay=spi2-2cs
 # Lock it so it doesn't scale down and lower SPI speed
 core_freq=400
 core_freq_min=400
+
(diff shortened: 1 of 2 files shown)


linknum23 (approved):
Looks good. Please add a PR for maging as well. This is really an internal tool so I would rather merge in the current state so it can be tested and refined instead of  wait for it to be perfect.

Lohrer:
My goal in bringing this back to a usable state was to make it so anyone could use it. Starting from a Raspberry Pi OS image means we don't have to host the image, plus (at least the initial) download is smaller.

linknum23:
These changes get us a lot closer to that goal.

PR #431, opened 2023-01-05
Invert RCA check

linknum23:
#414 introduced a bug that inverted the Analog/Digital multiplexer. This fixes the issue.

### DIFF
--- a/amplipi/ctrl.py
+++ b/amplipi/ctrl.py
@@ -352,7 +352,7 @@ class Api:
     """
     try:
       sid = int(sinput.replace('stream=',''))
-      return sid in RCAs
+      return sid not in RCAs
     except:
       return False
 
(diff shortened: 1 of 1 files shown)


amplipi/ctrl.py
```diff
@@ -352,7 +352,7 @@ def _is_digital(self, sinput: str) -> bool:
     """
     try:
       sid = int(sinput.replace('stream=',''))
-      return sid in RCAs
+      return sid not in RCAs
     except:
       return False
```
Lohrer, line 357: Can this return True as a default instead? I'm not sure what would cause an exception there but Digital is technically the safer default from a hardware perspective. Since the Analog RCA inputs are floating, random noise will be on any unconnected RCA inputs.

linknum23: Yup, that's what it should do!

Lohrer (approved):

PR #438, opened 2023-01-27
Add debug profile for web app

linknum23:
This should make debugging the web interface simpler


klay2000 (approved):
works, tested by running, adding breakpoint to ctrl.py, and moving a stream

Issue #332, opened 2022-06-01
Add the option to remove zones/groups from a source


linknum23:
We should be able to set a zone to not be connected to any source. There are a couple of different components to this:

- [x] Add the concept of being disconnected to a zone (source_id=-1)
- [x] #449
- [x] #446

PR #445, opened 2023-02-02
Cleanup and small changes

linknum23:
This integrates some cleanup and features added when testing virtual sources:
  * Add default icon for internet radio
  * Robustify config loading (handle stream failures and fix number of sources to hardware)
  * #332  (no app support yet)


linknum23:
Please put this on amplipi13 and see what happens. Try to break it.

amplipi/ctrl.py
```diff
@@ -482,6 +500,15 @@ def _update_src_info(self, src: models.Source):
     else:
       src.info = models.SourceInfo(img_url='static/imgs/disconnected.png', name='None', state='stopped')
 
+  def _get_source_config(self, sources: Optional[List[models.Source]]=None) -> List[bool]:
+    """ Convert the preamp's source configuration """
+    if not sources:
+      sources = self.status.sources
+    src_cfg = [True] * 4
+    for s, src in enumerate(sources):
+      src_cfg[s] = self._is_digital(sources[s].input)
```
klay2000, line 509: probably should be `src` instead of `sources[s]`

klay2000:
I hear popping when I restart the pi from the web app, that's what the muting was supposed to prevent, right?

Lohrer:
> I hear popping when I restart the pi from the web app, that's what the muting was supposed to prevent, right?

The firmware reset effectively mutes all amplified zones, but it doesn't mute the pre-outs. That's actually a good catch that I should have caught when I added muting to the pre-outs in the first place.

As for the popping, I'm not sure what would be causing that but I did add a potential popping reducing fix in the `reduce_popping` branch. If I remember correctly @mrahman313  heard some popping during testing but I never got any feedback as to whether the change I made actually helped or not.

amplipi/ctrl.py
```diff
@@ -585,17 +610,22 @@ def set_zone(self, zid, update: models.ZoneUpdate, force_update: bool = False, i
         zone.name = name
         zone.disabled = disabled
 
-        # any disabled zone should not be able to play anything, mute it to be sure
-        if zone.disabled and not mute:
+        # parse and check the source id
+        sid = utils.parse_int(source_id, range(models.SOURCE_DISCONNECTED, models.MAX_SOURCES))
+
+        # any zone disabled by source disconnection or a 'disabled' flag should not be able to play anything
+        implicit_mute = zone.disabled or sid == models.SOURCE_DISCONNECTED
```
Lohrer, line 617: It would be fairly trivial to support zones being physically disconnected from all sources in firmware, is that something we should add to the todo list or is this good enough?

amplipi/streams.py
```diff
@@ -782,10 +785,12 @@ def info(self) -> models.SourceInfo:
     loc = f'{src_config_folder}/currentSong'
     source = models.SourceInfo(name=self.full_name(),
                               state=self.state,
-                              img_url=self.logo,
+                              img_url='static/imgs/internet_radio.png',
```
Lohrer, line 788: It's probably fine, but this icon looks like it should be for FM radio, not internet radio.

tests/test_rest.py
```diff
@@ -498,6 +498,30 @@ def test_patch_zone_mute_disable(client, zid):
...
+def test_patch_zone_mute_disconnect(client, zid):
+  """ Unmute then disconnect a zone """
+  rv = client.patch('/api/zones/{}'.format(zid), json={'mute': False, 'vol_f': 0.5})
+  assert rv.status_code == HTTPStatus.OK
+  jrv = rv.json()
+  s = find(jrv['zones'], zid)
+  assert s is not None
+  assert s['mute'] == False
+  assert s['vol_f'] == 0.5
+  rv = client.patch('/api/zones/{}'.format(zid), json={'source_id': -1})
```
Lohrer, line 511: Should this use models.SOURCE_DISCONNECTED?

Lohrer (changes requested):

amplipi/ctrl.py
linknum23: Definitely add it to the todo list.

amplipi/ctrl.py
linknum23: It is limited to physical sources. I don't see a future where we have more sources than physical sources though.

tests/test_rest.py
linknum23: No but it probably needs a constant. The tests attempt to be almost completely separated.

amplipi/streams.py
linknum23: Hmm check out the fmradio icon: https://github.com/micro-nova/AmpliPi/blob/main/web/static/imgs/fmradio.png

linknum23:
> > I hear popping when I restart the pi from the web app, that's what the muting was supposed to prevent, right?
> 
> The firmware reset effectively mutes all amplified zones, but it doesn't mute the pre-outs. That's actually a good catch that I should have caught when I added muting to the pre-outs in the first place.
> 
> As for the popping, I'm not sure what would be causing that but I did add a potential popping reducing fix in the `reduce_popping` branch. If I remember correctly @mrahman313 heard some popping during testing but I never got any feedback as to whether the change I made actually helped or not.

Looking like this is fixed now!

Lohrer (approved):

PR #453, opened 2023-03-10
Make 'ap2' and 'disabled' fields updateable

linknum23:
@klay2000 can you test this in the UI and make sure changing these fiields works as expected?


klay2000 (approved):
LGTM, tested and the box actually stays ticked after hitting save now.

PR #455, opened 2023-03-14
Make mono to stereo conversion automatic (fixes fmradio)

### LINKED ISSUE #454: RTL-SDR source playing only from left channel
Only left channel plays the FM radio. Happens on all audio outputs except Source 1.

linknum23:
Closes #454.

To test a nearby radio station is 91.7, verify that stereo audio is output on each source.

### DIFF
--- a/streams/fmradio.py
+++ b/streams/fmradio.py
@@ -62,8 +62,15 @@ def main():
 
   update = False
 
+  # configure RTLSDR and and process FM radio signal
+  #  NOTE: FM decoding is typically done using -M wbfm which implies "-M fm -s 170k -A fast -r 32k -l 0 -E deemp" which sounds ok, but has a lot of noise
+  #  The additional parameters specified below sound much better in our experience
   rtlfm_args = f'rtl_fm -M fm -f {freq}M -s 171k -A std -p 0 -l 0 -E deemp -g 20 -F 9'.split()
+  # metadata processing using redsea
   redsea_args = ['redsea', '-u', '-p', '--feed-through']
+  # alsa output stage
+  # - convert mono output -> stereo using alsa routes chXm2s
+  # - normally mono conversion happens automatically but only the left channel was playing for ch1, ch2, and ch3
   aplay_args = ['aplay', '-r', '171000', '-f', 'S16_LE', '--device', f'{args.output}m2s']
 
   rtlfm_proc = subprocess.Popen(args=rtlfm_args, bufsize=1024, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
--- a/streams/fmradio.py
+++ b/streams/fmradio.py
@@ -68,9 +68,9 @@ def main():
   rtlfm_args = f'rtl_fm -M fm -f {freq}M -s 171k -A std -p 0 -l 0 -E deemp -g 20 -F 9'.split()
   # metadata processing using redsea
   redsea_args = ['redsea', '-u', '-p', '--feed-through']
-  # alsa output stage
-  # - convert mono output -> stereo using alsa routes chXm2s
-  # - normally mono conversion happens automatically but only the left channel was playing for ch1, ch2, and ch3
+  # ALSA output stage
+  # - rtl_fm output is mono
+  # - mono to stereo conversion happens automatically via ALSA plug module
   aplay_args = ['aplay', '-r', '171000', '-f', 'S16_LE', '--device', f'{args.output}']
 
   rtlfm_proc = subprocess.Popen(args=rtlfm_args, bufsize=1024, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


Lohrer:
If you'd rather just use the auto mono->stereo conversion that the `plug` plugin uses, we can add a route plugin in between plug and dmix. The current problem is plug thinks it's output is 8 channels, not 2, and I think that is messing with the auto conversions. So the current situation: [1/2]->plug->[8]->dmix. [N]=>N channel audio.

With the following setup we'll get: [1/2]->plug->[2]->route->[8]->dmix

```
pcm.ch1 {
    type                plug
    slave.pcm {
        type            route
        slave.pcm       "dmixer"
        slave.channels  8
        ttable.0.6      0.64
        ttable.1.7      0.64
    }
    slave.channels      2
}
pcm.ch2 {
    type                plug
    slave.pcm {
        type            route
        slave.pcm       "dmixer"
        slave.channels  8
        ttable.0.0      0.64
        ttable.1.1      0.64
    }
    slave.channels      2
}
pcm.ch3 {
    type                plug
    slave.pcm {
        type            route
        slave.pcm       "dmixer"
        slave.channels  8
        ttable.0.4      0.64
        ttable.1.5      0.64
    }
    slave.channels      2
}
```

linknum23:
Oddly this does not work with vlc on ch1 and ch3. Works fine for ch2.

linknum23:
With the extra layer in play here is what the verbose output -vvv  looks like from vlc:
```
[0204b6d8] alsa audio output debug: using ALSA device: ch1
[0204b6d8] alsa audio output debug:  Plug PCM: Route conversion PCM
  Transformation table:
    0 <- none
    1 <- none
    2 <- none
    3 <- none
    4 <- none
    5 <- none
    6 <- 0*0.64
    7 <- 1*0.64
```

Without the extra layer:
```
[0135d6d8] alsa audio output debug: using ALSA device: ch1
[0135d6d8] alsa audio output debug:  Plug PCM: Direct Stream Mixing PCM

```

Lohrer:
Not so odd I guess, ch2 uses physical channels 0 and 1 on the 8-channel CM6206, so no remapping is required. What is weird is that the FM radio stream (mono) and Spotify (stereo) both worked so I assumed the rest of the stereo streams would work...

My AmpliPi has been borrowed so I can't test, but I assume when you try ch3 you get:
```
Plug PCM: Route conversion PCM
  Transformation table:
    0 <- none
    1 <- none
    2 <- none
    3 <- none
    4 <- 0*0.64
    5 <- 1*0.64
    6 <- none
    7 <- none
```

Lohrer:
Those messages come from https://github.com/alsa-project/alsa-lib/blob/master/src/pcm/pcm_dmix.c and https://github.com/alsa-project/alsa-lib/blob/master/src/pcm/pcm_route.c and look exactly like what I'd expect. I think if there's an error it might lie elsewhere...

linknum23:
Another way to do this would be with 2 layers of plug:
```
pcm.ch1 {
    type                plug
    slave.pcm {
        type            plug
        slave.pcm       "dmixer"
        slave.channels  8
        ttable.0.6      0.64
        ttable.1.7      0.64
    }
    slave.channels      2
}
pcm.ch2 {
    type                plug
    slave.pcm {
        type            plug
        slave.pcm       "dmixer"
        slave.channels  8
        ttable.0.0      0.64
        ttable.1.1      0.64
    }
    slave.channels      2
}
pcm.ch3 {
    type                plug
    slave.pcm {
        type            plug
        slave.pcm       "dmixer"
        slave.channels  8
        ttable.0.4      0.64
        ttable.1.5      0.64
    }
    slave.channels      2
}
```

linknum23:
2 layers of plug works  and was tested with Pandora, Airplay, Spotify, Internet Radio, and FM Radio

linknum23 (commented):
I think this setup makes the most sense going forward since it handles any weird cases where mono audio will get played. We will need to continue testing this to make sure there aren't any performance penalties with the double plug setup.

linknum23:
@Lohrer Do you think this makes sense? 
Basic testing with 4 various streams playing seems to use about 15-20% for each stream which seems pretty reasonable.

streams/fmradio.py
```diff
@@ -62,9 +62,16 @@ def main():
...
 
+  # configure RTLSDR and and process FM radio signal
+  #  NOTE: FM decoding is typically done using -M wbfm which implies "-M fm -s 170k -A fast -r 32k -l 0 -E deemp" which sounds ok, but has a lot of noise
+  #  The additional parameters specified below sound much better in our experience
   rtlfm_args = f'rtl_fm -M fm -f {freq}M -s 171k -A std -p 0 -l 0 -E deemp -g 20 -F 9'.split()
+  # metadata processing using redsea
   redsea_args = ['redsea', '-u', '-p', '--feed-through']
-  aplay_args = ['aplay', '-r', '171000', '-f', 'S16_LE', '--device', '{}'.format(args.output)]
+  # alsa output stage
+  # - convert mono output -> stereo using alsa routes chXm2s
```
Lohrer, line 72: I assume this can be removed/updated?

Lohrer (changes requested):
Looks good! I just think there's a some comments that might need to reflect the latest changes.

Lohrer (approved):

PR #459, opened 2023-03-29
Aux Stream

### LINKED ISSUE #329: Add aux and optical inputs as streams
Support the analog (aux) and S/PDIF (optical) inputs in AmpliPi. Currently they can be manually added, see #244, but a way to switch between aux and optical needs to be created. Some discussion has already occurred on [Discourse](https://amplipi.discourse.group/t/cant-get-aux-in-working/73) as well.

klay2000:
Adds a stream for aux, closes #329.


linknum23 (commented):
Both "Optical Input" devices tested had noticeable buffering problems that caused the audio to skip every second or so. 

The fix appears to use the `alsaloop` utility to directly pipe the audio back. Something like the following:

    alsaloop -P ch0  -C plughw:cmedia8chint,0 -t 100000
   
That fixes the cache dropouts of vlc but has a different issue, that we previously saw when prototyping virtual-streams, a bus error failure caused by a bug in alsaloop that is fixed in more recent versions.

I think there are a couple of options here. We disable Optical for now with a note in the UI or wait until upgrading AmpliPi's OS to get the latest changes in `alsaloop`.

linknum23:
My vote is to disable optical (with an api failure when it is changed and a greyed out checkbox in the ui)

linknum23:
We really need to expedite 64-bit support to enable this.

rtertiaer:
@klay2000 if you intend to ship this eventually, can you re-target this PR to `main`? Github automagically closed it when I deleted `develop` :woozy_face: sorry!

linknum23:
Can you squash and rebase this on main?

amplipi/streams.py
```diff
@@ -973,6 +973,72 @@ def info(self) -> models.SourceInfo:
...
+    reconnect_needed = False
+    if 'disabled' in kwargs:
+      self.disabled = kwargs['disabled']
+    if 'name' in kwargs:
+      self.name = kwargs['name']
+    if reconnect_needed:
+      last_src = self.src
+      self.disconnect()
+      time.sleep(0.1) # delay a bit, is this needed?
+      self.connect(last_src)
```
linknum23, line 992: These LOC are unused

amplipi/streams.py
```diff
@@ -973,6 +973,72 @@ def info(self) -> models.SourceInfo:
...
+    self.bkg_thread.start()
+    return
+
+  def wait_on_proc(self):
+    """ Wait for the vlc process to finish """
+    if self.proc is not None:
+      self.proc.wait() # TODO: add a time here
+    else:
+      time.sleep(0.3) # handles mock case
+    self.state = 'stopped' # notify that the audio is done playing
```
linknum23, line 1026: wait on proc is used for announcements, I'm not sure how useful it is here. Realistically we don't need it but it may be useful for debugging whether or not the aux is still playing.

amplipi/streams.py
```diff
@@ -973,6 +973,72 @@ def info(self) -> models.SourceInfo:
...
+
+    if self.mock:
+      self._connect(src)
+      # make a thread that waits for a couple of secends and returns after setting info to stopped
+      self.bkg_thread = threading.Thread(target=self.wait_on_proc)
+      self.bkg_thread.start()
+      return
+
+    # Set input source
+    amixer_command = f'amixer -D usb71 set "PCM Capture Source" "Line"'
```
linknum23, line 1006: My isntincts are that we should be using the hw:cmedia8chint audio device created by our udev rules. Our tests have a similar command string in them:
```
# Ensure the Aux Input (Line on the CMedia chip) is selected for capture.
subprocess.run(shlex.split('amixer -D hw:cmedia8chint set "PCM Capture Source",0 Line'),

```

linknum23: We should probably add a utils.py function for this, maybe something like `enable_aux_input`.  utils.py is where most of this type of functionality is kept.

amplipi/streams.py
```diff
@@ -973,6 +973,72 @@ def info(self) -> models.SourceInfo:
...
+    subprocess.check_call(amixer_command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    # Start audio via runvlc.py
+    vlc_args = f'cvlc -A alsa --alsa-audio-device {utils.real_output_device(src)} alsa://plughw:cmedia8chint,0 vlc://quit'
+    print(f'running: {vlc_args}')
+    self.proc = subprocess.Popen(args=vlc_args.split(), stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    self._connect(src)
+    # make a thread that waits for a couple of secends and returns after setting info to stopped
+    self.bkg_thread = threading.Thread(target=self.wait_on_proc)
+    self.bkg_thread.start()
```
linknum23, line 1017: As discussed below this is probably not needed

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "amplipi"
-version = "0.3.1"
+version = "0.3.0+4cf6490-aux-optical-stream-dirty"
```
linknum23, line 3: Please clean this out.

amplipi/ctrl.py
```diff
@@ -204,6 +204,16 @@ def reinit(self, settings: models.AppSettings = models.AppSettings(), change_not
...
 
+    # make sure the config file contains the aux stream
+    has_stream = False
+    for s in self.status.streams:
+      if s.type == "aux":
+        has_stream = True
+        break
+
+    if not has_stream:
+      self.create_stream(models.Stream(type="aux", name="Aux"))
```
linknum23, line 215:
```suggestion
    has_aux_stream = False
    for s in self.status.streams:
      if s.type == "aux":
        has_aux_stream = True
        break

    if not has_aux_stream:
      self.create_stream(models.Stream(type="aux", name="Aux"))
```

linknum23 (changes requested):
Can you clean this up a bit and then squash the changes into a single commit?

amplipi/streams.py
```diff
@@ -1006,6 +1006,52 @@ def info(self) -> models.SourceInfo:
...
+
+  def connect(self, src):
+    """ Connect a short run VLC process with audio output to audio source """
+    print(f'connecting {self.name} to {src}...')
+
+    if self.mock:
+      self._connect(src)
+      # make a thread that waits for a couple of secends and returns after setting info to stopped
+      self.bkg_thread = threading.Thread(target=self.wait_on_proc)
+      self.bkg_thread.start()
```
linknum23, line 1029:
```suggestion
```

amplipi/streams.py
```diff
@@ -1006,6 +1006,52 @@ def info(self) -> models.SourceInfo:
...
+    print(f'running: {vlc_args}')
+    self.proc = subprocess.Popen(args=vlc_args.split(), stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    self._connect(src)
+    return
+
+  def disconnect(self):
+    if self._is_running():
+      self.proc.kill()
+      if self.bkg_thread:
+        self.bkg_thread.join()
```
linknum23, line 1046:
```suggestion
```

amplipi/streams.py
```diff
@@ -1006,6 +1006,52 @@ def info(self) -> models.SourceInfo:
...
+
+    if self.mock:
+      self._connect(src)
+      # make a thread that waits for a couple of secends and returns after setting info to stopped
+      self.bkg_thread = threading.Thread(target=self.wait_on_proc)
+      self.bkg_thread.start()
+      return
+
+    # Set input source
+    utils.enable_aux_input()
```
linknum23, line 1033: Should we handle the exception here?

amplipi/streams.py
```diff
@@ -1006,6 +1006,52 @@ def info(self) -> models.SourceInfo:
     source.track = "Not currently supported"
     return source
 
+class Aux(BaseStream):
+  """ A stream to play from the aux input. """
+  def __init__(self, name: str, disabled: bool = False, mock: bool = False):
+    super().__init__('aux', name, disabled=disabled, mock=mock)
+    self.bkg_thread = None
```
linknum23, line 1013:
```suggestion
```

linknum23:
No change?

rtertiaer (changes requested):
can you make note of the changes here in the `CHANGELOG`?

amplipi/ctrl.py
```diff
@@ -944,7 +955,7 @@ def delete_stream(self, sid: int, internal=False) -> ApiResponse:
     """Deletes an existing stream"""
     try:
       # RCA streams are intrinsic to the hardware and can't be removed
```
Lohrer, line 957:
```suggestion
      # Analog streams are intrinsic to the hardware and can't be removed
```

amplipi/streams.py
```diff
@@ -1006,6 +1006,46 @@ def info(self) -> models.SourceInfo:
...
+    super().__init__('aux', name, disabled=disabled, mock=mock)
+
+  def reconfig(self, **kwargs):
+    if 'disabled' in kwargs:
+      self.disabled = kwargs['disabled']
+    if 'name' in kwargs:
+      self.name = kwargs['name']
+
+  def connect(self, src):
+    """ Connect a short run VLC process with audio output to audio source """
```
linknum23, line 1021:
```suggestion
    """ Use vlc to connect audio output to audio source """
```

linknum23 (approved):
I haven't retested this, but all of the test updates make sense.

web/src/utils/getIcon.jsx
```diff
@@ -8,6 +8,7 @@ import plexamp from "@/../static/imgs/plexamp.png";
 import lms from "@/../static/imgs/lms.png";
 import internetradio from "@/../static/imgs/internet_radio.png";
 import rca from "@/../static/imgs/rca_inputs.jpg";
+import aux from "@/../static/imgs/aux_input.png";
```
Lohrer, line 11: Why not just use the .svg here? Then we can delete the .png.

web/static/imgs/aux_input.svg
Lohrer, line 1: The black-on-grey here is very hard to see, can we add color or something?

rtertiaer (approved):
thx!

klay2000:
@Lohrer lincoln and I bounced back and forth on the icon, take a look at what we came up with

PR #460, opened 2023-03-31
Streamer Software

klay2000:
Changes for supporting streamer box, eeprom(s), and eink display.


linknum23 (changes requested):
you will need to add support for the streamer config in app.load_factory_config right now it defaults to a config with 6 zones. 

Michael will review the changes to the display code.

Lohrer (changes requested):
First off, rebase on develop.

Things to fix that I've noticed so far:
* RCA inputs shouldn't be available.
* Factory reset config adds zones.
* Updates fail due to not being able to program the preamp firmware.

Minor issue that can be ignored or fixed later:
* Zones still show up in the API documentation page.
* The example preset "Mute Some" returns "TypeError: response.presets is undefined"
* Mute all still occasionaly gets added (not sure by what) as a preset (which fails to run properly).
* The RCA audiodetector will be installed (note, there's currently a bug where a deploy won't install the audiodetector but the updater will)
* Shortcuts to tests that won't work are still placed on the Desktop.
* `Error: no preamps found`. AmpliPi still tries to connect to the non-existent preamp. I think enough checks are in place that no attempted I2C messages are sent but we should probably be using the "Mock" preamp.

Open questions:
* Keep amplipi hostname?
* I get bluetooth errors at startup even though no bluetooth streams are configured. It's possible this is fixed by rebasing on develop which has bluetooth updates.
```
/usr/bin/bluetoothctl
Error checking for bluetooth hardware: Command '['bluetoothctl', 'show']' timed out after 0.5 seconds
bluetooth dongle unavailable
```

linknum23:
@klay2000 Can you look at the changes I made to the display and verify that they make sense?

jonahshader (commented):
No major concerns, looks good

linknum23 (approved):

PR #461, opened 2023-04-13
Lms metadata 2

SteveMicroNova:
Draft of LMS Metadata, breaks sometimes but prepared to be tested


requirements.txt
```diff
@@ -1,35 +1,49 @@
-adafruit-circuitpython-rgb-display==3.11.0
```
linknum23, line 1: This is a lot of changes to a file I would expect minimal changes from.

amplipi/streams.py
```diff
@@ -1077,21 +1080,27 @@ def connect(self, src):
 
       self.proc = subprocess.Popen(args=lms_args)
       self._connect(src)
+
+      self.metadata_reader.connect()
```
linknum23, line 1084: This metadata reader needs to be launched as a separate process since connect is expected to return immediately after the LMS stream is started.

The separate process is what will be writing to the metadata file.

streams/lms_metadata.bash
```diff
@@ -0,0 +1,8 @@
+SERVER="192.168.0.177"
```
linknum23, line 1:
```suggestion
#TODO: remove this file
SERVER="192.168.0.177"
```

streams/lms_metadata.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/python3
+"""LMS Metadata"""
```
linknum23, line 2: This needs to be refactored to run independently of AmpliPi as a python script

linknum23 (changes requested):

streams/find_server
linknum23, line 1: Just a heads up this will end up being stored in bin/ARCH/ and get installed by the installer

streams/lms_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+      f = open(f"lms_{str(self.player_name).replace(' ', '_')}_metadata.json", 'wt', encoding='utf-8')
+      try:
+        fcntl.flock(f, fcntl.LOCK_EX)
+        json.dump(meta, f, indent = 2)
+      finally:
+        fcntl.flock(f, fcntl.LOCK_UN)
+        f.close()
+
+      if self.dump:
+        f = open(f"{str(self.player_name).replace(' ', '_')}_track_raw.json", "w")
```
linknum23: Linter suggestion: use python with syntax here

streams/lms_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+import subprocess
+from typing import Optional
+
+class LMSMetadataReader:
+  """A class for getting metadata from a Logitech Media Server."""
+
+  # meta_ref is probably an unneccessary variable to pass as an arg since it's obscured from the user, but we can eventually make it an optional setting for the user
+  def __init__(self, name: str, meta_ref: Optional[int] = 2, dump: Optional[bool] = False):
+    self.player_name = name
+    self.IP = None
```
linknum23, line 19: all caps would typically indicate a constant value. please make self.IP lowercase here

streams/lms_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+
+  def connect(self):
+    """Discovers LMS Player and then requests metadata repetitively"""
+    f = open(f"lms_{str(self.player_name).replace(' ', '_')}_metadata.json", 'wt', encoding='utf-8')
+    json.dump({'track': 'Loading...', 'artist': 'Loading...', 'album': 'Loading...', 'image_url': 'static/imgs/lms.png'}, f, indent = 2)
+    f.close()
+    x = 0
+
+    # Much faster method of connecting to the metadata server using code from: https://github.com/ralph-irving/squeezelite/blob/master/tools/find_servers.c
+    IPFind = subprocess.run('streams/find_server', capture_output=True, text=True)
```
linknum23, line 32: How should find_server failures be handled here? Maybe some retries?

streams/lms_metadata.py
```diff
@@ -0,0 +1,145 @@
...
+            meta["artist"] = song_data["artist"]
+            meta["album"] = song_data["album"]
+            meta["image_url"] = song_data["artwork_url"]
+          except KeyError:
+            # Sometimes, KeyError will occur when switching to Spotify/Pandora from a different stream type since the json that LMS sends is formatted differently
+            print(f"KeyError, trying again in {self.meta_ref_rate} seconds...")
+
+            meta["track"] = song_data["title"]
+            meta["image_url"] = song_data["artwork_url"]
+        # File locking so to reduce errors, without locks here and on the read cycle you can sometimes read while writing, which will read an empty file and crash the stream
```
linknum23, line 115: I don't think the file locking stuff should be needed here, right?

streams/lms_metadata.py
```diff
@@ -0,0 +1,145 @@
...
+      except:
+        # When first creating an LMS stream, there can be random errors that will close the while loop
+        # typically when asking the player for info when there isn't a player linked to the stream yet
+        print("FAIL")
+        pass
+
+    while self.connected:
+      try:
+        track_json = {"id": 1, "method": "slim.request", "params": [ self.player_name, ["status", "-",100] ]}
+        track_info = requests.post(f'http://{self.IP}:9000/jsonrpc.js 2>/dev/null', json=track_json, timeout=200)
```
linknum23, line 60: Does it make sense to make a method for the json request response handling to a fixed endpoint?

linknum23: Also: I would suggest refactoring all of this code to use a variable that store the server base url: "http://{self.IP}:9000"

streams/lms_metadata.py
```diff
@@ -0,0 +1,145 @@
...
+    IPFind = subprocess.run('streams/find_server', capture_output=True, text=True)
+    # Uses re.search because find_server.c spits out as '{Hostname}:{port} ({IP})', so I scrape the data from inbetween the parentheses to get the IP
+    self.IP = re.search(r"\((.*?)\)", IPFind.stdout).group(1)
+
+
+    # When not connected, search for player to connect to by the proper name
+    while not self.connected:
+      try:
+        player_json = {"id": 1,	"method": "slim.request",	"params": [self.player_name, ["players", "-", 100, "playerid"]]}
+        player_info = requests.get(f'http://{self.IP}:9000/jsonrpc.js', json=player_json, timeout=10)
```
linknum23, line 41: why is a different timeout used here?

streams/lms_metadata.py
linknum23: Should this be integrated in a loop of everything?

streams/lms_metadata.py
linknum23: I typically would name this ip_addr

streams/lms_metadata.py
```diff
@@ -0,0 +1,145 @@
...
+    """Discovers LMS Player and then requests metadata repetitively"""
+    f = open(f"lms_{str(self.player_name).replace(' ', '_')}_metadata.json", 'wt', encoding='utf-8')
+    json.dump({'track': 'Loading...', 'artist': 'Loading...', 'album': 'Loading...', 'image_url': 'static/imgs/lms.png'}, f, indent = 2)
+    f.close()
+    x = 0
+
+    # Much faster method of connecting to the metadata server using code from: https://github.com/ralph-irving/squeezelite/blob/master/tools/find_server.c
+    IPFind = subprocess.run('streams/find_server', capture_output=True, text=True)
+    # Uses re.search because find_server.c spits out as '{Hostname}:{port} ({IP})', so I scrape the data from inbetween the parentheses to get the IP
+    self.IP = re.search(r"\((.*?)\)", IPFind.stdout).group(1)
```
linknum23, line 34: What happens when this regex fails?

streams/lms_metadata.py
```diff
@@ -0,0 +1,145 @@
...
+        song_json = {"id":2,"method":"slim.request","params":[ self.player_name, ["songinfo","-",100,f"track_id:{track_id}"]]}
+        song_info = requests.post(f'http://{self.IP}:9000/jsonrpc.js 2>/dev/null', json=song_json, timeout=200)
+        song_load = json.loads(song_info.text)
+
+        x = 0
+        song_data = {}
+        for item in song_load['result']['songinfo_loop']:
+          for info in item:
+            song_data[info] = song_load['result']['songinfo_loop'][x][info]
+          x += 1
```
linknum23, line 73: this could be replaced with something like:
```python
  def flatten(lms_info: dict) -> dict:
    flat_data = {}
    for item in lms_info:
      for info in item:
        flat_data[info] = lms_info[x][info]
      x += 1
    return flat_data
```

streams/find_server.c
linknum23, line 1: We will need to add the compilation of this binary to the build script, probably rename this to find_lms_server.c for context and add a blurb stating this file is from squeezelite with agithub reference.

streams/lms_metadata.py
```diff
@@ -16,29 +16,34 @@ class LMSMetadataReader:
...
-    IPFind = subprocess.run('streams/find_server', capture_output=True, text=True)
-    # Uses re.search because find_server.c spits out as '{Hostname}:{port} ({IP})', so I scrape the data from inbetween the parentheses to get the IP
-    self.IP = re.search(r"\((.*?)\)", IPFind.stdout).group(1)
-
 
     # When not connected, search for player to connect to by the proper name
     while not self.connected:
       try:
+        # Much faster method of connecting to the metadata server using code from: https://github.com/ralph-irving/squeezelite/blob/master/tools/find_server.c
+        ip_find = subprocess.run(['streams/find_server'], capture_output=True, text=True)
```
linknum23: This doesn't appear to handle the case where there are multiple lms servers but the client is only connected to one. For instance you may want to try connecting the lms client to one of the other lms servers (not yours). Is there a way to search for just the server this cliient is connected to (with its name or mac?).

streams/lms_metadata.py
```diff
@@ -16,29 +16,34 @@ class LMSMetadataReader:
   # meta_ref is probably an unneccessary variable to pass as an arg since it's obscured from the user, but we can eventually make it an optional setting for the user
   def __init__(self, name: str, meta_ref: Optional[int] = 2, dump: Optional[bool] = False):
     self.player_name = name
-    self.IP = None
+    self.locale = None # locale replaces IP, it is a concatenation of both the IP and the port in {IP}:{port} format
```
linknum23: maybe just call this server? locale doesn't mean much 
A better comment might be just # {ip}:{port}

streams/lms_metadata.py
```diff
@@ -16,29 +16,34 @@ class LMSMetadataReader:
...
+        # Much faster method of connecting to the metadata server using code from: https://github.com/ralph-irving/squeezelite/blob/master/tools/find_server.c
+        ip_find = subprocess.run(['streams/find_server'], capture_output=True, text=True)
+        print(f'STDOUT: {ip_find.stdout}')
+        # Uses re.search because find_server.c spits out as '{Hostname}:{port} ({IP})', so I scrape the data from inbetween the parentheses to get the IP
+        ip = re.search(r'\((.*?)\)', ip_find.stdout).group(1)
+        print(f'IP: {ip}')
+        port = re.search(r':(\d{4})', ip_find.stdout).group(1)
+        print(f'PORT: {port}')
+        self.locale = f"{ip}:{port}"
+
```
linknum23, line 54: i would hide the stdout printout under a debug flag specified by the command line like
```
if self.debug:
   print(f'find_server output: {ip_find.stdout}')
```
condense the rest of the printouts to something like print('Found LMS server: {ip}:{port}')

streams/lms_metadata.py
```diff
@@ -48,21 +53,22 @@ def connect(self):
             self.connected = True
           else:
             time.sleep(0.1)
-      except:
+      except Exception as e:
         # When first creating an LMS stream, there can be random errors that will close the while loop
         # typically when asking the player for info when there isn't a player linked to the stream yet
-        print("FAIL")
+        print(f"FAIL: {e}")
```
linknum23: To me FAIL would indicate some fatal error that would make this script crash. Is there a more useful message? Should this only print when the debug flag is passed?

amplipi/streams.py
```diff
@@ -1104,13 +1104,13 @@ def disconnect(self):
...
     # Opens and reads the metadata.json file every time the info def is called
     # uses a file lock so that the file cannot be read while writing and vice/versa. If you let those processes run into eachother, there are errors
-    try:
-      meta_read = open(f"lms_{str(self.name).replace(' ', '_')}_metadata.json", "r", encoding="utf-8")
-      fcntl.flock(meta_read, fcntl.LOCK_EX)
-      self.meta = json.loads(meta_read.read())
-    finally:
-      fcntl.flock(meta_read, fcntl.LOCK_UN)
-      meta_read.close()
+    with open(f"lms_{str(self.name).replace(' ', '_')}_metadata.json", "r", encoding="utf-8") as meta_read:
```
linknum23: This does not handle any file i/o errors. I would expect an i/o error to be suppressed in this method and for the info to be returned to be the default lms info or something about being unable to connect to LMS server.

streams/lms_metadata.py
```diff
@@ -113,29 +119,26 @@ def connect(self):
...
 
       # a sleep equal to the meta_ref_rate, that way the metadata refreshes on a set schedule while looping instead of just doing it at all times always
       time.sleep(self.meta_ref_rate)
 
 if __name__ == '__main__':
-  parser = argparse.ArgumentParser(description='LMS Metadata')
+  parser = argparse.ArgumentParser(description="LMS Metadata Reader - a script for finding an LMS player with a given name and extracting the name of the song, album, and artist as well as getting the album picture")
   parser.add_argument('--name', type=str, required=True, help='The name of the LMS Player')
   parser.add_argument('--ref', type=int, default=2, help='The frequency of metadata refresh cycles')
   parser.add_argument('--dump', action='store_true', help="""Create raw json dumps directly from the LMS server,
```
linknum23, line 141: maybe call this debug and use it to enable some of the extra printouts as well/?

streams/lms_metadata.py
```diff
@@ -113,29 +119,26 @@ def connect(self):
...
             meta["image_url"] = song_data["artwork_url"]
         # File locking so to reduce errors, without locks here and on the read cycle you can sometimes read while writing, which will read an empty file and crash the stream
-        f = open(f"lms_{str(self.player_name).replace(' ', '_')}_metadata.json", 'wt', encoding='utf-8')
-        try:
-          fcntl.flock(f, fcntl.LOCK_EX)
-          json.dump(meta, f, indent = 2)
-        finally:
-          fcntl.flock(f, fcntl.LOCK_UN)
-          f.close()
+        with open(f"lms_{str(self.player_name).replace(' ', '_')}_metadata.json", 'wt', encoding='utf-8') as f:
```
linknum23: Using flock here will make the info process block and cause the server to have delays. To avoid this read/write problem it looks like the standard way is to write to a temporary file and then move the temporary file to the actual file name. When this happens the reading process continues reading the old file until it is closed.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "amplipi"
-version = "0.2.1"
+version = "0.2.1+fc5f368-lms_metadata_2-dirty"
```
linknum23, line 3: Looks like the deploy script left this change here. This file should not be changed.

requirements.txt
```diff
@@ -23,6 +22,42 @@ pytest-dependency==0.5.1
 python-multipart==0.0.5
 python-vlc==3.0.18121
 PyYAML==6.0
```
linknum23, line 24: Just a reminder this file has a lot of changes. Can you revert this file to the old one and see what is actually needed?

linknum23: You can do this with: `git checkout develop requirements.txt`

amplipi/streams.py
```diff
@@ -1073,25 +1079,56 @@ def connect(self, src):
...
+        'artist': 'Error: Could Not Find LMS Server',
+        'image_url': 'static/imgs/lms.png'
+      }
+
     source = models.SourceInfo(
       name=self.full_name(),
       state=self.state,
-      img_url='static/imgs/lms.png',
-      track='check LMS for song info',
+      img_url= self.meta['image_url'],
```
linknum23, line 1128: there is no garauntee that all of those fields exist. What if album is not populated into self.meta?

linknum23: How about something like `self.meta.get('track', '')`

amplipi/streams.py
```diff
@@ -1073,25 +1079,56 @@ def connect(self, src):
           # NOTE: port 9000 is assumed
           server.replace('localhost', socket.gethostname())
         lms_args += [ '-s', server]
+
+        meta_args = ['python3', 'streams/lms_metadata.py', '--name', self.name, "--client", self.server]
```
linknum23, line 1083: I would assume the parameter would be --server this is a little confusing...

streams/lms_metadata.py
```diff
@@ -0,0 +1,160 @@
...
+            json.dump(song_load, f, indent = 2)
+      except Exception as e:
+        print(f"Error: {e}, trying again in {self.meta_ref_rate} seconds...")
+
+      time.sleep(self.meta_ref_rate)
+
+if __name__ == '__main__':
+  parser = argparse.ArgumentParser(description="LMS Metadata Reader - a script for finding an LMS player with a given name and extracting the name of the song, album, and artist as well as getting the album picture")
+  parser.add_argument('--name', type=str, required=True, help='The name of the LMS Player')
+  parser.add_argument('--client', type=str, help='The name of the client running the LMS server', metavar="SERVER_ADDRESS")
```
linknum23, line 154: I would expect this parameter to be server

linknum23: client is confusing here because the lms clients are running on amplipi

linknum23, line 154:
```suggestion
  parser.add_argument('--server', type=str, help='The hostname of the server running the LMS server', metavar="HOSTNAME")
```

streams/lms_metadata.py
```diff
@@ -0,0 +1,160 @@
...
+import subprocess
+import requests
+
+class LMSMetadataReader:
+  """A class for getting metadata from a Logitech Media Server."""
+
+  # meta_ref is probably an unneccessary variable to pass as an arg since it's obscured from the user, but we can eventually make it an optional setting for the user
+  def __init__(self, name: str, client: Optional[str] = None, meta_ref: Optional[int] = 2, debug: Optional[bool] = False):
+    self.player_name = name
+    self.client = client
```
linknum23, line 19: maybe server and resolved_server? or server_host and server_addr

streams/lms_metadata.py
```diff
@@ -0,0 +1,160 @@
...
+          try:
+            meta["track"] = track_data["title"]
+            meta['artist'] = None
+            meta['album'] = song_data['remote_title']
+            meta["image_url"] = f"http://{self.server}/music/{song_data['coverid']}/cover.jpg?id={song_data['coverid']}"
+          except KeyError:
+            # Sometimes, KeyError will occur when switching between streams with different metadata formatting
+            print(f"KeyError, trying again in {self.meta_ref_rate} seconds...")
+
+            meta["track"] = song_data["title"]
```
linknum23, line 119: I would suggest fully updating meta in the case you don't update all of the fields to avoid having lingering old values intermixed with new data. You could use a data class to handle this .

streams/find_lms_server
linknum23, line 1: This file should not be included. The correct binary is copied here at install time.

linknum23 (changes requested):

streams/lms_metadata.py
```diff
@@ -0,0 +1,173 @@
...
+        track_id = track_load["result"]["playlist_loop"][0]["id"]
+        song_json = {"id":2,"method":"slim.request","params":[ self.player_name, ["songinfo","-",100,f"track_id:{track_id}"]]}
+        song_info = requests.post(f'http://{self.address}/jsonrpc.js', json=song_json, timeout=1000)
+        song_load = json.loads(song_info.text)
+
+        song_data = self.flatten(song_load['result']['songinfo_loop'])
+        track_data = self.flatten(track_load['result']['playlist_loop'])
+
+
+        # Previously this code was a series of try-catches and if-elses
```
linknum23, line 132: This type of comment doesn't belong in the final code

streams/lms_metadata.py
```diff
@@ -0,0 +1,173 @@
...
+          self.meta.artist = song_data.get('artist')
+          self.meta.album = song_data.get('album')
+          self.meta.track = song_data.get('title')
+          self.meta.image_url = song_data.get('artwork_url')
+        else:
+          self.meta.artist = song_data.get('title')
+          self.meta.album = song_data.get('remote_title')
+          self.meta.track = track_data.get('title')
+          self.meta.image_url = 'static/imgs/lms.png'
+          if song_data.get('coverid'):
```
linknum23, line 144: You may want to split up the cover art handling from the other metadata. What happens if artist, album, and title are populated by the file is hosted locally? What happens if the artwork_url isnt provided but artist is?

streams/lms_metadata.py
```diff
@@ -0,0 +1,173 @@
...
+  """A class for getting metadata from a Logitech Media Server."""
+
+  # meta_ref is probably an unneccessary variable to pass as an arg since it's obscured from the user, but we can eventually make it an optional setting for the user
+  def __init__(self, name: str, server: Optional[str] = None, meta_ref: Optional[int] = 2, debug: Optional[bool] = False):
+    self.player_name = name
+    self.server = server
+    self.address = None # {ip}:{port}
+    self.meta_ref_rate = meta_ref
+    self.debug = debug
+    if server:
```
linknum23, line 51: Lots or duplication here, maybe use a variable to store the track message?

linknum23: Or better yet reinitialize album and track if server is not present

streams/lms_metadata.py
linknum23: As far as I can tell this is not resolved

amplipi/streams.py
```diff
@@ -1073,25 +1081,58 @@ def connect(self, src):
...
+        self.meta_proc = subprocess.Popen(args=meta_args)
+      else:
+        meta_args = ['python3', 'streams/lms_metadata.py', '--name', self.name]
+        self.meta_proc = subprocess.Popen(args=meta_args)
       # TODO: allow port to be specified with server (embedding it in the server URL does not work)
 
       self.proc = subprocess.Popen(args=lms_args)
       self._connect(src)
+
+      # TODO: Add metadata support? This may have to watch the output log?
```
linknum23, line 1095: This TODO can be safely removed now

linknum23:
All of the changes made make sense, this still needs real world testing.

SteveMicroNova:
This has been rebased and merge conflicts have been solved, I'd like someone to take a look at it before I go merging it to dev though

All the things that are failing are in files that have nothing to do with this PR, so those may be preexisting issues or evidence of a faulty merge

amplipi/streams.py
```diff
@@ -1152,15 +1152,15 @@ def reconfig(self, **kwargs):
...
+  def _activate(self, vsrc: int):
     """ Connect a squeezelite output to a given audio source
     This will create a LMS client based on the given name
     """
     f = open(f"lms_{str(self.name).replace(' ', '_')}_metadata.json", "w", encoding="UTF-8")
     json.dump(self.meta, f, indent = 2)
     f.close()
     if self.mock:
-      self._connect(src)
+      self._connect(vsrc)
```
linknum23, line 1163: Please revert this change, connect should not be called directly by a PersistantStream.

amplipi/streams.py
```diff
@@ -1199,19 +1199,17 @@ def connect(self, src):
       # TODO: allow port to be specified with server (embedding it in the server URL does not work)
 
       self.proc = subprocess.Popen(args=lms_args)
-      self._connect(src)
-
-
+      self._connect(vsrc)
```
linknum23, line 1202: Please revert this change, connect should not be called directly by a PersistantStream.

amplipi/streams.py
```diff
@@ -1199,19 +1199,17 @@ def connect(self, src):
...
       print(f'error starting lms: {exc}')
 
   def _deactivate(self):
     if self._is_running():
       self.proc.kill()
-    self._disconnect()
-    if self.meta_proc is not None:
-      self.meta_proc.terminate()
-      self.meta_proc = None
+      self._disconnect()
```
linknum23, line 1209: Please revert this change, disconnect should not be called directly by a PersistantStream.

linknum23: I don't know if meta_proc handling needs to be there, I'm trying to remember why it was removed

linknum23:
Lets try this out again on a real unit as well. Then I'll show ya how to rebase these changes.

linknum23:
We should rebase this on the latest develop and see if there are any failures.

amplipi/streams.py
```diff
@@ -1179,23 +1190,55 @@ def _activate(self, vsrc: int):
...
+      self._activate(vsrc)
+
+
     except Exception as exc:
       print(f'error starting lms: {exc}')
 
   def _deactivate(self):
     if self._is_running():
       self.proc.kill()
+    self._deactivate()
```
linknum23, line 1211: Weird infinite recursion here
```suggestion
```

linknum23:
Looks like we messed something up during the rebase. See above

rtertiaer:
@SteveMicroNova if you intend to work on this, can you re-target this PR to `main`? Github automagically closed it when I deleted `develop` :woozy_face: sorry!

linknum23:
@SteveMicroNova  can you squash and rebase this?

linknum23:
I would also suggest retesting it, since some changes were made to the LMS stream in the 0.3.0 release.

linknum23 (changes requested):
See above

SteveMicroNova:
Closing as stale due to #664

PR #466, opened 2023-04-26
Audiodetector install during deploy

Lohrer:
Looks like we missed adding the audiodetector flag to configure.py in the deploy script.


linknum23 (approved):
Tested on a spare amplipi. The audiodetector service is started and there arent any error messages on startup with no inputs playing.

PR #486, opened 2023-08-02
Add password functionality to the updater interface

### LINKED ISSUE #477: Create password protection functionality for the updater/administrative interface
This does not need to be fancy; it should probably be a line item using a hash in the configuration file. We do not need to be implementing users here.

rtasson:
This PR adds simple HTTP Basic authentication for the updater. It does not attempt to document this feature fully, nor present an functionality in the frontend to set this password.  This ought to resolve #477 .

The trick to this PR wasn't in the implementation of the authentication itself, but only optionally presenting it in the setup requesting it (ie, having a `admin_password_hash` configured in `house.json`'s `info` object.) I will self-review these potentially slippery parts of the code and would appreciate an extra set of eyes on it.

I chose to hash this password using [Argon2id](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id). This might seem like overkill, except we're storing the equivalent of a root password on these network connected devices; we should expect it to have strong resistance to attack. (Now the weak point is (the lack of) HTTPS :joy: )  The parameters were a smidge more tricky; the compromise here involves making a hash that's both light enough for a single iteration on a Raspberry Pi Compute 3 module, yet heavy enough to resist brute force on lots of hardware. The parameters m=19456 (19 MiB), t=2, p=1 are presently recommended as a minimum and take about 220ms on our current hardware, the upper end of what we ought to tolerate in response times; this is what I've recommended in the config documentation and tested with.

self-review in 3, 2, 1...

### DIFF
--- a/amplipi/models.py
+++ b/amplipi/models.py
@@ -811,6 +811,7 @@ class Info(BaseModel):
   mock_streams: bool = Field(default=False, description='Are streams being faked? Used for testing.')
   online: bool = Field(default=False, description='can the system connect to the internet?')
   latest_release: str = Field(default='unknown', description='Latest software release available from GitHub')
+  admin_password_hash: Optional[str] = Field(description='An optional Argon2id hash of a password for the update interface. Recommended params: m=19456,t=2,p=1')
   fw: List[FirmwareInfo] = Field(default=[], description='firmware information for each connected controller or expansion unit')
 
   class Config:
@@ -824,6 +825,7 @@ class Info(BaseModel):
             'mock_streams': False,
             'online': True,
             'latest_release': '0.1.8',
+            'admin_password_hash': '$argon2id$v=19$m=19456,t=2,p=1$bWljcm9ub3Zh$e9tlH+uUmDuNk4kXdGrX9NnBLk7uHjJMs7lmZ6EEoBg',
             'fw': [
               {
                 "version": "1.6",
--- a/amplipi/models.py
+++ b/amplipi/models.py
@@ -811,7 +811,6 @@ class Info(BaseModel):
   mock_streams: bool = Field(default=False, description='Are streams being faked? Used for testing.')
   online: bool = Field(default=False, description='can the system connect to the internet?')
   latest_release: str = Field(default='unknown', description='Latest software release available from GitHub')
-  admin_password_hash: Optional[str] = Field(description='An optional Argon2id hash of a password for the update interface. Recommended params: m=19456,t=2,p=1')
   fw: List[FirmwareInfo] = Field(default=[], description='firmware information for each connected controller or expansion unit')
 
   class Config:
@@ -825,7 +824,6 @@ class Info(BaseModel):
             'mock_streams': False,
             'online': True,
             'latest_release': '0.1.8',
-            'admin_password_hash': '$argon2id$v=19$m=19456,t=2,p=1$bWljcm9ub3Zh$e9tlH+uUmDuNk4kXdGrX9NnBLk7uHjJMs7lmZ6EEoBg',
             'fw': [
               {
                 "version": "1.6",
--- a/amplipi/updater/asgi.py
+++ b/amplipi/updater/asgi.py
@@ -63,29 +63,40 @@ dir_path = os.path.dirname(real_path)
 app.mount("/static", StaticFiles(directory=f"{dir_path}/static"), name="static")
 
 INSTALL_DIR = os.getenv('INSTALL_DIR', os.getcwd())
+USER_CONFIG_DIR = os.path.join(os.path.expanduser('~'), '.config', 'amplipi')
 
 # if we have a broken configuration, the updater should still function
-# as a failsafe
+# as a failsafe. This structure & some code was copied from
+# https://github.com/micro-nova/AmpliPi/blob/8368a4a79f536757d7f301612494b6788355aafc/amplipi/app.py#L753
+# except that we don't handle typing or HTML here - this is MVP updater code.
+identity = {
+  'name': 'AmpliPi',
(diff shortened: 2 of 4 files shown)


amplipi/updater/asgi.py
```diff
@@ -59,10 +62,42 @@ class ReleaseInfo(BaseModel):
 dir_path = os.path.dirname(real_path)
 app.mount("/static", StaticFiles(directory=f"{dir_path}/static"), name="static")
 
-HOME = f"{os.environ.get('HOME')}/amplipi-dev" # standard install directory
+INSTALL_DIR = os.getenv('INSTALL_DIR', os.getcwd())
```
rtasson, line 65: This is one spot where I've changed the 'public interface' so to speak, but I think it functions. Prior, we took the `$HOME` var for a particular user and appended `/amplipi-dev`; simultaneously though, we've set the systemd service argument `WorkingDirectory` to the directory where AmpliPi is actually deployed, without making too many assumptions bout its name. This paradigm holds true as far back as commit 3a79c3c4eff994a254919be2d7930fca4e1213e5 titled "WIP web updates close to working", so for the straightforward vanilla case this ought to be fine; hackers may run into mild speedbumps with this though. In those cases, I've exposed an env var INSTALL_DIR (at least for the updater) while defaulting to the cwd.

amplipi/updater/asgi.py
```diff
@@ -59,10 +62,42 @@ class ReleaseInfo(BaseModel):
...
+
+# if we have a broken configuration, the updater should still function
+# as a failsafe
+try:
+  with open(f"{INSTALL_DIR}/house.json") as f:
+    config = json.load(f)
+except:
+  config = {"info": {}}
```
rtasson, line 73: a little brittle considering future other uses of the config in the updater , but oughta function for now. is that copacetic?

amplipi/updater/asgi.py
```diff
@@ -59,10 +62,42 @@ class ReleaseInfo(BaseModel):
...
+
+# The `auto_error` arg below prevents the HTTPBasic middleware from prompting for a password
+# when initial loading of a page while an admin hash is not configured.
+security = HTTPBasic(auto_error = True if 'admin_password_hash' in config['info'].keys() else False)
```
rtasson, line 77: between this and lines 84-86, we ought not prompt nor require authorization unless an `admin_password_hash` is set. Without the initialization of `HTTPBasic` this way, configurations without a password set would always prompt... and `check_password` below would accept anything :upside_down_face:

requirements.txt
```diff
@@ -1,5 +1,6 @@
 adafruit-circuitpython-rgb-display==3.11.0
 aiofiles==23.1.0
+argon2-cffi==21.3.0
```
rtasson, line 3: This is worth testing on a vanilla AmpliPi, as my setup is not very close to upstream at the moment.

rtasson: I'm gonna test this now with a fresh 0.2.1.

linknum23:
My initial suggestion is to integrate these changes into amplipi's identity file instead of house.json since it is only read and not written in-flight like house.json.

Here's a snippet of how app.py uses it

```
# Identity - allows ui customization
identity : Dict[str, Union[str, RawHTML]] = {
  'name': 'AmpliPi',
  'website': 'http://www.amplipi.com',
  'html_logo': '<span class="text-white">Ampli</span><span class="text-danger">Pi</span>'
}
# Load fields from special identity file (if it exists), falling back to default values above
try:
  with open(os.path.join(USER_CONFIG_DIR, 'identity'), encoding='utf-8') as identity_file:
    potential_identity = json.load(identity_file)
    for key, val in identity.items():
      identity[key] = potential_identity.get(key, val)
except FileNotFoundError:
  pass
except Exception as e:
  print(f'Error loading identity file: {e}')
# escape html fields
for key in identity:
  if 'html' in key:
    identity[key] = RawHTML(identity[key])
```

linknum23 (commented):
The rest of the changes look good. I'll test it and figure out how the password checks work tomorrow.

amplipi/models.py
```diff
@@ -824,6 +825,7 @@ class Config:
             'mock_streams': False,
             'online': True,
             'latest_release': '0.1.8',
+            'admin_password_hash': '$argon2id$v=19$m=19456,t=2,p=1$bWljcm9ub3Zh$e9tlH+uUmDuNk4kXdGrX9NnBLk7uHjJMs7lmZ6EEoBg',
```
linknum23, line 828: missed this one

requirements.txt
rtasson: she tested :nail_care:

amplipi/updater/asgi.py
```diff
@@ -59,10 +62,53 @@ class ReleaseInfo(BaseModel):
...
+      detail = "Incorrect password",
+      headers = { "WWW-Authenticate": "Basic" }
+    )
 
 @app.get('/update')
-def get_index():
+def get_index(creds: HTTPBasicCredentials = Depends(check_password)):
   """ Get the update website """
   # FileResponse knows nothing about the static mount
   return FileResponse(f'{dir_path}/static/index.html')
```
linknum23, line 114: NOTE: In the future we should apply the identity file, here, jinja style.

rtasson: I don't understand this comment. can you elaborate?

linknum23: Sorry didn't see this. Just a note for the future. We would use the contents of the identity file to configure the name and logo shown on the updater using jinja templates to render the HTML file.

linknum23 (approved):

Issue #478, opened 2023-07-17
Give the new UI a once-over for fit & finish before marking it 'stable'


rtertiaer:
- [x] `eslint` all React
- [x] implement `eslint` in Github Actions
- [x] Create custom build with ElasticAPM/Sentry/etc frontend JS traceback collection and abuse it
- [x] Look for, file & fix any bugs discovered in the process.

Since 2/3 of the checkboxes here do not have a well-defined stop point, we'll go based on feeling or time elapsed for those.

rtasson:
@SteveMicroNova and I had discussed using ElasticAPM RUM in a temporary fashion to hash out bugs. However, after talking this out with Lincoln, I'm pretty sure it makes sense to implement it as an opt-in debug configuration.

Last commit I see [here](https://github.com/micro-nova/AmpliPi/compare/Webapp2-RUM) is [f810437cee](https://github.com/micro-nova/AmpliPi/commit/f810437cee96f482d2996549737b4654235b4749); I'll pick this up from there and run with it.

rtertiaer:
I'm closing this particular ticket, with #467 being merged (even though we're still just on release candidates of 0.3.0, it's actual launch.)

PR #487, opened 2023-08-03
Bump all user-facing copyright dates

rtasson:
Minor PR in support of #478


klay2000 (commented):
So this is actually only gonna update copyright dates in the old ui, I'll link you to the file where it's located in the new ui

rtasson:
Okay, I think I'm seeing now that this isn't the new ui at all rofl. I'm closing this without merge

Issue #489, opened 2023-08-07
Create A/B partition flashing partitioning scheme


rtasson:
Upgrading AmpliPi software and operating system requirements in-place is sometimes not possible or introduces significant risk. This is the case for switching operating systems, and upgrading from 32bit to 64bit. We should eventually adopt a flashing scheme similar to Android, where we flash a system image to a distinct partition and boot into that, failing back to the prior partition if it does not work. The Raspberry Pi firmware natively supports this functionality through [`tryboot_a_b`](https://www.raspberrypi.com/documentation/computers/config_txt.html#tryboot_a_b). This work supports 

At the moment, we do not need to fully support flashing images to disk from existing software... but the sooner we can change the partition table, the better, as online shrinking of Ext4 is unsupported; the partition scheme, once written is incredibly sticky, while other problems are much more reasonably solved in software further down the line. Considering that there's state to carry over between images, we ought to probably have a small "state" partition at the end of the disk as well.

Currently, AmpliPi system images are ~6gb and take 10-15mins to flash. Upon first boot, `init_resize.sh` will expand first the partition and then the filesystem to fill all empty space on the eMMC. @Lohrer and I discussed the problem space a bit and flshed out a couple of options, each with tradeoffs:
1) Ship two types of images - a single partition containing the update, and a "full" 32GB image that contains all partitions. This enables both Linux and Windows users to un-brick their device (or us to flash/manufacture our own devices) the same way they're doing it now, with the caveat that flashing a full 32GB is extremely time-consuming (perhaps as much as an hour and a half.) However, updates can simply use the smaller single-partition, and we could probably get sophisticated about our flashing too to take less time. This is extremely simply accomplished with our existing `scripts/image_pi`.

2) Ship an image that has two partitions like now, `/boot` and `/`, but supply it with our own custom [`init_resize.sh`](https://github.com/RPi-Distro/raspi-config/blob/408bde537671de6df2d9b91564e67132f98ffa71/usr/lib/raspi-config/init_resize.sh) that creates our extra two partitions upon first boot. This pushes the job of partitioning the devices to the very edge and requires maintaining our own version of `init_resize.sh`; however, it also solves the problem of shipping big images to flash, and also means we only need to ship one image, probably a much better user experience. The updater will probably simply mount the image using `losetup` and shove a new partition where it needs to be. 

Leaning towards option 2, after much hemming & hawing.

linknum23:
[...]

PR #492, opened 2023-08-08
feat: initial boot partitioning script

rtasson:
Since we don't manage `/boot/cmdline.txt` (yet), I think this PR is feature-complete and implements #489 . This is used during initial boot, whereby the kernel is informed that _this_ script is actually `init` in `/boot/cmdline.txt`. At the end of this script, it removes this operand. This same procedure (and a decent chunk of the errata) was gently re-utilized from [raspi-config's `init_resize.sh`](https://github.com/RPi-Distro/raspi-config/blob/408bde537671de6df2d9b91564e67132f98ffa71/usr/lib/raspi-config/init_resize.sh); cheers on the wisdom & code.

This has been tested and run a handful of times with a custom image; when this is accepted, I'll bake an image with this.

One day I'd like to make an `amplipi-scripts` package or similar, because it feels real awkward to host a fake init in `/home` (especially since that will eventually be a different partition), but that's not for today.

On our current CM3's with 32G of memory, the partitioning table ends up looking like this:
```bash
pi@amplipi:~$ sudo parted /dev/mmcblk0
GNU Parted 3.2
Using /dev/mmcblk0
Welcome to GNU Parted! Type 'help' to view a list of commands.
(parted) p                                                                
Model: MMC BJTD4R (sd/mmc)
Disk /dev/mmcblk0: 31.3GB
Sector size (logical/physical): 512B/512B
Partition Table: msdos
Disk Flags: 

Number  Start   End     Size    Type     File system  Flags
 1      4194kB  273MB   268MB   primary  fat32        lba
 2      273MB   10.6GB  10.4GB  primary  ext4
 3      10.6GB  20.9GB  10.3GB  primary
 4      20.9GB  31.3GB  10.3GB  primary
```

### DIFF
--- a/scripts/first_boot_partitioning
+++ b/scripts/first_boot_partitioning
@@ -1,0 +1,78 @@
+#!/bin/sh
+# create a partition scheme on first bootup that looks like this:
+# - boot partition, original size
+# - system partition #1, 10G
+# - system partition #2, 10G
+# - state & /home partition, 10G
+# this scheme permits us to install new operating systems over the air and in the future
+# simplify upgrades. during time of implementation, we have not roadmapped this OTA functionality yet.
+# ref: https://github.com/micro-nova/AmpliPi/issues/489
+#
+# A lot of this script was inspired by (or copied from) Raspberry Pi's init_resize.sh; cheers yall.
+# ref: https://github.com/RPi-Distro/raspi-config/blob/408bde537671de6df2d9b91564e67132f98ffa71/usr/lib/raspi-config/init_resize.sh
+# ref: https://github.com/RPi-Distro/raspi-config/blob/408bde537671de6df2d9b91564e67132f98ffa71/LICENSE
+
+echo "starting first boot partitioning tasks..."
+
+# bail if any command uses an unset variable
+set -u
+
+# spit out the execution of each line; in case we do bail we know where it failed
+set -x
+
+# kernel panic messages will quickly overwhelm the screen if we don't trap the exit during
+# potential troubleshooting
+trap 'sleep 10' EXIT
+
+# set up stuff a regular init might help with
+mount -t proc proc /proc
+mount -t sysfs sys /sys
+mount -t tmpfs tmp /run
+mkdir -p /run/systemd
+mount /boot
+mount / -o remount,ro
+
+ROOT_PART_DEV=$(findmnt / -o source -n)
+ROOT_PART_NAME=$(echo "${ROOT_PART_DEV}" | cut -d "/" -f 3)
+ROOT_DEV_NAME=$(echo /sys/block/*/"${ROOT_PART_NAME}" | cut -d "/" -f 4)
+ROOT_DEV="/dev/${ROOT_DEV_NAME}"
+
+# sanity checks. Should any of these conditions be true, users must manage their own
+# partitioning or reflash.
+partprobe
+blkid ${ROOT_DEV}p3
+if [ $? -eq 0 ] ; then
+  echo "${0}: detected a third partition; exiting."
+  exit 1
+fi
+
+# bail if anything exits with a non-zero exit status
+set -e
+
+# Actually partition.
+parted ${ROOT_DEV} --align optimal --script resizepart 2 34% # system partition 1
+parted ${ROOT_DEV} --align optimal --script mkpart primary 34% 67% # system partition 2
+parted ${ROOT_DEV} --align optimal --script mkpart primary 67% 100% # /home
+
+partprobe ${ROOT_DEV}
+
+# e2fsck _will_ return nonzero on this run.
+set +e
+e2fsck -f ${ROOT_DEV}p2
+if [ $? -ge 3 ]; then
+  echo "${0}: e2fsck -f ${ROOT_DEV}p2 failed; exiting"
+  exit 2
+fi
+set -e
+# resize2fs actually needs the filesystem RW
+mount / -o remount,rw
+resize2fs ${ROOT_DEV}p2
+
+sed -i 's| init=/home/pi/amplipi-dev/first_boot_partitioning||' /boot/cmdline.txt
+
+sync
+
+# fancy reboot for when we're actually init; lines from reboot_pi() in init_resize.sh
+echo b > /proc/sysrq-trigger
+sleep 5
+exit 0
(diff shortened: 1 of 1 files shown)


scripts/first_boot_partitioning
```diff
@@ -0,0 +1,78 @@
...
+  exit 1
+fi
+
+# bail if anything exits with a non-zero exit status
+set -e
+
+# Actually partition.
+parted ${ROOT_DEV} --align optimal --script resizepart 2 34% # system partition 1
+parted ${ROOT_DEV} --align optimal --script mkpart primary 34% 67% # system partition 2
+parted ${ROOT_DEV} --align optimal --script mkpart primary 67% 100% # /home
```
rtasson, line 55: it feels extremely nice to not be tied to a particular disk geometry here. if percentages have undesirable properties here, i'm all ears.

scripts/first_boot_partitioning
```diff
@@ -0,0 +1,78 @@
...
+
+# bail if any command uses an unset variable
+set -u
+
+# spit out the execution of each line; in case we do bail we know where it failed
+set -x
+
+# kernel panic messages will quickly overwhelm the screen if we don't trap the exit during
+# potential troubleshooting
+trap 'sleep 10' EXIT
```
rtasson, line 25: gentle balance here; 10s feels fine enough to wait patiently during first boot of the appliance, but also enough time to snap a picture on your cellphone during a broken install (speaking from today's experience :woozy_face: )

scripts/first_boot_partitioning
Lohrer: For the reasonable future we have the same eMMC on every device so my initial instinct would have been to hard-code values. But really the only potential advantage I can see there is correctly causing an error if we somehow got shipped a wrong 8 or 16 GB Pi. But if that ever is an actual issue we can add it in the final testing/checkout procedure. So sure, you've got my thumbs up to % it up.

scripts/first_boot_partitioning
```diff
@@ -0,0 +1,78 @@
...
+ROOT_PART_NAME=$(echo "${ROOT_PART_DEV}" | cut -d "/" -f 3)
+ROOT_DEV_NAME=$(echo /sys/block/*/"${ROOT_PART_NAME}" | cut -d "/" -f 4)
+ROOT_DEV="/dev/${ROOT_DEV_NAME}"
+
+# sanity checks. Should any of these conditions be true, users must manage their own
+# partitioning or reflash.
+partprobe
+blkid ${ROOT_DEV}p3
+if [ $? -eq 0 ] ; then
+  echo "${0}: detected a third partition; exiting."
```
Lohrer, line 45: I can't think of anything else super great off the top of my head here. Really this is intended to only run at first boot (when the AmpliPi image is flashed) and deletes itself from init afterwards, so I don't see a HUGE need for too many checks. The third partition check is definitely a great sanity check.

Lohrer (commented):
It appears this does not copy p2 into the new p3. Is that something we wanted to do? If it's much effort it's probably not worth adding now since without it we'll still be just as safe as we were.

rtasson:
@Lohrer rofl you're absolutely right; we did discuss doing that. I will do that now. 

note for myself: the list of _new_ things to bake into a new image that are _not_ managed at the moment is
* this addition to `/boot/cmdnline.txt`: `init=/home/pi/amplipi-dev/scripts/first_boot_partitioning`
* a `tryboot` configuration in `/boot/config.txt`. I will have to test this a bit.

rtasson:
After a great deal of back & forth, flashing this, I believe I've arrived at a good place with this for re-review.

Lohrer (approved):
This looks good for imaging new Pis. Since I had to figure out what was going on, I added some documentation for the partitioning, hopefully it makes sense.

scripts/first_boot_partitioning
```diff
@@ -0,0 +1,96 @@
...
+ROOT_DEV_NAME=$(echo /sys/block/*/"${ROOT_PART_NAME}" | cut -d "/" -f 4)
+ROOT_DEV="/dev/${ROOT_DEV_NAME}"
+
+# sanity checks. Should any of these conditions be true, users must manage their own
+# partitioning or reflash.
+partprobe
+blkid ${ROOT_DEV}p3
+if [ $? -eq 0 ] ; then
+  echo "${0}: detected a third partition; removing this script from initial boot & exiting."
+  sed -i 's/ init=\/home\/pi\/amplipi-dev\/scripts\/first_boot_partitioning//' /boot/cmdline.txt
```
linknum23, line 47:
```suggestion
  sed -i 's@ init=/home/pi/amplipi-dev/scripts/first_boot_partitioning@@' /boot/cmdline.txt
```

scripts/first_boot_partitioning
```diff
@@ -0,0 +1,96 @@
...
+echo "${0}: creating recovery partition..."
+mount -o remount,ro /
+sleep 5
+dd if=${ROOT_DEV}p2 of=${ROOT_DEV}p3 bs=2M oflag=dsync status=progress
+mount -o remount,rw /
+parted ${ROOT_DEV} --align optimal --script resizepart 2 34% # system partition 1
+
+# then repair and expand our filesystems within these partitions
+for PART in ${ROOT_DEV}p2 ${ROOT_DEV}p3; do
+    # we don't run e2fsck on the mounted root partition.
```
linknum23, line 78: why?

linknum23 (commented):
How is the actual partition fallback handled?

Lohrer:
> How is the actual partition fallback handled? Or will that be implemented in the future?

Implemented in the future, that requires updater changes. Mahdee needs to image more boxes ASAP so this just prevents the need to re-flash when we move to 64-bit.

scripts/first_boot_partitioning
Lohrer: Yeah I kinda agree haha, looping over 2 things and they are both handled differently seems slightly silly but I wasn't going to raise a big stink.

linknum23: Nah I just wanted the comment to explain why is doesn't make sense to fsck the mounted partition. It probably just doesnt work?

rtasson:
@linknum23 this is mentioned in my note to self: https://github.com/micro-nova/AmpliPi/pull/492#issuecomment-1671561138

we would implement the partition a/b swapping with a `tryboot` parameter in `config.txt`; however, we do not manage `config.txt` at all now, save for copying from a prior image and putting it in future ones. we will implement it manually during image creation for now, but I'd like to one day actually manage config.txt during image builds. it just has no representation here in this PR :upside_down_face:

linknum23 (approved):

PR #494, opened 2023-08-15
Upgrade UI to a modern webapp framework

linknum23:
This PR integrates pretty many UI changes made by @jonahshader and @klay2000 several months ago. The entire app/design was converted to react and made way less clunky.

If I recall there are several things to fix here:
- [ ] Fix weird zone and stream issues ( we think we tracked these down to the backed at some point)
- [ ] Deploy script fails on some conflicting dependencies

We should really merge this into develop so we can make issues on what is still broken and get feedback from users on the new interface.


rtasson:
I've merged this into `develop` on localhost & deployed+tested on an amplified box & a streamer box; nothing seems amiss, though you're possibly aware of the edge cases on the backend I oughta pay attention to. @linknum23 are you opposed to me pulling the trigger, or do you have a list of tests you typically perform that I can validate this against?

rtasson:
I've tested bluetooth on this as well; works a charm. Merging.

PR #504, opened 2023-08-29
ESLint + debug ElasticAPM RUM 

rtasson:
This is a wombo-combo of two distinct things - the work @SteveMicroNova did to ESLint the entire repo, and the ElasticAPM integration for a quick debug mode config to get feedback on `webapp_2`. This is the branch that's currently playing things in the office :sunglasses: I'm opting to PR this into `webapp_2` so we can begin working on merging #494 :upside_down_face: It might make sense to review this PR commit by commit, and skipping the large linting commits; should some concerns arise about some of the super buried bits of this, we can re-constitute a branch with cherry-picked commits and work off that.

Things I'd appreciate feedback on:

- [ ] is the eslint ruleset at [.eslintrc.cjs](https://github.com/micro-nova/AmpliPi/commit/b93f51c9b62a3d1dfefc1d762b2e69127c4ca976#diff-1cfbdfd2dc2aa46e827d0b9512aff1f04a1f7453b4329261497e53d8534d03a8) what we want to run with? I'm a little concerned that the changes are so profound in some of these commits.
- [ ] anybody opposed to targetting `es2022` for [top level await](https://caniuse.com/?search=top%20level%20await)? I know that makes me a smidge lazy... but Slack is deprecating my Firefox 107 browser on Sept 1st so I think we're probably fine lol


SteveMicroNova:
The main functional changes added in my commits would be just commenting out unused variables, so if there's something wrong I'd suggest searching for comment delineators, especially in the earlier commits

jonahshader (commented):
I think I prefer two spaces for indentation, and maybe even no semicolon. The rest of the formatting changes make sense. I like the react propTypes thing.

web/src/App.jsx
```diff
@@ -0,0 +1,231 @@
...
+try {
+    const r = await fetch("/debug");
+    const debugDocText = await r.text();
+    const debugDoc = JSON.parse(debugDocText);
+    const debug = (debugDoc && debugDoc["debug"]) ? true : false;
+
+    var apm = initApm({ // eslint-disable-line no-unused-vars
+        active: debug,
+        serviceName: "Amplipi",
+        serverUrl: debugDoc["apmHost"] ? debugDoc["apmHost"] : "",
```
rtasson, line 29: @jonahshader I use top level await here to instantiate a very early connection to an arbitrary APM server, specified from a debug endpoint. I could also resolve the promise myself, or use a fancy vite plugin to transform top level awaits into resolved ones. I just didn't see a reason not to track more modern browsers and it seemed least messy at the time.

rtasson, line 29: @jonahshader I use top level await here to instantiate a very early connection to an arbitrary APM server, specified from a debug endpoint. I could also resolve the promise myself, or use a fancy vite plugin to transform top level awaits into resolved ones. I just didn't see a reason not to track more modern browsers and it seemed least messy at the time.

jonahshader: Ah, got it. I think @linknum23 should answer this since he advocated for more backwards compatibility in the past.

rtasson: for context, introducing top level await drops support from estimates of 95% to 90% of global web browsers.

jonahshader:
Looking at the react static propType, react just [recommends using TypeScript](https://react.dev/reference/react/Component#static-proptypes) instead. How difficult would it be to turn this into a TypeScript project? Considering that TypeScript is a strict superset of JavaScript, making that change won't break anything, and we could slowly add type checking to our components as needed. We could also add type checks to our other non-component functions, which isn't supported by prop-types.

klay2000 (approved):
looks fine formatting wise to me other than the 4 space tabs

linknum23 (approved):
This is a good start. Lets integrate this so we can polish up the new interface.

PR #505, opened 2023-08-30
Web app2 select stream

SteveMicroNova:
ONLY MERGE THIS AFTER #504 TO AVOID PAINFUL MERGE CONFLICTS

This PR just adds the ability to change the stream from the player menu, that way you can change the song without leaving the screen. Wholesale copy+pasted code from the Home page to make this work, very small change.

This was based on branch Webapp2-Eslinting so there's far more commits than this branch actually owns, the only real ones are the bottom two and the only meaningful one is the one prior to the last since the last is just NPM package updates


rtasson (approved):
:call_me_hand:

rtasson:
yeah, I would probably only cherry-pick the necessary commit if this somehow doesn't recognize that this branch already has those other eslint'ing commits.

SteveMicroNova:
> Could you make the StreamBar appear to be clickable somehow? The stuff on the homepage all changes the mouse icon to the "clickable" icon, and they also are wrapped in Chip. Maybe we could make the StreamBar look like a larger StreamBadge?

Good idea, I'll solve this similarly to how I'm solving things in #507 and push a solution to this branch

Issue #497, opened 2023-08-16
WebApp2 - Remove dropdown volume bar from the player screen in some cases


SteveMicroNova:
Remove dropdown volume bar from the player screen if there's only one group/zone, instead use the volume bar for that group/zone only

This is to reduce the number of excess dropdowns, and by extension to reduce the number of excess clicks/taps required to change a specific zone's volume. There's no reason for there to be a dropdown that contains a single item / only contains another dropdown

PR #506, opened 2023-08-30
Solution to issue #497

SteveMicroNova:
#497


linknum23 (approved):
Nice. Looks good.

web/src/pages/Player/Player.jsx
```diff
@@ -24,6 +24,7 @@ const Player = () => {
     );
     const sourceIsInactive = getSourceInputType(selectedSource) === "none";
     const [expanded, setExpanded] = useState(false);
+    const [alone, setAlone] = useState(false);
```
jonahshader, line 27: Feels like the alone state might be redundant since its derived purely from other states. Could you compute the alone value here instead of in VolumeZones? That way we can get rid of the extra state and VolumeZones would no longer need to take in setAlone. 

Since alone depends on getFittestRep, that would need to be computed in Player and the output passed down to VolumeZones instead of VolumeZones computing it.

jonahshader (approved):
Looks good

SteveMicroNova:
2/3 approvals seems good enough for me to merge, closing this request

Issue #499, opened 2023-08-16
Webapp2 - Make more things responsive on the Home screen


SteveMicroNova:
The X to close a stream glows white when you hover over it, the other clickables (I avoid saying "buttons" because they're not all something that would be considered a button) should as well. Examples include the mute button, the stream name, and the zone list
 
This is just to make the styling more consistent, only about half of all clickables use the highlight on hover styling, and it seems that those have been chosen at random to some extent

SteveMicroNova:
Solved August 31st, closing as complete

PR #507, opened 2023-08-31
Webapp2 home responsiveness

SteveMicroNova:
Accidentally made these changes on top of my branch Webapp2-VolumeDropdown, the changes actually meant for this PR are all in the last commit

I just added hoverover interactivity on a bunch of clickables on the home page, as per issue #499


linknum23:
This is a perfect case for using cherry pick.

jonahshader (changes requested):
Looks like the current highlighting doesn't quite line up with the StreamBadge and ZonesBadge. Since they both are wrapped in Chip, and this is the only use of Chip, I think we should just add the clickable styling to Chip.scss, so .chip would get @include general.button-hover; in addition to cursor: pointer.

web/src/components/Chip/Chip.scss
```diff
@@ -8,4 +8,5 @@
   padding: 0.3rem 0.5rem;
   margin: 0.25rem;
   cursor: pointer;
+  @include general.button-hover;
```
jonahshader, line 11:
```suggestion
  @include general.highlight-hover;
```

Turns out button-hover includes a border-radius: 1rem which doesn't line up with chip's border-radius, so I think we should spit this up so highlight-hover just applies the background-color: $secondary and button-hover includes that and the border-radius. Maybe even rename button-hover to like rounded-button-hover or something.

web/src/components/VolumeSlider/VolumeSlider.scss
```diff
@@ -20,6 +20,8 @@
 
 .volume-slider-mute-icon {
   color: general.$mute-color;
+  cursor: pointer;
+  @include general.button-hover;
```
jonahshader, line 24: Can we also apply this to the non-mute icon?

web/src/components/Chip/Chip.scss
jonahshader: Also it appears that 'cursor: pointer' is always used alongside '@include general.button-hover', so that could be added to the 'general.button-hover' or 'general.highlight-hover' mixins.

jonahshader (changes requested):
A couple of css changes requested. Besides that the rest looks good!

jonahshader (approved):
Looks good, thanks!

klay2000 (approved):

PR #509, opened 2023-09-01
Empty groups

### LINKED ISSUE #508: Empty group fails to compute volume
See: https://github.com/micro-nova/AmpliPi/blame/eb806324e5c99fafcb1cdfc00c458d67c0d3f310/amplipi/ctrl.py#L744

linknum23:
This fixes #508 reported on our amplipi.discourse.group

@SteveMicroNova can you verify that the updated tests break when a group has no zones? Can you also verify that the fix makes sense?


tests/test_rest.py
```diff
@@ -1267,24 +1268,30 @@ def patch_group(json: Dict, expect_failure: bool = no_groups) -> Optional[Dict]:
     assert patched_group is not None
     return patched_group
 
+  # if a group has no zones, voluem control won't work as expected
```
SteveMicroNova, line 1271:
```suggestion
  # if a group has no zones, volume control won't work as expected
```
Minor spelling error in comment, not a real issue but might as well be fixed

SteveMicroNova (approved):
Single minor spelling issue I found, all the tests ran appropriately and it looks good otherwise

PR #510, opened 2023-09-01
Webapp2 lock volume

SteveMicroNova:
Adding a lock to the volume sliders on the frontend (and logic to support it to the backend) that will allow volume to be locked so that it can't be changed, generally useful when a zone is part of a group but doesn't want to be effected by the same volume changes as the group as a whole


rtertiaer:
@SteveMicroNova if you intend to work on this, can you re-target this PR to `main`? Github automagically closed it when I deleted `develop` :woozy_face: sorry!

linknum23:
My instincts are to close this until we decide how to implement this. Lets wait for more feedback on the new UI.

PR #511, opened 2023-09-06
feat: add development flag to scripts/configure.py

rtertiaer:
This PR adds a simple `--development` flag to `scripts/configure.py`. At present, it simply disregards the platform checking; I suspect it could be helpful to use elsewhere too though.

Also, `.gitignore`: we now have errant `node_modules/` floating around cuz of `webapp_2` development, and also my `.swp` files from `vi`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ Jinja2==3.1.2
 loguru==0.6.0
 mypy==1.0.0
 netifaces==0.11.0
-numpy==1.21.6
+numpy
 Pillow==9.4.0
 psutil==5.9.4
 pydantic==1.10.4
--- a/scripts/configure.py
+++ b/scripts/configure.py
@@ -992,6 +992,6 @@ if __name__ == '__main__':
   result = install(os_deps=flags.os_deps, python_deps=flags.python_deps, web=flags.web,
           display=flags.display, audiodetector=flags.audiodetector,
           firmware=flags.firmware, password=flags.password,
-          restart_updater=flags.restart_updater)
+          restart_updater=flags.restart_updater, development=flags.development)
   if not result:
     sys.exit(1)
--- a/scripts/configure.py
+++ b/scripts/configure.py
@@ -989,7 +989,9 @@ if __name__ == '__main__':
     print('  WARNING: expected some arguments, check --help for more information')
   if sys.version_info.major < 3 or sys.version_info.minor < 7:
     print('  WARNING: minimum python version is 3.7')
-  install(os_deps=flags.os_deps, python_deps=flags.python_deps, web=flags.web,
+  result = install(os_deps=flags.os_deps, python_deps=flags.python_deps, web=flags.web,
           display=flags.display, audiodetector=flags.audiodetector,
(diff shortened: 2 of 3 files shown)


linknum23 (approved):

linknum23:
LGTM

scripts/configure.py
```diff
@@ -989,7 +989,9 @@ def failed():
     print('  WARNING: expected some arguments, check --help for more information')
   if sys.version_info.major < 3 or sys.version_info.minor < 7:
     print('  WARNING: minimum python version is 3.7')
-  install(os_deps=flags.os_deps, python_deps=flags.python_deps, web=flags.web,
+  result = install(os_deps=flags.os_deps, python_deps=flags.python_deps, web=flags.web,
           display=flags.display, audiodetector=flags.audiodetector,
           firmware=flags.firmware, password=flags.password,
           restart_updater=flags.restart_updater)
+  if not result:
+    sys.exit(1)
```
rtertiaer, line 997: note to self: test that updater successfully restarts after this change, cuz it uses `subprocess.Popen` to call this and may care about exit status.

requirements.txt
```diff
@@ -11,7 +11,7 @@ Jinja2==3.1.2
 loguru==0.6.0
 mypy==1.0.0
 netifaces==0.11.0
-numpy==1.21.6
+numpy
```
rtertiaer, line 14: numpy is imported to speed up rendering on the TFT display, as recommended by https://github.com/adafruit/Adafruit_CircuitPython_RGB_Display . it's only used a small handful of places here and without a version bound; validate all versions of numpy haven't substantially changed the functions used here: https://github.com/adafruit/Adafruit_CircuitPython_RGB_Display/blob/main/adafruit_rgb_display/rgb.py and also that the TFT display works well.

rtertiaer: The only code that uses numpy in the Adafruit library is here: https://github.com/adafruit/Adafruit_CircuitPython_RGB_Display/blob/8cfb0db7d36eb46ef1e2a575b03f5e630bfc2c94/adafruit_rgb_display/rgb.py#L68

The functions & methods in question are:
```
numpy.array()
numpy.dstack().flatten().tolist()
```

On first brush with the documentation, these seem well supported (indeed, foundational) from 1.21 thru 1.26, the latest numpy release.

This still probably deserves a run on a box with a TFT display to validate.

rtertiaer:
lastly, I manually merged #524 into a local branch and validated that mypy and pylint (and all other tests, though those felt ancillary) were happy with these changes.

requirements.txt
rtertiaer: I validated this functions on a box with a TFT display.

linknum23:
I think the real thing to test here since configure.py was modified is can we update from this version to something else

linknum23 (approved):

rtertiaer:
in terms of configure.py, I've tested:
* vanilla configure.py -> new configure.py
* new configure.py -> new configure.py
* new configure.py -> vanilla configure.py

PR #513, opened 2023-10-03
Fix output device from streamer-lms merge

rtasson:
There are a lot of style-type things that `pylint` is angry about, but this is the actual reason it's erroring on the `webapp_2` branch. @linknum23 , is this correct? seems the options are `real_output_device` and `virtual_output_device`; I don't have enough context to guess which is correct :woozy_face:


linknum23 (approved):
This should fix our issues for now. Just a heads up we will need to convert LMS and Bluetooth to be Persistent streams in the future. At that point they will use the virtual devices.

PR #514, opened 2023-10-03
add type annotations for einkdisplay.py

rtasson:
This PR adds simple type annotations to einkdisplay.py, which should allow mypy tests to pass.


rtasson:
c'mon Github, just rerun the full jobs :sweat_smile:

PR #518, opened 2023-10-04
Password CRUD; authenticated main UI

### LINKED ISSUE #517: Password CRUD; protect general interface
We have implemented backend functionality to password protect the updater interface. We need to create user-facing CRUD for this functionality, as well as allow the general interface to be password protected as well.

rtertiaer:
This PR implements simple password Create, Update and Delete functionality. It also introduces API Key & cookie authentication to the main UI when a password is configured. Once implemented, this resolves #517 .


linknum23:
I think the set password text might be wrong. It mentions a username but doesn't show one on the ui? 
2023-10-04_1068x810

amplipi/utils.py
```diff
@@ -320,3 +325,50 @@ def vol_db_to_float(vol: int, db_min: int = models.MIN_VOL_DB, db_max: int = mod
 def debug_enabled() -> bool:
   """ Returns true or false if debug is enabled """
   return pathlib.Path.home().joinpath(".config/amplipi/debug.json").exists()
+
```
linknum23, line 328: Do we need RawHTML? I don't see it being used?

amplipi/updater/asgi.py
```diff
@@ -291,6 +291,38 @@ def install(creds: HTTPBasicCredentials = Depends(security)):
...
+@app.post('/password')
+def set_admin_password(input: PasswordInput, creds: HTTPBasicCredentials = Depends(security)):
+  """ Sets the admin password """
+  if len(input.password) == 0:
+    del identity['admin_password_hash']
+  else:
+    # The below takes ~200ms to hash on a RPi3 CM and is currently (fall 2023) recommended
+    # as a 'safer' default by OWASP
+    argon2_params = Argon2Params(
+      time_cost=2,
```
rtasson, line 306: if literally every request on the API now has to go through this (because we're using HTTPBasic auth, instead of token bearer), this probably ought to be 1 instead. adding 200ms for every request is absolutely something a user can feel, in terms of response times.

amplipi/utils.py
linknum23: Nevermind, just realized this got moved to utils, makes sense.

rtasson:
@linknum23 The `username` field is something _all_ web browsers prompt for when doing HTTP Basic authentication. However, as of right now we don't use it. I can see a few paths:
* clarify the text here to note that the username field here is something the web browser will prompt for, not something you need to set
* actually get & set a username 
* set the username as 'admin' by default and simply inform the user.

amplipi/updater/asgi.py
linknum23: What level of effort is using token bearer? Are we planning on implementing it soon?

linknum23:
> set the username as 'admin' by default and simply inform the user.

Does this mean the username in the browser popup can be pre-populated?

If so that seems the most straightforward way to handle this.

amplipi/updater/asgi.py
rtasson: The level of effort is about 2x as much as this, on top of this PR. It's accomplish-able, but I don't personally want to create and manage a login workflow, which is what that would entail; HTTP Basic shifts that responsibility to the browser.

rtasson:
> Does this mean the username in the browser popup can be pre-populated?

I do not believe the username could be prepoulated in an HTTP Basic login flow, no.

rtasson:
After discussing this back & forth in person with @linknum23, it probably just makes sense to actually create an authentication workflow.

rtertiaer:
I've changed this quite a bit in the current rendition.
* We now manage users in a `users.json` file from auth.py, removing a dependency on the identity file
* When a password is set through the updater, an `admin` user is created (though that username is not user-facing) and a password hash and an API key are associated with the user
* The API key can be used via the query param `api_key` or via a cookie named `amplipi-session`; this cookie is set upon successful log in.
* Whenever authentication is necessary, the application renders a very barebones authentication flow. When complete, that flow will send the user back to the requested resource.
* The hooks in both the `updater` and main app are a custom router for `/auth`, an unauthenticated exception & handler added to the `FastAPI()` instance, and optionally the actual `CookieOrParamAPIKey` injected dependency (only if the "admin" user has a password set.)

This is pretty well typed and tested at the moment. I still need to add documentation/comments to this before it's ship-able, but would appreciate feedback on my methodology (or anything else) before I walk too far down that path.

amplipi/auth.py
```diff
@@ -0,0 +1,216 @@
...
+  except json.JSONDecodeError as e:
+    print(f'Error loading users file as JSON: {e}')
+    print('Moving the old one to a backup and creating a users file from defaults.')
+    os.rename(USERS_FILE, f"{USERS_FILE}.{time.time()}")
+    with open(USERS_FILE, encoding='utf-8', mode='w') as repair_file:
+      json.dump(users, repair_file)
+  except Exception as e:
+    print(f'Error loading identity file: {e}')
+    raise e
+  return users
```
rtertiaer, line 75: this entire function is pretty much copypasta from the `set_identity` function I created (but stopped using) in `utils.py`. There's a strong case to be made to move this into its own function, except there's also an aversion to tying this file into `utils.py` or any other location in this codebase, to permit the updater to function if a lot of the rest of the codebase is broken. thoughts?

linknum23: I think this makes sense as is for the most part. The only thing I would do is make set/get_users private so that we can change the format in the future without effecting code outside this file.

linknum23 (approved):
Looks good! I really like how easily auth integrates with the system.

Minor suggestions:
- make get/set_users private with an _
- Make the users api key reported in the info field  (and show it in the **Help** section)

linknum23 (commented):
All of these changes look great. Lemme test it out on lamplipi.local

linknum23:
Everything still works as expected. A couple of things to note:
- An update using the deploy script took over an hour, this seems to be related to an `npm install` in the configure script
- The previous password may have been reset during the update, please verify this does not happen in the normal update case.

Assuming the update process works fine for you, I think this is good to go. Also the seamless integration of all of this looks impressive, nice work!

rtertiaer:
I was not able to replicate a re-deploy clearing out the password. Looking at the code in _get/_set_users, things are pretty well guarded; the two cases where the users file is cleared here is when _get_users() cannot parse the file as JSON, or when a FileNotFound exception is pushed up. Anything else re-raises and _set_users doesn't handle it, which happens before it writes the user file.

I'll definitely keep an eye out for this in wider testing later in the release cycle for 0.3.0, but I feel reasonable about shipping this into `develop` for now.

rtertiaer:
also, a last note: we'll be addressing the `npm install`/`npm run build` taking an eternity in another PR and will ship before we call 0.3.0 stable.

Issue #516, opened 2023-10-03
Feedback on unit display during first boot


rtasson:
First boot of a freshly imaged AmpliPi will take a long time, due to creating partitions and copying a recovery OS onto them (#492). This process takes a while; end users need visual feedback that indicates this is happening, or else it may be incomplete and leave the AmpliPi in a poorly defined state.

rtertiaer:
would also be really nice to have during initial unboxing; possibly a QR code to the manual.

klay2000:
After some investigation we'll probably just make the updater write to the screen in #599 and for this issue show a message on boot as well as clear the screen on a clean shutdown.

PR #519, opened 2023-10-05
Remove time-consuming copy on first boot

rtasson:
This PR removes the time-consuming copying process on first boot. It leaves in place the partition creation process. This will allow us to punt on #516 as a blocker for 0.3.0 stable release.

I have not tested this yet & will not merge til I have done so, but there's nothing stopping feedback at the moment.

### DIFF
--- a/scripts/first_boot_partitioning
+++ b/scripts/first_boot_partitioning
@@ -63,18 +63,12 @@ set -e
 # Leave p2 shrunk for now so as to reduce copy time.
 parted ${ROOT_DEV} --align optimal --script mkpart primary 34% 67% # Create system partition 2
 parted ${ROOT_DEV} --align optimal --script mkpart primary 67% 100% # Create /home
+parted ${ROOT_DEV} --align optimal --script resizepart 2 34% # system partition 1
 partprobe ${ROOT_DEV}
 
-# copy system partition 1 to system partition 2, and resize both filesystems to fill
-echo "${0}: creating recovery partition..."
-mount -o remount,ro /
-sleep 5
-dd if=${ROOT_DEV}p2 of=${ROOT_DEV}p3 bs=2M oflag=dsync status=progress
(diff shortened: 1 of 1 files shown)


scripts/first_boot_partitioning
```diff
@@ -63,18 +63,12 @@ set -e
...
-sleep 5
-dd if=${ROOT_DEV}p2 of=${ROOT_DEV}p3 bs=2M oflag=dsync status=progress
-mount -o remount,rw /
 parted ${ROOT_DEV} --align optimal --script resizepart 2 34% # system partition 1
+partprobe ${ROOT_DEV}
 
 # then repair and expand our filesystems within these partitions
-for PART in ${ROOT_DEV}p2 ${ROOT_DEV}p3; do
+# This structure is left in place for future partitions... but it basically does nothing now.
+for PART in ${ROOT_DEV}p2; do
```
Lohrer, line 71: Maybe just remove this? The loop over 2 things that are then handled completely differently was kinda silly in the first place...

rtasson: i'm inclined to leave it in place only cuz the note about e2fsck returning >0 exit code on successful runs was unexpected and took some debugging, but i can if you care. agreed it was silly to handle the cases differently :woozy_face:

rtasson:
This has now been tested too.

scripts/first_boot_partitioning
Lohrer: Ah I misread the `!=` as `=`, this makes more sense now. Plus since you've already tested it as written I'm fine with it staying this way.

Lohrer (approved):

Lohrer:
The display code appears to fail the tests, I assume that's not due to this PR? If not, we should still fix those quickly so we don't keep ignoring test outputs.

PR #528, opened 2023-11-10
Fix Disconnected Zone Bug

### LINKED ISSUE #527: Look into stream audio playing out the wrong zones
altenrative: spruce up logging around vulnerable bits so we can get good feedback in the field

klay2000:
This likely fixes #527


linknum23:
Looking at this again. I think we need to fix the preamp code to handle an input of -1.  Also that max code should do all zones sources I think.

linknum23:
Here's what we need to do to fix this. Sorry this is probably a little too prescriptive
- [ ] Add a test that sets zone sources to -1 and back
- [ ] Add an assert in rt.py Mock update_zone_sources that checks if any of the sources are outside range(4)
- [ ] Verify the test fails
- [ ] Fix the code in set_zones to configure the source_ids sent to the rt within range(4) suggestion use source_id=0

v3
linknum23, line 1: What is this file?

amplipi/ctrl.py
```diff
@@ -722,14 +722,17 @@ def set_zone(self, zid, update: models.ZoneUpdate, force_update: bool = False, i
           mute = True
           update_mutes = True
 
-        # update the zone's associated source (defaulting to source 0 if disconnected)
+        # update the zone's associated source
         zones = self.status.zones
         if update_source_id or force_update :
-          zone_sources = [zone.source_id for zone in zones]
-          zone_sources[idx] = max(sid, 0) # default disconnected zones to source 0
+          zone_sources = [utils.clamp(zone.source_id, 3, 0) for zone in zones]
```
linknum23, line 728:
```suggestion
          zone_sources = [utils.clamp(zone.source_id, 0, 3) for zone in zones]
```

linknum23: That doesn't change functionality, but matches the clamp signature

linknum23 (approved):
Looks good, you may want to squash your commits after removing that weird file

PR #530, opened 2023-11-13
Show better network error info

Lohrer:
Improved error shown on TFT display if network interface missing. Previously it either showed the IP address, or disconnected if there was no IP address assigned. Now there will be 3 possibilities: No network interface, no address assigned (link down), or showing the assigned IP address.


rtertiaer (approved):
:shipit:

PR #533, opened 2023-11-15
Build Webapp Off Amplipi

klay2000:
Changed where webapp is built and added building the webapp to the release docs.


rtertiaer (approved):
:raised_hands:

PR #536, opened 2023-11-21
Prevent node_modules from being deployed

rtertiaer:
The location of the `node_modules` stuff in #533 was incorrect; `poetry build` creates the tarball that is deployed, and this moves that command to after `mv node_modules [...]`.


rtertiaer:
re-jiggered this with some input from Lincoln.

klay2000 (approved):

PR #540, opened 2023-11-30
DLNA Metadata

klay2000:


linknum23:
Just a heads up on this. We are waiting to integrate this after the current release to avoid too many close to the wire changes. Sorry for the delay.

rtertiaer:
@klay2000 can you re-target this PR to `main`? Github automagically closed it when I deleted `develop` :woozy_face:

amplipi/streams.py
```diff
@@ -78,7 +78,7 @@ class BaseStream:
   def __init__(self, stype: str, name: str, only_src=None, disabled: bool=False, mock=False):
     self.name = name
     self.disabled = disabled
-    self.proc: Optional[subprocess.Popen] = None
+    self.metadata_proc: Optional[subprocess.Popen] = None
```
linknum23, line 81: The first proc is intended to be the main running process, most streams also have a secondary metadata process. To handle streams that don't have a metadata process _is_running() should check the main running process. proc is a pretty generic name there may be a better one.

linknum23: Recent changes to process handling have added code that looks like this for carefully shutting down processes without leaving any zombies. Here's what the spotify equivalent looks like:

```python3
    if self.proc:
      try:
        self.proc.terminate()
        self.proc.communicate(timeout=3)
      except Exception as e:
        print(f"failed to terminate spotify stream: {e}")
        self.proc.kill()
      self.proc = None
    if self.mpris:
      try:
        del self.mpris
      except Exception:
        pass
      self.mpris = None
```

amplipi/streams.py
```diff
@@ -802,51 +805,77 @@ def connect(self, src):
...
+
     self._connect(src)
 
   def disconnect(self):
     if self._is_running():
-      os.killpg(os.getpgid(self.proc.pid), signal.SIGKILL)
-      self.proc2.kill()
+      self.metadata_proc.kill()
+      self.dlna_proc.kill()
+      self.fifo.close()
```
linknum23, line 841: Is it better to close the fifo first?

amplipi/streams.py
```diff
@@ -802,51 +805,77 @@ def connect(self, src):
...
-          source.track = d['title']
-        return source
-    except Exception:
+      data = json.load(open(f'{self.src_config_folder}/meta.json'))
+      print(data)
+      source.state = data['state']
+      if source.state != 'stopped': # if the state is stopped, just use default values
+        source.artist = data['artist']
+        source.track = data['title']
+        source.album = data['album']
```
linknum23, line 855: We may want to add some defaults to these like:

```suggestion
        # handle unpopulated fields
        source.artist = data.get('artist')
        source.track = data.get('title')
        source.album = data.get('album')
````

File: amplipi/streams.py [hunk 4]
```diff
@@ -802,51 +805,77 @@ def connect(self, src):
...
-          source.album = d['album']
-        if 'artist' in d:
-          source.artist = d['artist']
-        if 'title' in d:
-          source.track = d['title']
-        return source
-    except Exception:
+      data = json.load(open(f'{self.src_config_folder}/meta.json'))
+      print(data)
+      source.state = data['state']
```
linknum23, line 851:
```suggestion
      source.state = data.get('state', 'stopped') if data else 'stopped'

File: amplipi/streams.py [hunk 5]
```diff
@@ -802,51 +805,77 @@ def connect(self, src):
...
+      if source.state != 'stopped': # if the state is stopped, just use default values
+        source.artist = data['artist']
+        source.track = data['title']
+        source.album = data['album']
+        if data['album_art'] != '':
```
linknum23, line 856:
```suggestion
       if data.get('album_art'):

File: streams/dlna_meta.py [hunk 6]
```diff
@@ -1,68 +1,210 @@
...
+          if transport_state != "STOPPED": # reset the stop counter if the state is not stopped
+            stop_counter = 0
+
+        except Exception as e:
+          print(f"Error: could not get transport state: {e}")
+
+        # try to get song-info from the service and parse it
+        try:
+          meta_xml = ET.fromstring(service.GetMediaInfo(InstanceID=0)["CurrentURIMetaData"])
+          for i in meta_xml.iter():
```
linknum23, line 71: We may want to be careful here about synchronizing artist/title/album/album_art data. Should all be set to ''  or None when we get new metadata?

streams/dlna_meta.py
```diff
@@ -1,68 +1,210 @@
...
+            if i.tag == "{http://purl.org/dc/elements/1.1/}title":
```
linknum23, line 72: It looks like they are using an xml serialization schema. Is it more robust to use a python deserializer that can handle point updates to the schema?

streams/dlna_meta.py
```diff
@@ -1,68 +1,210 @@
 #!/usr/bin/python3
```
linknum23, line 1: This line should be removed so we can launch this with the same version of python that amplipi is running

amplipi/streams.py
```diff
@@ -802,51 +805,77 @@ def connect(self, src):
...
+                 self.src_web_folder, '-d']
     dlna_args = ['gmediarender', '--gstout-audiosink', 'alsasink',
                 '--gstout-audiodevice', utils.real_output_device(src), '--gstout-initial-volume-db',
                 '0.0', '-p', f'{portnum}', '-u', f'{self.uuid}',
-                '-f', f'{self.name}', '--logfile',
-                f'{src_config_folder}/metafifo']
-    self.proc = subprocess.Popen(args=meta_args, preexec_fn=os.setpgrp, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    self.proc2 = subprocess.Popen(args=dlna_args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                '-f', f'{self.name}']
+    self.metadata_proc = subprocess.Popen(args=meta_args)
```
linknum23, line 830: This should be launched with the version of python we are running so we don't have to depend on the system's version of python. Otherwise the pi will need to have upnpy installed on the systems version of python.

linknum23 (changes requested):
This is a really nice upgrade but needs a little cleanup. I'm about 60% done reviewing it but theres already some changes needed.

amplipi/streams.py
linknum23: Actually based on how the metadata script works, this is probably not necessary.

amplipi/streams.py
```diff
@@ -1074,18 +1103,18 @@ def connect(self, src):
...
-      return self.proc.poll() is None
+    if self.metadata_proc:
+      return self.metadata_proc.poll() is None
     return False
 
   def disconnect(self):
     if self._is_running():
-      os.killpg(os.getpgid(self.proc.pid), signal.SIGKILL)
-    self.proc = None
+      os.killpg(os.getpgid(self.metadata_proc.pid), signal.SIGKILL)
```
linknum23, line 1116: Should we try to send something over the fifo to exit cleanly? 

It looks like sending a newline would cleanly exit the command handler thread.  We could add something at the bottom of the metadata script that waits for the command thread to exit and then cases the metadata reader thread to exit as well.

streams/dlna_meta.py
```diff
@@ -1,68 +1,210 @@
...
 
-parser = argparse.ArgumentParser(description="dlna metadata interpreter")
-parser.add_argument('dlna_config_dir')#, description='directory used for dlna song info generation')
-args = parser.parse_args()
+# Reads metadata from the upnp AVTransport and writes it to a json file in a loop.
+def metadata_reader(metadata_path: str, album_art_dir: str, service: SSDPDevice, debug: bool = False):
+  """ Read metadata from the upnp AVTransport and write it to a file. """
+  last_file = ''           # last cover art file downloaded
+  stop_counter = 0         # counter to prevent empty metadata on transitions
+  metadata = {"state":"",  # metadata dict to write to file
```
linknum23, line 31: Making this use a dataclass would clean up this code a bit.

streams/dlna_meta.py
linknum23: My initial investigation doesn't show many useful things for this, so no worries if it doesn't make sense.

amplipi/streams.py
linknum23: Or better yet send a close command over the fifo?

linknum23: Recent changes to process handling have added code that looks like this for carefully shutting down processes without leaving any zombies. Here's what the spotify equivalent looks like:

```python3
    if self.proc:
      try:
        self.proc.terminate()
        self.proc.communicate(timeout=3)
      except Exception as e:
        print(f"failed to terminate spotify stream: {e}")
        self.proc.kill()
      self.proc = None
    if self.mpris:
      try:
        del self.mpris
      except Exception:
        pass
      self.mpris = None
``` 

You may want to add some of that idea to this disconnect method.

linknum23 (dismissed):
Done! Sorry this feedback took so long.

amplipi/utils.py
```diff
@@ -393,3 +393,12 @@ def set_identity(settings: Dict):
...
     json.dump(identity, identity_file)
+
+def careful_proc_shutdown(proc: subprocess.Popen, proc_name="process"):
+  """ Shutdown a process, waiting for it to exit """
+  try:
+    proc.terminate()
+    proc.communicate(timeout=3)
+  except Exception as e:
+    print(f"failed to terminate {proc_name}: {e}")
+    proc.kill()
```
rtertiaer, line 404: you still want to `proc.communicate` or `proc.wait` after `kill` to [prevent zombies](https://www.baeldung.com/cs/process-lifecycle-zombie-state).

streams/dlna_meta.py
```diff
@@ -1,68 +1,212 @@
...
-    s1 = line.split('] ')
-    return s1[1]
+        if debug:
+          print("writing metadata", metadata)
+
+        # empty file and write metadata
+        f.truncate(0)
+        f.seek(0)
+        json.dump(metadata, f)
+        f.flush()
```
rtertiaer, line 108: there's not necessarily a TODO here, but file churn like this every 3s on the MMC instead of a tmpfs or similar sketches me out. I'm reading that `flush()` doesn't necessarily mean data ends up on disk (`sync`/`fsync` syscalls do that), but it still reads like this is gonna hit the MMC a ton over time. if you have any quick and clever ideas on where to stash this stuff, such that source configs persist but metadata for these sources does not, and without adding yet another line to fstab, I'm v game

rtertiaer (changes requested):

amplipi/utils.py
rtertiaer: also love this util function <333

streams/dlna_meta.py
rtertiaer: maybe the answer is literally to include a reference to the metadata file used in the persistent source config, stash it in /tmp, and handle the case where it disappears?

klay2000: I was under the impression these were in a tempfs

rtertiaer: this is actually in a tmpfs already, @klay2000 pointed out. nevermind :upside_down_face:

rtertiaer (approved):
:shipit:

amplipi/streams.py
```diff
@@ -848,56 +848,92 @@ def connect(self, src):
...
+    self._uuid = uuid_gen()
     portnum = 49494 + int(src)
 
-    # Make the (per-source) config directory
-    src_config_folder = f'{utils.get_folder("config")}/srcs/{src}'
-    os.system(f'mkdir -p {src_config_folder}')
+    # Make the (per-source) config and web directories
+    self._src_config_folder = f'{utils.get_folder("config")}/srcs/{src}'
+    os.system(f'rm -r {self._src_config_folder}')
+    os.system(f'mkdir -p {self._src_config_folder}')
```
rtertiaer, line 858: nitpick: I'd prefer `os.mkdir()`/`os.rmdir()` but meh.

amplipi/streams.py
```diff
@@ -848,56 +848,92 @@ def connect(self, src):
...
-          source.album = d['album']
-        if 'artist' in d:
-          source.artist = d['artist']
-        if 'title' in d:
-          source.track = d['title']
-        return source
-    except Exception:
+
+      data = json.load(open(f'{self._src_config_folder}/meta.json'))
+      print(data)
```
rtertiaer, line 902:
```suggestion
```

amplipi/streams.py
```diff
@@ -848,56 +848,92 @@ def connect(self, src):
...
+        source.track = data.get('title', '')
+        source.album = data.get('album', '')
+        if data.get('album_art', '') != '':
+          source.img_url = f'generated/{self.src}/{data.get("album_art", "")}'
+
+      source.supported_cmds=self.supported_cmds # set supported commands only if we hear back from the DLNA server
+      self._got_data = True
+    except Exception as e:
+      if self._got_data: # ignore if we havent gotten data yet since we're still waiting for the metadata process to start
+        print(f'Error getting DLNA info: {e}')
```
rtertiaer, line 915: I wonder what will land first - this or #661 :thinking:

rtertiaer (approved):
I mostly only care about the suggestion. Otherwise LGTM; feel free to :shipit: w/ that `print(data)` removed.

amplipi/streams.py
SteveMicroNova: I won the race
Let me know if you need help implementing anything to have this branch be in line with #661, though it's fairly straightforward

streams/dlna_meta.py
```diff
@@ -1,68 +1,214 @@
...
-  return u
-
-f = open(cs_loc, 'w')
-f.write("")
-f.close()
+    if debug:
+      print(f"Found file at {fifo_path}")
+    if os.path.isfile(fifo_path): # pipes return false for isfile
+      print(f"Error: file at {fifo_path} is not a fifo")
+      sys.exit(1)
```
SteveMicroNova, line 128: Make sure this is rebased on main and replace most instances of print with the appropriate [logging level](https://docs.python.org/3/library/logging.html#logging-levels)

Import or create a logger for each file like [this](https://github.com/micro-nova/AmpliPi/pull/661/files#diff-6c7221b76e99dbda44da1b28ea9905978bc07da6bead60b39df334570459e086R84-R85), using `getLogger()` and supplying a name to the logger. If you provide the same name as elsewhere, it will be the same logger and setting the logging level will be unnecessary.

SteveMicroNova (commented):
Not giving explicit Request Changes status as my only issue is a nitpick, an "it would be better if..." and not an "This is broken unless..." kind of change
It would still be better if that change were made, or add an issue that says "add python logging to these files" to say you'll do it another day but are merging now

amplipi/streams.py
```diff
@@ -870,56 +870,87 @@ def connect(self, src):
...
+                 f'{utils.get_folder("streams")}/dlna_meta.py',
+                 f'{self.name}',
+                 f'{self._src_config_folder}/cmd',
+                 f'{self._src_config_folder}/meta.json',
+                 self._src_web_folder,
+                ]
+                # '-d']
+
+    self.proc = subprocess.Popen(args=dlna_args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    self._metadata_proc = subprocess.Popen(args=meta_args, stdin=subprocess.PIPE, stdout=sys.stdout, stderr=sys.stderr)
```
SteveMicroNova, line 904: If this process contains loggers, send the logs to the sys.sdtout (and errors to sys.stderr) directly as I do in LMS._activate() around line 1373:

`self.meta_proc = subprocess.Popen(args=meta_args, stdout=sys.stdout, stderr=sys.stderr)`

If the logs still get output to `journalctl -f` with PIPE, feel free to not change this

SteveMicroNova (approved):
One potential nitpick, test for it and decide whether it's necessary or not but other than that you're gtg for me :+1:

PR #549, opened 2023-12-04
Implement Logitech Media Server in mainsteam AmpliPi

### LINKED ISSUE #532: Mainline LMS
We should be able to toggle LMS mode on & off via a button, either in the main UI or the updater. This will allow us to ditch the `streamer_lms` branch as something we maintain.

rtertiaer:
Filing this as a draft for now. When completed, this ought to close #532 . I'd like some early feedback because we'd like to ship 0.3.0 before the holidays.


amplipi/streams.py
```diff
@@ -1124,9 +1124,10 @@ def info(self) -> models.SourceInfo:
 
 class LMS(PersistentStream):
   """ An LMS Stream using squeezelite"""
-  def __init__(self, name: str, server: Optional[str] = None, disabled: bool = False, mock: bool = False):
+  def __init__(self, name: str, server: Optional[str] = None, disabled: bool = False, mock: bool = False, deallocatable: bool = True):
```
linknum23, line 1127: I was under the impression `deallocateable` would keep streams from being deleted.  At this point it needs to be a property of a base stream.

linknum23: I'm debating this a little now in my head though, because the intention is to keep the LMS stream running in the background.

rtertiaer: I debated with this a bit too and would love to riff on this a bit IRL; is this a property we'd want for streams other than these LMS clients, to create some sort of persistence all the time on other long-running streams?

amplipi/streams.py
```diff
@@ -1186,9 +1187,10 @@ def _activate(self, vsrc: int):
       print(f'error starting lms: {exc}')
 
   def _deactivate(self):
-    if self._is_running():
-      self.proc.kill()
-    self.proc = None
+    if self.deallocatable:
```
linknum23, line 1190: I'm not sure if we want to block this entirely. I'm trying to think through how a soft reset works without this.

amplipi/streams.py
linknum23: We might want this for other streams, although I don't have a concrete use case in mind.

Tomorrow work to riff on this?

linknum23: You can also just call me now

rtertiaer:
per in-person review:

- [ ] remove deallocatable stuff - PersistentStreams class does most or all of what we want
- [ ] Make the LMS mode frontend stuff a toggle-able switch, with state
- [ ] ensure in LMS mode, streams are eagerly started/`activate()`d instead of lazily (when finally attached to a source)

amplipi/ctrl.py
```diff
@@ -335,6 +279,11 @@ def reinit(self, settings: models.AppSettings = models.AppSettings(), change_not
       if stream.id:
         try:
           self.streams[stream.id] = amplipi.streams.build_stream(stream, self._mock_streams)
+          # If we're in LMS mode, we need to start these clients on each boot, not when they get assigned to a
+          # particular source; the client+server connection bootstrapping takes a while, which is a less than ideal
+          # user experience.
+          if self.lms_mode and stream.type == 'lms':
+            self.streams[stream.id].activate()
```
rtertiaer, line 286: I considered doing this in the constructor but the controller is the bit that knows & cares about if it's in LMS mode.

amplipi/streams.py
```diff
@@ -1202,8 +1200,7 @@ def info(self) -> models.SourceInfo:
     return source
 
   def disconnect(self):
```
linknum23, line 1202: This already exists in the base class

linknum23: Just remove this method

scripts/configure.py
```diff
@@ -159,6 +159,14 @@
   'lms' : {
     'apt': ['libcrypt-openssl-rsa-perl'], # needed for ShairTunes2W support
     'copy' : [{'from': 'bin/ARCH/squeezelite', 'to': 'streams/squeezelite'}],
+    'script' : [
```
linknum23, line 162: Should this upgrade older versions?

linknum23: Its really fine if it doesnt for now, but we may want to add support for this in the next release 0.3.1?.

rtertiaer: I think we should manage a full-blown repository moving forward, which will address this. There are a few cases in the world where not upgrading versions is plainly a problem, but those cases are also aware that they need to reflash to get the bugfixes in latest LMS anyhow (and the 0.3.0 image we produce will have this version.)

linknum23:
Testing this on `lamplipi` shows the lms clients dropping after a reset. Heres the sequence I tested.
1. Enable LMS mode
2. Set each source to its corresponding RCA input
3. Went to config->reset
4. All LMS clients are missing now from the LMS Server UI hosted on port 9000

rtertiaer:
@linknum23 I was not able to replicate your problem above. I'd love to hash this out a smidge in person, because there's some wackiness with our multiple test beds that I think is complicating things a little bit.

rtertiaer:
@linknum23 I think the thing you bumped into is fixed now. The bug was that every other restart of `squeezelite` leaves the ALSA loopback in a funny state and it cannot properly bind to the 'device'; it then simply bails with an error. The fix ultimately required us building our own `squeezelite` .deb for [an upstream commit](https://github.com/ralph-irving/squeezelite/commit/c9027119369caf85f02952d009420fb74c51b994), which permitted us to manage `squeezelite` using `process_monitor.py`.

I'm going to continue testing this.

linknum23 (approved):
This PR really needs #557 for zones to work right but as is it is pretty fantastic. Nice work @rtertiaer

PR #550, opened 2023-12-04
Update how-to GIFs

SteveMicroNova:
Move contents of docs/imgs/app_demos to docs/imgs/app_demos/old

Add new GIFs to docs/imgs/app_demos featuring webapp2

Replace all references to old GIFs with references to new GIFs


SteveMicroNova:
There isn't a real way to test this I don't think, if it isn't sized properly we'll just have to use the gif resizer script I found while searching for all references to ".gif" in this repo

rtertiaer (changes requested):
I'd much prefer the old gifs simply get removed. They'll still be in the git history, but otherwise they serve no documentation purposes for the currently living version of the webapp - they only serve to take up logical and digital space.

wrt image sizes, all the old gifs are much much smaller - the largest is 1.3M, and the average is about 300K. we definitely need to aim for a similar size for these image; it's important for folks on less than stellar connections (think LTE modems) that these load quickly, and for example Google considers how responsive a website loads (ie how reasonable its images are sized) when recommending web content. Google recommends [this resource for optimising images](https://web.dev/explore/fast#optimize-your-images); I'd totally be game for moving to a nice embedded video format to keep these sizes small, especially since they're very well supported in 2023.

rtertiaer (commented):
why's the "gif creation process" doc being deleted? it doesn't seem like there's a replacement document.

rtertiaer (changes requested):
One last thing; can you squash all these commits? otherwise we have these 10mb images hanging out in the git history, but they live for a commit or two, just serving to take up space and bloat our repo.

linknum23:
I would like to suggest a better commit message than Squash... Maybe "Update Web interface how-to GIFs"?

linknum23 (approved):
This all looks great. I would suggest fixing a couple of the nitpicks but I think this documents the new interface well.

PR #557, opened 2023-12-07
Actually update the zone's source id

linknum23:
Fixes delayed source id issue


linknum23:
@klay2000 Can you play with adding audio to different zones and make sure this fixes stuff.

klay2000 (approved):
Fixes the bug but we might want to add a test in the future to prevent it being added back.

PR #559, opened 2023-12-08
Switch to trunk based development & explicit semantic versioning

rtertiaer:
This PR updates our release documentation. This came up because of some oddities during the stamping of the 0.3.0 release candidate 0. 

I'm trying to more closely follow a [more conventional trunk-based development](https://trunkbaseddevelopment.com/), where there exists only one trunk and version branches never get merged back into `main`. As the doc was before this PR, we'd at least be merging compiled javascript back into `main` regularly, which is no bueno. I think it's also a bit simpler and cleaner.

### DIFF
--- a/NEW_RELEASE.md
+++ b/NEW_RELEASE.md
@@ -2,19 +2,6 @@
 
 So you think you are ready to make a release, eh? Follow the steps below :)
 
-## Making a prelease
-- [ ] Merge main into develop
-- [ ] Make a branch off develop called VERSION-prelease
-- [ ] Update the changelog with notes and the latest version (and commit it!)
-- [ ] Update the API by running the `scripts/create_spec` script and commit the resulting changes to `docs/amplipi_api.yaml` with `git commit --patch -m "Update API specification used by GitHub Pages"`
-- [ ] Build the webapp with `npm run build` and force add the changes with `git add -f web/dist`
-- [ ] Use poetry to bump the version with `poetry version ${VERSION}-alpha.0 && git add pyproject.toml && git commit -m "Bump pre-release version"`
-- [ ] Tag the changes so we can make a pre-release on GitHub `git tag -as ${VERSION}-alpha.0 -m '' && git push origin ${VERSION}-alpha.0`
-- [ ] Make a pre-release using the GitHub interface
-- [ ] Use the AmpliPi updater to update to the pre-release
-- [ ] Test it (see below), if it needs changes follow the steps above again but stay on the pre-release branch and increment the number after alpha eg `alpha8`
-- When it works skip below to [Finalizing the release](#finalizing-the-release)
-
 ## Testing
 - [ ] Web App (use chrome, firefox, and safari)
   - [ ] Test group and zone volume control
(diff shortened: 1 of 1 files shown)


NEW_RELEASE.md
```diff
@@ -4,10 +4,10 @@ So you think you are ready to make a release, eh? Follow the steps below :)
 
 ## Making a prelease
 - [ ] Merge main into develop
-- [ ] Make a branch off develop called VERSION-prelease
+- [ ] Make a trunk branch off develop using the version number, ie `0.3.0`
```
Lohrer, line 7: Shouldn't this be a version branch? The _trunk_ branch for us is `main` unless I'm misunderstanding your proposed new workflow.

rtertiaer: I've gone slash & burn with this doc; have a gander at the latest version :)

linknum23:
Given our constraint of needing to check in "compiled" javascript for the
next couple of releases I think this makes sense from a high level.

rtertiaer:
@linknum23 @Lohrer is there further discussion to be had here, or is this approved/denied?

I ask cuz I just merged https://github.com/micro-nova/AmpliPi/pull/575 into the wrong trunk :upside_down_face:

NEW_RELEASE.md
```diff
@@ -34,11 +21,19 @@ So you think you are ready to make a release, eh? Follow the steps below :)
...
-- [ ] merge the prerelease branch into develop and merge that into the main branch
-- [ ] use poetry to bump the version `poetry version ${VERSION} && git add pyproject.toml && git commit -m "Bump release"`
-- [ ] use git to delete any prelease tags `git push origin -d $(git tag -l "${VERSION}-alpha*") && git tag -d $(git tag -l "${VERSION}-alpha*")`
-- [ ] Tag the changes so we can make a release on GitHub `git tag -as ${VERSION} -m '' && git push --follow-tags`
-- [ ] Make a release using the GitHub interface, add notes from the CHANGELOG.md
-- [ ] Make a new Raspberry Pi image to program onto units before shipping.
+## Versioning
+This project follows [Semantic Versioning](https://semver.org/). Here are some examples of versions we've used before:
+* `0.2.1`: this is a public general-availability bugfix release of the `0.2` feature release.
+* `0.3.0-alpha.0` is the first alpha release of the `0.3` feature release.
```
linknum23, line 27: probably add beta and rc examples as well

linknum23:
The only issue here is that some documentation that links to github may be ahead of the current release. We may be able to handle those things on a case by case basis though.

linknum23 (approved):

rtertiaer:
> The only issue here is that some documentation that links to github may be ahead of the current release. We may be able to handle those things on a case by case basis though.

there are two ways of handling this I think. one, we could generate docs on a version-by-version basis, like a drop down menu up top of a generated site. additionally, putting a note besides each new feature or edgecase that we program for stating which version it was introduced or changed is very helpful. you can see both examples of this setup in [the Python docs](https://docs.python.org/3/library/os.html#os.environb) (the bits that say "_New in 3.9_", and the dropdown up top.) All that's required of us is to regularly maintain our manual with changes (hopefully the PR checklist helps) and one day get good at generating documentation sites, instead of just leaning on github (but that's definitely a future TODO.)

NEW_RELEASE.md
```diff
@@ -34,11 +21,19 @@ So you think you are ready to make a release, eh? Follow the steps below :)
...
+
+## Making a release
+- [ ] Ensure the PR(s) with your features & fixes are merged into `main`.
+- [ ] Check out and pull `main`
+- [ ] Update the changelog with notes and the latest version (and commit it!)
```
Lohrer, line 32: This reads as if the changelog/notes should be committed directly onto main, but that's not possible right?

rtertiaer: ope, good catch. I think the latest rendition of this oughta be "more correct".

Lohrer (approved):
Seems reasonable to me, but I've never created a release so I'd suggest waiting on the wisdom of Lincoln.

PR #562, opened 2023-12-11
Do not After=network.target the -display & audiodetector services

rtertiaer:
During some network oddities where the network initialization does not start or fail quickly but hangs for a long time, the display and audiodetector services also hang. I've removed that "weak dependency". This potentially allows us to give better feedback during bootup.

This still needs to be tested, in particular because of the display service potentially having a true dependency on networking at the moment; thus, I've filed this as a draft for now.

### DIFF
--- a/scripts/configure.py
+++ b/scripts/configure.py
@@ -502,7 +502,6 @@ def _display_service(directory: str):
   return f"""\
 [Unit]
 Description=Amplipi Front Panel Display
-After=network.target
 
 [Service]
 Type=simple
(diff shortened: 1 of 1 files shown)


rtertiaer:
With un-bonded grounds, the behavior of the display did not change. With bonded grounds, this briefly starts with "None" for the interface stuff, but otherwise boots up normally & "just works".

I think we're also guarded against any potential exceptions with systemd restarting the service over time.

But all in all, I can't demonstrate that this PR changes behavior at all, which could be good or pointless in this context :joy: It still feels more correct to have these services not necessarily wait for the network to be available; they should not have any hard dependencies on the network.

Lohrer (approved):

PR #565, opened 2023-12-11
Clean up running spotify instances when removed from configuration (#564)

### LINKED ISSUE #564: Ensure persistent streams are fully deallocated when their definitions are removed from the configuration

rtertiaer:
This PR fixes #564 , as well as cleans up a couple of things. The bug #564 was that even when a spotify stream was deleted from the configuration, it ran in the background. This was due to the `Spotify(PersistentStream)` class overriding `PersistentStream`'s `__del__`. I've removed `Spotify`'s `__del__`; [it appears to be detritus from before it was a `PersistentStream`](https://github.com/micro-nova/AmpliPi/commit/860e6f34c132b5b7e05788d40d66a71612911b88).

### DIFF
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -284,6 +284,7 @@ class PersistentStream(BaseStream):
       try:
         # must use terminate as kill() cannot be intercepted
         self._cproc.terminate()
+        self._cproc.communicate(timeout=5)
 
       except Exception as e:
         print(f'PersistentStream disconnect error: {e}')
@@ -511,7 +512,6 @@ class Spotify(PersistentStream):
 
     self.connect_port: Optional[int] = None
     self.mpris: Optional[MPRIS] = None
-    self.proc_pid: Optional[int] = None
     self.supported_cmds = ['play', 'pause', 'next', 'prev']
 
   def reconfig(self, **kwargs):
@@ -523,9 +523,6 @@ class Spotify(PersistentStream):
       reconnect_needed = True
     if reconnect_needed and self.is_activated():
       self.reactivate()
-
-  def __del__(self):
-    self.disconnect()
 
   def is_persistent(self):
     return True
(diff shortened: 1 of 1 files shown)


amplipi/streams.py
```diff
@@ -284,6 +284,7 @@ def disconnect(self):
       try:
         # must use terminate as kill() cannot be intercepted
         self._cproc.terminate()
+        self._cproc.communicate(timeout=5)
```
rtertiaer, line 287: `terminate()` just fires off a signal; `communicate` actually prevents zombies from piling up. Doing things this way can also more appropriately bubble up errors caught during a timeout exception. This was ultimately not the cause of the issue tho, just a worthy improvement.

amplipi/streams.py
```diff
@@ -511,7 +512,6 @@ def __init__(self, name: str, disabled: bool = False, mock: bool = False):
 
     self.connect_port: Optional[int] = None
     self.mpris: Optional[MPRIS] = None
-    self.proc_pid: Optional[int] = None
```
rtertiaer, line 514: This is unused.

amplipi/streams.py
```diff
@@ -524,9 +524,6 @@ def reconfig(self, **kwargs):
     if reconnect_needed and self.is_activated():
       self.reactivate()
 
-  def __del__(self):
-    self.disconnect()
-
```
rtertiaer, line 529: This is the actual bugfix.

amplipi/streams.py
```diff
@@ -572,9 +569,11 @@ def _activate(self, vsrc: int):
...
   def _deactivate(self):
     try:
-      self.proc.kill()
-    except Exception:
-      pass
+      self.proc.terminate()
+      outs, errs = self.proc.communicate(timeout=3)
+    except Exception as e:
+      print(f"failed to terminate spotify stream: {e}")
+      kouts, kerrs = self.proc.kill()
```
rtertiaer, line 576: same as above - just improved process handling.

linknum23 (approved):
This makes a lot of sense. Should we "robustify" other streams using the terminate/communicate paradigm also?

Nice catch.

PR #566, opened 2023-12-12
fix regression introduced in Spotify bugfix

rtertiaer:
This PR fixes a bug introduced during #565 ; we don't use the outputs of `communicate`, and `kill` doesn't even output. The tests failed on this; it's not clear why Github let me merge anyhow, when our branch protections require passing tests for everybody, including administrators :woozy_face:

### DIFF
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -570,10 +570,10 @@ class Spotify(PersistentStream):
   def _deactivate(self):
     try:
       self.proc.terminate()
-      outs, errs = self.proc.communicate(timeout=3)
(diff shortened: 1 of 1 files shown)


linknum23 (commented):
I think this needs further explanation. Why was it failing?

linknum23:
TLDR: It worked fine but the linter caught some unused and non-existent return values which were reported as errors.

rtertiaer:
this is deployed & waiting on others to test on the appliance in-house; then stamping out 0.3.0 :confetti_ball:

Lohrer (approved):

Issue #542, opened 2023-12-04
Create 0.3.0 image


Lohrer:

rtertiaer:
shipped a 0.3.1 image. this is complete~

PR #575, opened 2023-12-19
Allow configure.py to care less about being run on a streamer or not

### LINKED ISSUE #573: Audiodetector service ought to be installed on all images, but 'gatekept' from boxes without RCA in
Right now, Audiodetector is gatekept during `configure.py`, but this should not be the only place it should be gatekept; we ought to be able to install one image on all boards and have it function correctly. Either modify the service or the `.service` to handle the case where there are no RCA inputs.

rtertiaer:
This PR allows `configure.py` to care less about it being run on a streamer. This permits us to generate one system image that can run on both types of appliances. The changes aren't hefty:
* deploy all test shortcuts on all appliance types, regardless of their relevance for that particular type. This does not affect end users, but internally we'll need to communicate with manufacturing (@mrahman313 :eyes: )
* deploy the `audiodetector` service on everything, even boxes without RCA in - but gatekeep the service launch with a conditional.

I've tested this on both regular and streamer appliances. The test shortcuts get deployed everywhere. The `audiodetector` service correctly does not launch on streamers, wie so:
```sh
pi@amplipi:~ $ systemctl --user status amplipi-audiodetector.service 
● amplipi-audiodetector.service - Amplipi RCA Input Audio Detector
   Loaded: loaded (/home/pi/.config/systemd/user/amplipi-audiodetector.service; enabled; vendor preset: ena[...]
   Active: inactive (dead)
Condition: start condition failed at Tue 2023-12-19 18:33:30 EST; 1min 17s ago
           └─ ConditionPathExists=!/home/pi/.config/amplipi/is_streamer was not met

Dec 19 18:33:30 amplipi systemd[583]: Condition check resulted in Amplipi RCA Input Audio Detector being sk[...]
```

This PR fixes #573 , which came up during image creation, #542.

### DIFF
--- a/scripts/configure.py
+++ b/scripts/configure.py
@@ -517,6 +517,7 @@ def _audiodetector_service(directory: str):
   return f"""\
 [Unit]
 Description=Amplipi RCA Input Audio Detector
+ConditionPathExists=!/home/pi/.config/amplipi/is_streamer
 
 [Service]
 Type=simple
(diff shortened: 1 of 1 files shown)


scripts/configure.py
```diff
@@ -517,6 +517,7 @@ def _audiodetector_service(directory: str):
   return f"""\
 [Unit]
 Description=Amplipi RCA Input Audio Detector
+ConditionPathExists=!/home/pi/.config/amplipi/is_streamer
```
rtertiaer, line 520: Documentation [here](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#ConditionPathExists=).

linknum23 (approved):
LGTM. The easiest way to validate this works is to upgrade a streamer and then yank that CM3+ and put it in a main unit. Please try something like that before merging this.

rtertiaer:
spoke IRL; since I've already tested on existing appliances and we'll be doing further testing during the pressing of 0.3.1, we can ship this right now.

PR #578, opened 2023-12-20
Update production tests

Lohrer:
- Updated the test sheets
- Fixed the "program expander" and "program 2nd expander" tests that 100% just didn't work.
- Removed optical input from tests
- Fans/power/temp test now asks if a main unit or expander is being tested, and only checks that unit instead of both.
- Handle SIGINT in USB test to allow exiting before the timeout.
- Prevent multiple SIGINTs from causing multiple attempts at killing processes in `tests.py`
- Also updated VS code python formatting and linting settings. The Python plugin spun off the formatting and linting into separate plugins.


amplipi/tests.py
```diff
@@ -4,21 +4,23 @@
...
 from typing import Optional, Sequence
-import sys
-import subprocess
-import signal
 import argparse
 import requests
+import shlex
+import signal
+import subprocess
+import sys
```
rtertiaer, line 12: :sunglasses:

amplipi/tests.py
```diff
@@ -62,37 +64,40 @@ def available(self) -> bool:
     except:
       return False
 
-RCA_INPUTS = { sid: 996 + sid for sid in range(models.MAX_SOURCES) }
+
+RCA_INPUTS = {sid: 996 + sid for sid in range(models.MAX_SOURCES)}
```
rtertiaer, line 68: suggestion: it may (or may not) make more sense to `from amplipi.defaults import RCAs` and use that instead of this.

amplipi/tests.py
```diff
@@ -103,64 +108,67 @@ def pst_all_zones_to_src(name: str, src: int, _input: str, vol=-50):
...
+    presets += [pst_all_zones_to_src('aux-in', 0, f'stream={AUX_PLAYBACK["id"]}', -40)]
 
   prev_cfg = client.get_status()
   if is_streamer:
-    client.load_config(models.Status(zones=[], streams=[BEATLES_RADIO, EXTRA_INPUTS_PLAYBACK]))
+    client.load_config(models.Status(zones=[], streams=[BEATLES_RADIO, AUX_PLAYBACK]))
   else:
-    client.load_config(models.Status(zones=[models.Zone(id=z, name=f'Zone {z + 1}') for z in all_zones(exp_unit)], streams=[BEATLES_RADIO, EXTRA_INPUTS_PLAYBACK]))
+    zones = [models.Zone(id=z, name=f'Zone {z + 1}') for z in all_zones(exp_unit)]
+    client.load_config(models.Status(zones=zones, streams=[BEATLES_RADIO, AUX_PLAYBACK]))
```
rtertiaer, line 164: :relieved:

rtertiaer (approved):
a single suggestion. the case could be made to either implement it or keep it as-is.

love your ~aesthetic~, it's pleasing to read.

amplipi/tests.py
Lohrer: These 4 imports make me feel like I should just be using Bash

amplipi/tests.py
Lohrer: I'll take a look at that. I don't think that existed when I wrote this. This change is only popping up here because of auto-format actually kicking in now that I fixed the things VS Code has been yelling at us about for ~weeks~ ~months~ years.

Lohrer:
Apparently I broke mypy with a function attribute so hang tight while I blow everything up into a full on class...

amplipi/tests.py
rtertiaer: it definitely did not exist before a month ago even. you could definitely leave it for future refactor, np.

Issue #579, opened 2023-12-21
Fix service advertisement interface selection


Lohrer:
The zeroconf test correctly assumes the default interface should be used to advertise the amplipi service, but AmpliPi itself chooses the first available interface. On AmpliPis as-shipped this should always work, but on development computers or if a USB network adapter is plugged in things can go wrong.

Issue #432, opened 2023-01-06
Handle start before DHCP resolution


linknum23:
in some logs we have seen the following failure in our MDNS advertisement:

```
Dec 22 01:06:22 amplipi authbind[807]: AmpliPi zeroconf - unable to register service on one of ['eth0', 'wlan0', 'eth0'],             they all are either not available or have no IP address.
Dec 22 01:06:22 amplipi authbind[807]: AmpliPi zeroconf - is this running on AmpliPi?
Dec 22 01:06:22 amplipi authbind[807]: AmpliPi zeroconf - registering service: ServiceInfo(type='_amplipi._tcp.local.', name='amplipi-none._amplipi._tcp.local.', addresses=[b'\x00\x00\x00\x00'], port=80, weight=0, priority=0, server='amplipi.local.', properties={'path': '/api/', 'name': 'AmpliPi', 'vendor': 'MicroNova', 'version': '0.1.9', 'web_app': 'http://amplipi.local/', 'documentation': 'http://amplipi.local/doc'}, interface_index=None)
Dec 22 01:06:22 amplipi authbind[807]: Process Process-1:
Dec 22 01:06:22 amplipi authbind[807]: Traceback (most recent call last):
Dec 22 01:06:22 amplipi authbind[807]:   File "/usr/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
Dec 22 01:06:22 amplipi authbind[807]:     self.run()
Dec 22 01:06:22 amplipi authbind[807]:   File "/usr/lib/python3.7/multiprocessing/process.py", line 99, in run
Dec 22 01:06:22 amplipi authbind[807]:     self._target(*self._args, **self._kwargs)
Dec 22 01:06:22 amplipi authbind[807]:   File "./amplipi/app.py", line 870, in advertise_service
Dec 22 01:06:22 amplipi authbind[807]:     zeroconf = Zeroconf(ip_version=IPVersion.V4Only, interfaces=[ip_addr])
Dec 22 01:06:22 amplipi authbind[807]:   File "/home/pi/amplipi-dev/venv/lib/python3.7/site-packages/zeroconf/_core.py", line 410, in __init__
Dec 22 01:06:22 amplipi authbind[807]:     listen_socket, respond_sockets = create_sockets(interfaces, unicast, ip_version, apple_p2p=apple_p2p)
[...]

PR #433, opened 2023-01-06
Handle start before DHCP resolution

linknum23:
Handle start before DHCP resolution
closes #432 

Also try and include #579

PR #583, opened 2023-12-21
Handle start before DHCP resolution 

linknum23:
Fixes https://github.com/micro-nova/AmpliPi/issues/432

* [x]  fix https://github.com/micro-nova/AmpliPi/issues/579
 
### Checklist

* [x] Update the CHANGELOG

Also try and include https://github.com/micro-nova/AmpliPi/issues/579

### DIFF
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -22,6 +22,8 @@ This serves the amplipi webpp and underlying rest api, that it uses.
 The FastAPI/Starlette web framework is used to simplify the web plumbing.
 """
 
+import logging
+import sys
 import argparse
 
 import os
@@ -55,7 +57,8 @@ from sse_starlette.sse import EventSourceResponse
 import netifaces as ni
 from socket import gethostname, inet_aton
 from zeroconf import IPVersion, ServiceInfo, Zeroconf
-from multiprocessing import Queue
+from multiprocessing import Event, Queue
+from multiprocessing.synchronize import Event as SyncEvent
 
 # amplipi
 import amplipi.utils as utils
@@ -888,6 +891,14 @@ def on_shutdown():
 # MDNS
 
 
+# TEST: use logging in this module
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
+sh = logging.StreamHandler(sys.stderr)
+sh.setFormatter(logging.Formatter('%(name)s: %(levelname)s - %(message)s'))
+log.addHandler(sh)
+
+
 def get_ip_info(iface: str = 'eth0') -> Tuple[Optional[str], Optional[str]]:
   """ Get the IP address of interface @iface """
   try:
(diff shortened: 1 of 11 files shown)


linknum23:
Note: this was copied from #433

linknum23:
@Lohrer I believe #579 is solved as well. 

As for testing:

This will need some mileage before merging. Specifically looking at logs after a restart, logs after yanking the ethernet interface, other similar tests.

amplipi/app.py
```diff
@@ -897,60 +898,78 @@ def get_ip_info(iface: str = 'eth0') -> Tuple[Optional[str], Optional[str]]:
...
+def _find_best_iface(ifaces: List[str], ok: Callable) -> Union[Tuple[str, str, str], Tuple[None, None, None]]:
+  # Try to find the best interface to advertise on,
+  # retrying in case DHCP resolution is delayed
+  ip_addr, mac_addr = None, None
+  retry_count = 5
+  while ok() and not (ip_addr and mac_addr) and retry_count > 0:
+    for iface in ifaces:
+      ip_addr, mac_addr = get_ip_info(iface)
+      if ip_addr and mac_addr:
+        return ip_addr, mac_addr, iface
```
Lohrer, line 940: Calling the first interface with an IP address the 'best' interface is probably valid in most cases, but this still fails on my computer. The first Ethernet port, call it eth0, has a local address with no internet, so it's not the default gateway. The second Ethernet port, call it eth1, is the default gateway. `test_rest.py:test_zeroconf()` advertises on ni.gateways()['default'], so eth1. AmpliPi advertises on the alphabetically first eth0, so the test fails.

A default gateway isn't guaranteed to exist, for example if a computer and AmpliPi are connected link-locally. But clearly selecting the first IP address isn't perfect either. I'd suggest checking for a default gateway and if present using that. If not present, then fall back to the first IP address method here.

In the case where someone has plugged in a USB network adapter into an AmpliPi (most likely WiFi but could be Ethernet) and they both have valid routes to a real network, I'm not sure we can do anything so I guess I'm not worried about handling that. Unless we can do multiple advertisements if there are multiple valid interfaces with IPs?

amplipi/app.py
```diff
@@ -897,65 +898,83 @@ def get_ip_info(iface: str = 'eth0') -> Tuple[Optional[str], Optional[str]]:
...
+def _advertise_service(port: int, ok: Callable) -> None:
+  """ Underlying advertisement, can throw Exceptions when network is not connected """
+
   hostname = f'{gethostname()}.local'
   url = f'http://{hostname}'
 
-  # search for the best interface to advertise on
+  # enumerate interface to advertise on, starting with the pi defaults first
+  # ordering is important, as the pi defaults will be tried first
   ifaces = ['eth0', 'wlan0']  # default pi interfaces
```
Lohrer, line 954: Do we need to list these? If the issue here is them not showing up right away at boot, would it be better for _find_best_iface to dynamically enumerate the interfaces just like it is doing for IPs?

Lohrer (changes requested):
I'd like to test a thing or two on a real AmpliPi but long story short my AmpliPi is down due to not being able to build the web app. This looks like a good improvement so far though.

amplipi/app.py
linknum23: I dumped this in the latest version and switched to depending on the default gateway.

linknum23:
Alright I think this thing is ready!

Here's the new changes:
- `_find_best_iface()` now prefers the interface default gateway
- MDNS advertisement was migrated to its own file (it was completely separate from app so this was not a big deal)
- built-in python logging was added as a test, we will likely use this throughout AmpliPi in the future, for now it works as a drop in replacement

This was tested on the following sequence:
1. Boot AmpliPi connected to the main network
2. Disconnect
3. Connect again
4. Disconnect
5. Connect over laptop to link local connection
6. Disconnect
7. Connect to main network again
8. Disconnect
9. Reboot
10. Wait for system to start +1min
11. Connect to main network again

On each new connection the advertisement was able to identify its change of state and advertise its new ip address (if it had one) if it didnt have one it failed and waited 30 seconds to retry

amplipi/mdns.py
Lohrer, line 1: While this is _using_ MDNS, the only thing setup in here is DNS Service Discovery (DNS-SD). Unless we're going to have to implement other MDNS-related things in here in the future, I believe it would be more accurate to label this as either zeroconf or dns_sd or something like that.

linknum23: I'm open to whatever with a small preference to a shorter name.

linknum23:
@rtertiaer Can you check out how I used logging here? Also make sure the interface resolution make sense?

amplipi/mdns.py
```diff
@@ -0,0 +1,179 @@
...
+    # this attempts to handle events like router/switch resets
+    while ok():
+      delay_ct = 100
+      while ok() and delay_ct > 0:
+        sleep(0.1)
+        delay_ct -= 1
+      if ok():
+        new_ip_addr, _, new_iface = _find_best_iface(ok)
+        if new_ip_addr != ip_addr:
+          log.info(f'IP address changed from {ip_addr} ({good_iface}) to {new_ip_addr} ({new_iface})')
```
Lohrer, line 173: pylint is warning about f-strings in logging. Seems like the TLDR is it's more performant to pass log.info('Message %s', variable) since then if the logging level is changed the f-string doesn't have to be evaluated. It also means if an error occurs during evaluation of the f-string it gets logged. I suspect these aren't good enough reasons for us to have a second method of string formatting to be remembered, so I added `disable=logging-fstring-interpolation` to our .pylintrc. Feel free to remove this override and fix the logging strings if you feel otherwise. See here for some of the discussion on this: https://github.com/pylint-dev/pylint/issues/1788

amplipi/mdns.py
```diff
@@ -0,0 +1,179 @@
...
+from zeroconf import IPVersion, ServiceInfo, Zeroconf
+
+from amplipi import utils
+
+# TEST: use logging in this module
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
+sh = logging.StreamHandler(sys.stderr)
+sh.setFormatter(logging.Formatter('%(name)s: %(levelname)s - %(message)s'))
+log.addHandler(sh)
```
rtertiaer, line 38: :sunglasses:

amplipi/mdns.py
rtertiaer: > so I added `disable[...]`

:clap:

rtertiaer (approved):

amplipi/mdns.py
linknum23: Updated to zeroconf

amplipi/zeroconf.py
```diff
@@ -0,0 +1,179 @@
...
+from typing import Callable, Optional, Tuple, Union
+
+import netifaces as ni
+from zeroconf import IPVersion, ServiceInfo, Zeroconf
+
+from amplipi import utils
+
+# TEST: use logging in this module
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
```
rtertiaer, line 35: nitpick: I think about the only thing I might change about the logging would be to either:
1) not set the log level at all. this defaults to `logging.WARNING`, _but_ permits us to set the log level elsewhere/at the root logger (notably during app startup). Should we fully migrate to ussing `logging` at some point, we'll probably poke this then, which is why this is a nitpick, OR
2) use an environment variable to expose this straight away and set a default to `logging.INFO` when it is unset. I like this less, but gives us some flexibility straight away.

personally? I'd probably just leave this as-is til we migrate all `print` statements.

Lohrer (approved):

PR #585, opened 2023-12-22
Update requests in requirements.txt to avoid vulnerability

Lohrer:
I recently discovered https://github.com/micro-nova/AmpliPi/security/dependabot, and while I don't know anything about `npm` I took a look at the `pip` packages. Most of them aren't upgradeable until we get a new Python version, but `requests` can be upgraded to avoid https://github.com/micro-nova/AmpliPi/security/dependabot/4. 
 
### Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ pytest-dependency==0.5.1
 python-multipart==0.0.5
 python-vlc==3.0.18121
 PyYAML==6.0
-requests==2.28.2
+requests==2.31.0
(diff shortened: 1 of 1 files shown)


rtertiaer (approved):
our usage of the `requests` api appears to be explicitly the .get and .post methods, and some exception names. these are extremely unlikely to have broken within the 2.x series (if they semantically version) and mypy would catch that; i feel comfy shipping it with very little extra testing help on our part.

Issue #591, opened 2023-12-28
Webapp build instructions


Lohrer:
I tried to run the deploy script but it fails with `./scripts/deploy: line 150: npm: command not found`. Currently, for other dependencies the deploy script does it's best to auto-install, but at the very least some info about what's missing should be printed. For example, which npm version to install? On Debian/Ubuntu install with an apt package? Pip? Something else?

I did check doc/DEVELOPING.md and web/readme.md for any info on npm or building the webapp, but couldn't find anything.

I then tried to build on an AmpliPi with the steps in scripts/deploy, figuring the npm installed on AmpliPi (5.6.0) would work:
```sh
cd amplipi-dev/web
npm install
# npm run build
```

npm install failed when trying to install `esbuild@0.18.20`.

So I would like to request some documentation on how to build the webapp and/or run `scripts/deploy` to install a development AmpliPi version.

linknum23:
I have fought with node in the past and ended up using this thing called nvm to install a specific version of node: https://github.com/nvm-sh/nvm

It looks like my system uses node **v14.15.0** to build the webapp. So node 14 is probably sufficient.

So installing node with nvm looks something like this:
```
# install the latest? nvm to manage versions of node
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
# install and use node 14
nvm use 14
```

Installing this added silly stuff to my .bashrc. I made sure to add a comment indicating its ridiculousness so I wouldn't delete it.
```
# js is silly
 export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
 [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
```

linknum23:
There may be a better way to do this.

rtertiaer:
Good catch! this appears to be something missed from #533 . 

I'll add this to the 0.3 milestone; should be a simple fix to at least check and document this better.

Lohrer:
This is being added to DEVELOPING.md but to echo it here as well, turns out we need at least Node.js >=18. Anything older is deprecated. If the version of `nodejs` in apt is outdated, using NVM as Lincoln suggests is an easy alternative. Not trusting 'random script from the internet', on Ubuntu 20.04 I followed NVM's readme and simply ran:

```sh
[...]

PR #592, opened 2024-01-02
Update scripts/deploy to check for & possibly install npm

rtertiaer:
This PR fixes issue #591 , by having the deploy script check for the existence of `npm` and potentially informing the user to install it.
 
### Checklist

### DIFF
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -104,7 +104,7 @@ if ! which npm; then
   echo "npm & nodejs needs to be installed to use this script"
   exit 1;
 fi
-if [ $(nodejs --version | tr -d v | cut -f1 -d.) -lt 18 ]; then
+if [ $(node --version | tr -d v | cut -f1 -d.) -lt 18 ]; then
   echo "nodejs needs to be version 18 or above to use this script"
   exit 1;
 fi
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -71,7 +71,10 @@ echo -e "\nChecking dependencies"
 inst=false
 python='python'
 if which dpkg; then
-  # since dpkg exists lets use it to install python dependencies as system dependencies
+  # since dpkg exists lets use it to install deps as system packages
+  # to note: should any of these be installed and then removed again, dpkg-query will
+  # still exit with a 0 status due to the package being in the "residual-config" state.
+  # regardless, this ought to cover >90% of cases.
   for dep in python3-venv python3-pip; do
     dpkg-query -s $dep 1>/dev/null 2>/dev/null || inst=true
   done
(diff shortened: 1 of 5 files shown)


scripts/deploy
```diff
@@ -94,6 +97,14 @@ else
     echo "Failed to install python virtual environments, you may need to install this manually"
     exit 1;
   fi
+  if ! which npm; then
+    echo "npm needs to be installed to use this script"
+    exit 1;
+  fi
+  if [ $(npm version | cut -f1 -d.) -lt 9 ]; then
+    echo "npm needs to be version 9 or above to use this script"
```
rtertiaer, line 105: truly unclear what version of `npm` works as a minimum, but it definitely works on 9 and definitely breaks on 5. not terribly interested in spending a bunch of time to find out the minimum either... @linknum23 what's on your localhost? lol

linknum23: v9.11.2

linknum23: With 0.11.2: `npm does not support Node.js v9.11.2`

rtertiaer: we established @linknum23 actually runs npmv major version 6, _and_ that he does not want his machine's built-in to be targeted because he uses `nvm`. I think it's sane enough to just ship with `npm` >= 9.

linknum23: I think checking and failing if npm version is too old makes more sense for npm. Trying to install it from apt is complicated on some machines that aren't mine.

rtertiaer: yeah, it's a mess. alright, I moved it to its own conditional check in the latest revision.

scripts/deploy
```diff
@@ -96,6 +99,20 @@ else
   fi
 fi
 
+# check for `npm`
+if ! which npm; then
```
Lohrer, line 103: Should we check for `node` here as well?

scripts/deploy
```diff
@@ -96,6 +99,20 @@ else
...
+  echo "npm needs to be installed to use this script"
+  exit 1;
+fi
+if [ $(npm --version | cut -f1 -d.) -lt 9 ]; then
+  echo "npm needs to be version 9 or above to use this script"
+  exit 1;
+fi
+if [ $(nodejs --version | tr -d v | cut -f1 -d.) -lt 18 ]; then
```
Lohrer, line 111: I don't know the 'usual' way on installing node/npm, but with `nvm` the node version is what is selected. So it's hard to tell from the "npm needs to be version 9 or above to use this script" message which node version I should install. Because of this, I would suggest putting the node check before the npm check.

rtertiaer: swapped~

scripts/deploy
rtertiaer: npm requires node, and we require npm; this effectively checks for both.

Lohrer (approved):

PR #595, opened 2024-01-02
deploy /boot/config.txt with configure.py

### LINKED ISSUE #569: Manage /boot/config.txt in configure.py
[An issue came up in Discourse](https://amplipi.discourse.group/t/issues-after-upgrading-to-0-3-0/268/6?u=jimi). After discussing this a bit, we should be managing `/boot/config.txt` in `scripts/configure.py`.

rtertiaer:
This PR deploys /boot/config.txt on every firmware upgrade, with `configure.py`. Some older boxes had a `/boot/config.txt` that [creates problems on newer firmwares](https://amplipi.discourse.group/t/issues-after-upgrading-to-0-3-0/268/6?u=jimi). This will resolve that, at the cost of potentially overwriting an advanced user's custom `config.txt`. If we'd like to be more sophisticated about which versions we replace, I have some ideas - one is creating a list of checksums of prior config.txt revisions that can be replaced, and only replacing those.

Additionally, the version of `config/boot_config.txt` is _not_ what is currently baked into images; the diff is small:
```bash
$ diff config.txt config/boot_config.txt 
74a75,77
> 
> # Lower GPU memory from the default of 76 to 32 MB. 16 is the min but disables some features.
> gpu_mem=32
```
is there context for this change? It might be nice to get an extra 40mb of ram :)

I'm not quite sure if this should be documented elsewhere, besides the changelog.

This resolves #569 .
 
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   * Close the preset modal when the preset has been executed
 * Make zeroconf advertisement more robust to ip address changes
 * Remove deprecated old zeroconf advertisement
+* System
+  * Manage the /boot/config.txt Raspberry Pi firmware configuration file for bugfixes
 
 ## 0.3.1
 * System
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ include = [
   "config/*.conf",
   "config/bluetooth/*",
   "config/*.rules",
+  "config/boot_config.txt",
   "debs/**/*",
   "fw/*.sh", # no need to add full source tree
   "fw/bin/*",
--- a/scripts/configure.py
+++ b/scripts/configure.py
@@ -370,6 +370,10 @@ def _install_os_deps(env, progress, deps=_os_deps.keys()) -> List[Task]:
     _from = f"{env['base_dir']}/config/asound.conf"
     _to = "/etc/asound.conf"
     tasks += print_progress([Task(f"copy {_from} to {_to}", f"sudo cp {_from} {_to}".split()).run()])
+    # copy boot_config.txt RPi firmware configuration file
+    _boot_config_from = f"{env['base_dir']}/config/boot_config.txt"
+    _boot_config_to = "/boot/config.txt"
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Upcoming release
 * Web App
   * Fixed bug that allows disabled streams to be shown & selected
+* System
+  * Manage the /boot/config.txt firmware configuration file for bugfixes
```
Lohrer, line 7: I would avoid calling the software firmware, elsewhere in this repo the STM32 firmware is the only thing referred to as 'firmware'.

Lohrer: Ok after another half second of thought that totally should have happened before the previous reply I'm guessing Raspberry Pi is what's calling this 'firmware configuration'. In that case, I'd just change this to 'Manage the Raspberry Pi firmware configuration file /boot/config.txt for bugfixes'.

rtertiaer: yeah, that definitely could be wayyy clearer; i wanna call it what they call it, but need to differentiate it. I'll modify this a bit

Lohrer:
> is there context for this change? It might be nice to get an extra 40mb of ram :)
The context is it would be nice to get an extra 40 MiB of RAM. For our use case the GPU isn't used much so why give it extra resources that could be used for more audio streams?

linknum23 (approved):

Lohrer (changes requested):
I just changed the /boot/config.txt on my Pi, did a deploy with the changes here, then checked the diff of /boot/config.txt and config/boot_config.txt. My changes to /boot/config.txt were not overwritten, so I take it boot_config.txt isn't actually being copied for some reason.

linknum23:
Did you happen to look at the command line output?

linknum23:
That should be all that is needed. I'll verify mine gets overwritten.

linknum23:
`gpu_mem=32`
I think we are GTG!

Lohrer (approved):

rtertiaer:
good catch ya'll, thx

PR #603, opened 2024-01-04
Deploy fixes

Lohrer:

* Use poetry (pyproject.toml) for the base version number. Previously `git describe` was used to get the latest version tag. Since versions are not tagged in branches, `git describe` will never return anything higher than 0.3.0 going forward.
* Remove broken --mock-ctrl flag (I assume we don't need this anymore, if I'm wrong I can fix it instead of removing it)
* Added a `--no-deps` flag that removes the `--os-deps` and `--python-deps` flags for the `configure.py` call. This is useful for repeated deploys during development, since even with everything installed just checking for dependencies takes several minutes. A further optimization could be to avoid the reboot but I'm not sure of the consequences of that so I'll leave it for future work if someone desires that option.
* Robustified the success check. Basically the API check is now the same as the ping check.
 
### Checklist

  * In this case the `--help` text.

### DIFF
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -202,22 +202,29 @@ for i in {1..30}; do
   sleep 2
   printf "."
   if [[ "$ping_loc" == /c/Windows/system32/ping ]] && ping -n 2 $host; then
-    restart_finished=true;
-    break;
+    restart_finished=true
+    break
   elif ping -c 2 $host 1>/dev/null 2>/dev/null; then
-    restart_finished=true;
-    break;
+    restart_finished=true
+    break
   fi
 done
 echo
 if $restart_finished; then
-  printf "Restart successful, waiting for amplipi service";
-  for i in {1..20}; do
+  printf "Restart successful, waiting for amplipi service."
+  amplipi_found=false
+  for i in {1..30}; do
+    sleep 2
     printf "."
-    sleep 1
+    if curl -sX GET "${host}/api" -H "Accept: application/json" >/dev/null; then
+      amplipi_found=true
+      printf "\nUpdate finished.\n"
+      break
+    fi
   done
-  echo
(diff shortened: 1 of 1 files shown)


scripts/deploy
```diff
@@ -129,13 +126,8 @@ $python -m pip install --upgrade pip
...
 echo -e "Finished checking dependencies\n"
 
-if ! $mock_ctrl; then
-  # set ENABLE_HW flag since this is being deployed to a machine with the actual hardware setup
-  sed -i 's/DISABLE_HW = True/DISABLE_HW = False/' ../amplipi/rt.py
-fi
-
 # use a new version for the build but hide it from git since it isn't a real release
-old_version=$(poetry version -s)
+base_version=$(poetry version -s)
```
linknum23, line 130:
```suggestion
base_version=$(poetry version -s | sed 's/+.*//') # trim +XXXXX to avoid generating a bad version below
old_version=$(poetry version -s)
```

scripts/deploy
```diff
@@ -148,31 +140,27 @@ old_version=$(poetry version -s)
...
+npm install             # Install nodejs dependencies
+npm run build           # Build the web app
 popd
 
 # build release file (put in dist/)
 poetry build
 
 # revert to the old version so we don't modify extra files (the modified version info is in the build already)
-poetry version $old_version
+poetry version $base_version
```
linknum23, line 163:
```suggestion
poetry version $old_version
```

scripts/deploy
linknum23: This is not really necessary, but handles the case where this script failed building before setting the version back

linknum23 (approved):
I tested a couple of different iterations of this, works great and the changes are nice and clean!

PR #618, opened 2024-01-09
make image_pi bail if flashing failed

rtertiaer:
This PR makes `scripts/image_pi` bail if `dd` exits with non-zero status. @mrahman313 encountered this, where a transient network error caused multiple appliances to not flash correctly, but the script still reported at the end of the run "Flashing finished" as if it were successful. This interrupts that.

I'm explicitly opting not to add a ton of debug material here besides what dd outputs; this is a quick bugfix for production to continue.
 
### Checklist

### DIFF
--- a/scripts/image_pi
+++ b/scripts/image_pi
@@ -28,6 +28,17 @@ RED='\033[0;31m'
 GRN='\033[0;32m'
 NC='\033[0m'
 
+# Show user that an error occured before exiting
+handle_error () {
+  echo -e "\n${RED}An error occured! Imaging failed.${NC}\n"
+  read -p "Press any key to continue..." -rsn1
+  echo ""
+}
+
+# Bail if something goes wrong.
+trap handle_error ERR
+set -e
+
 img_file=''         # ...or use this image file explicitly
 instructions=true   # Skip manual instruction step
 while [[ "$#" -gt 0 ]]; do
@@ -86,8 +97,6 @@ est_time=$(echo "$img_size_mb/$write_speed_mb/60+0.5" | bc -l | cut -d'.' -f1)
 echo "Copying the image to the AmpliPi. This takes about $est_time minutes."
 echo "Go get a coffee or something :)"
 echo "Started at: $(date)"
-trap "read -p 'error! press any key to continue...'" ERR
-set -e # bail if something goes wrong.
 sudo dd if=$img_file of=$diskpath bs=4MiB status=progress
 echo "Finished at: $(date)"
 
--- a/scripts/image_pi
+++ b/scripts/image_pi
@@ -35,10 +35,6 @@ handle_error () {
   echo ""
 }
 
-# Bail if something goes wrong.
-trap handle_error ERR
-set -e
-
 img_file=''         # ...or use this image file explicitly
 instructions=true   # Skip manual instruction step
 while [[ "$#" -gt 0 ]]; do
@@ -97,7 +93,7 @@ est_time=$(echo "$img_size_mb/$write_speed_mb/60+0.5" | bc -l | cut -d'.' -f1)
 echo "Copying the image to the AmpliPi. This takes about $est_time minutes."
 echo "Go get a coffee or something :)"
 echo "Started at: $(date)"
-sudo dd if=$img_file of=$diskpath oflag=sync bs=1M status=progress
+sudo dd if=$img_file of=$diskpath oflag=sync bs=1M status=progress || handle_error
 echo "Finished at: $(date)"
 
 echo -e "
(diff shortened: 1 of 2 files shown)


Lohrer:
I just tested this by running `./scripts/image_pi ~/mn/images/amplipi_0.3.1.img`, then once it got to the actual copying part I yanked the USB cable. It printed out "Imaging successful!"...

rtertiaer:
I am extremely loathe to manage local disk images on the flashers themselves. Distributing these across multiple flashers is a huge pain. 

This is explicitly not intended to catch the case you mentioned; it is intended to catch the case like this, which it _does_ catch:
```
trap "read -p 'error! press any key to continue...'" ERR
set -e # bail if something goes wrong.
dd if=NOSUCHFILE of=some_destination_literally_doesnt_matter bs=4MiB status=progress
```

Lohrer:
I can't reproduce any case where `dd` explicitly fails, it just waits forever for the file to be accessible again then resumes once the network connection is back. I'm running this on my more up-to-date system so maybe this is different on the tester? In any case, I think printing a message on error is a great idea and I actually moved the `set -e` and `trap` to the top of the script so that any error will be trapped. Let me know what you think but I did test it a couple times to make sure there's no spurious errors elsewhere that have gone ignored.

Also, I added `oflag=sync` so that the progress report is actually correct. I figured I could sneak this in here and it's a minor change that I believe can only do good and I've been using for a while myself. Oh, and this would have been `oflag=dsync`, but you said that doesn't work on your system right?

rtertiaer:
> I actually moved the set -e and trap to the top of the script so that any error will be trapped.

This hasn't usually been wise in most non-trivial bash scripts I've maintained; I'd personally revert this change. Any non-zero exit status is "an error", but you'd be surprised how often CLI applications return a non-zero exit status to give some sort of indication that a program completed in a funny way. It may work on your system now, but very possibly won't for the general case, in the future, on a BSD, etc. This is why I put it right before the command I _actually_ want to bail if it fails.

> it just waits forever for the file to be accessible again
This happened when the network mount simply wasn't mounted. In this case, the script functions as intended. It functionally behaves like this snippet:
```
trap "read -p 'error! press any key to continue...'" ERR
set -e # bail if something goes wrong.
dd if=NOSUCHFILE of=some_destination_literally_doesnt_matter bs=4MiB status=progress
```

> Also, I added oflag=sync so that the progress report is actually correct. I figured I could sneak this in here and it's a minor change that I believe can only do good and I've been using for a while myself. Oh, and this would have been oflag=dsync, but you said that doesn't work on your system right?

This change is incompatible with OS X; perhaps that's what you're remembering? Either `dsync` or `sync` work just fine on my Linux system. I'll leave this up to you whether to rip out or not, considering we now have nice Balena Etcher docs for end-users and `image_pi` is probably more explicitly for our usage. (fwiw, it's already incompatible because of both `bs=4MiB status=progress`.)

Lohrer:
> This is why I put it right before the command I actually want to bail if it fails.
If we only care if one command failed, why not just check it's exit code? Without a `set +e`, anything after the `dd` that throws an error will also cause an early exit.

> This happened when the network mount simply wasn't mounted.
I tried that, and the file existence check earlier in the script caused the whole thing to bail before getting to `dd`. Do I have to yank the network between `if [ -z $img_file ]` and `dd`? I guess there could be significant time there since it's waiting for a user to press enter. I hadn't thought of that...

Regarding the `oflag=sync`, if we're treating this as internal use only then I won't care about compatibility outside Debian-based systems. But I want to add it because I assume it will be helpful to production to be able to see how far along the flashing is. Without `oflag=sync` we might as well remove `status=progress` because it becomes meaningless.

scripts/image_pi
```diff
@@ -86,7 +97,7 @@ est_time=$(echo "$img_size_mb/$write_speed_mb/60+0.5" | bc -l | cut -d'.' -f1)
 echo "Copying the image to the AmpliPi. This takes about $est_time minutes."
 echo "Go get a coffee or something :)"
 echo "Started at: $(date)"
-sudo dd if=$img_file of=$diskpath bs=4MiB status=progress
+sudo dd if=$img_file of=$diskpath oflag=sync bs=1M status=progress
```
Lohrer, line 100: Ok, I finally was able to reproduce the problem that this PR fixes with the "remove access to the file while waiting for bootup" as mentioned above. Since avoiding the global `set -e` was suggested, here's my suggestion (with removing the `trap handle_error ERR` and `set -e` lines above.
```suggestion
sudo dd if=$img_file of=$diskpath oflag=sync bs=1M status=progress || handle_error
```

Lohrer (changes requested):

Lohrer (approved):

PR #624, opened 2024-01-18
make upload button disabled until config is selected

klay2000:
Makes upload button in configuration settings disabled until a configuration is selected.

### DIFF
--- a/web/src/pages/Settings/Config/Config.jsx
+++ b/web/src/pages/Settings/Config/Config.jsx
@@ -53,11 +53,17 @@ const LMSModeHandler = () => {
 
 const Config = ({ onClose }) => {
     const [file, setFile] = React.useState([]);
+    const [filePicked, setFilePicked] = React.useState(false);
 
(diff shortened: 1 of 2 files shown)


web/src/pages/Settings/Config/Config.jsx
```diff
@@ -53,11 +53,17 @@ const LMSModeHandler = () => {
 
 const Config = ({ onClose }) => {
     const [file, setFile] = React.useState([]);
+    const [file_picked, setFilePicked] = React.useState(false);
```
rtertiaer, line 56: ngl, the mixing of camelCase and snake_case gives me the ick, but that's definitely just a nitpick.

rtertiaer (approved):

web/src/pages/Settings/Config/Config.jsx
klay2000: woops, just saw that, should be fixed now

rtertiaer (approved):

Issue #621, opened 2024-01-12
Re-add web/static to deploy


Lohrer:
In commit 725ccf0e0da7db960f8a2a23a60ec869d5eec427 `web/static` was removed from deploys/releases. This appears to be a necessary directory, doing a deploy to a clean Pi gives the log:

```txt
Jan 12 08:14:47 amplipi197 systemd[610]: Started Amplipi Home Audio System.
Jan 12 08:14:50 amplipi197 authbind[14531]: Traceback (most recent call last):
Jan 12 08:14:50 amplipi197 authbind[14531]:   File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
Jan 12 08:14:50 amplipi197 authbind[14531]:     "__main__", mod_spec)
Jan 12 08:14:50 amplipi197 authbind[14531]:   File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
Jan 12 08:14:50 amplipi197 authbind[14531]:     exec(code, run_globals)
Jan 12 08:14:50 amplipi197 authbind[14531]:   File "/home/pi/amplipi-dev/venv/lib/python3.7/site-packages/uvicorn/__main__.py", line 4, in <module>
Jan 12 08:14:50 amplipi197 authbind[14531]:     uvicorn.main()
Jan 12 08:14:50 amplipi197 authbind[14531]:   File "/home/pi/amplipi-dev/venv/lib/python3.7/site-packages/click/core.py", line 1157, in __call__
Jan 12 08:14:50 amplipi197 authbind[14531]:     return self.main(*args, **kwargs)
Jan 12 08:14:50 amplipi197 authbind[14531]:   File "/home/pi/amplipi-dev/venv/lib/python3.7/site-packages/click/core.py", line 1078, in main
Jan 12 08:14:50 amplipi197 authbind[14531]:     rv = self.invoke(ctx)
Jan 12 08:14:50 amplipi197 authbind[14531]:   File "/home/pi/amplipi-dev/venv/lib/python3.7/site-packages/click/core.py", line 1434, in invoke
Jan 12 08:14:50 amplipi197 authbind[14531]:     return ctx.invoke(self.callback, **ctx.params)
Jan 12 08:14:50 amplipi197 authbind[14531]:   File "/home/pi/amplipi-dev/venv/lib/python3.7/site-packages/click/core.py", line 783, in invoke
[...]

PR #633, opened 2024-01-30
Fix static asset handling

rtertiaer:

This PR fixes static asset handling during deployment, as noted in #621. Static assets should be handled via rollup to prevent browser caching of old assets; there should be zero assets served from `STATIC_DIR`, only `WEB_DIR`. 

Along the way, I upgraded `@vitejs/plugin-react` to suppress the massive quantities of Rollup warnings during build from `use client`. There were literally 10000+ lines of these warnings:
```
node_modules/@mui/icons-material/esm/index.js (1:0) Module level directives cause errors when bundled, "use client" in "node_modules/@mui/icons-material/esm/index.js" was ignored.
```
 
This PR is a draft until I can test a deploy on hardware. At present, my debug webserver+vite hotloaded copy functioned correctly.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # AmpliPi Software Releases
 
+## Upcoming version
+* Web App
+  * Improve static asset deployment
+
 ## 0.3.3
 * Web App
   * Improve functionality of configuration page
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,8 @@ include = [
   "streams/**/*",
   "bin/**",
   "docs/**",
-  "web/dist"
+  "web/dist",
+  "web/static"
 ]
 exclude = [
   "scripts/deploy", "scripts/bootstrap_pi" # only support development
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -27,7 +27,7 @@
       "devDependencies": {
         "@types/react": "^18.0.28",
         "@types/react-dom": "^18.0.11",
-        "@vitejs/plugin-react": "^3.1.0",
+        "@vitejs/plugin-react": "^4.2.1",
         "eslint": "^8.47.0",
         "eslint-plugin-react": "^7.33.1",
         "https": "^1.0.0",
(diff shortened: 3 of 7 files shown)


rtertiaer:
Further complicating things is that the static dir is theoretically needed to give a rendition of the stream playing, as per `amplipi/app.py`'s `get_image()` (though it's unclear if that is live anywhere?)

It's simple enough to add `web/static` back in to `pyproject.toml`, so I'll do so.

linknum23:
Tread carefully here, some tests and other things depend on web/static:
```bash 
link@link:~/mn/amplipi (main)$ ag web/static/
amplipi/tests.py
253:      ap2.announce(models.Announcement(source_id=3, media='web/static/audio/aux_in.mp3'))
276:  digital_msgs = [models.Announcement(source_id=src, media=f'web/static/audio/digital{src+1}.mp3', vol=-30)
278:  analog_msgs = [models.Announcement(source_id=src, media=f'web/static/audio/analog{src+1}.mp3') for src in range(4)]
305:  digital_msgs = [models.Announcement(source_id=src, media=f'web/static/audio/digital{src+1}.mp3', vol=-30)

docs/conf.py
64:html_favicon = "../web/static/imgs/amplipi_icon_white.png"
65:html_logo = "../web/static/imgs/amplipi_icon_white.png"

docs/index.rst
6:.. image:: ../web/static/imgs/amplipi_banner_white.png

index.html
22:    <script type="module" src="web/static/js/libs/rapidoc/9.3.3/rapidoc-min.js"></script>
26:      <img slot="nav-logo" src="web/static/imgs/amplipi_icon_black.png"
```

rtertiaer:
oof. thanks for pointing those out. I guess I'll just add the `/static` endpoint back in :woozy_face:

linknum23:
We can host those files somewhere else if it makes more sense.

rtertiaer:
I've tested this latest rendition on an actual box & she works a charm.

For future reference, I think the only things mentioned above in the "tread carefully here" that actually block removing the `/static` endpoint are the references in `index.html`, for the OpenAPI spec hjosting.

linknum23 (approved):
Verified the static/ dependent tests still work with a from scratch install. The rest of this seems reasonable.

PR #634, opened 2024-01-31
Changes to Webapp for Mobile App

klay2000:
PR for changes allowing saving IP address to mobile app, also changed some settings to assist with JS formatting. To test this you'll need to be on the internal testing track for the android app and have an android device so come grab me when you get around to it and we'll set that up.


klay2000:
Sorry for the accidental formatting changes!

rtertiaer (approved):

Issue #610, opened 2024-01-08
Make stream infrastructure plug-able / stream plugins


rtertiaer:
Making each stream a plugin gives us a bunch of benefits:
* stream development should be less tied to amplipi development; in other words, it should hopefully be easier to accept contributions from external folks
* stream stability should not affect amplipi stability (though we've done a good job of this thus far.)
* handling streams as separate processes, not necessarily managed by the frontend, should allow us to use a dedicated process manager, like `supervisord` or `systemd` template units; this should also improve stream stability and give us security benefits as well.

The work done for long-running background streams really set the table for this work (see #262 ); we just need to define the interface better, split out our existing streams into plugins, implement a robust process manager, etc.

PR #646, opened 2024-02-27
Hide fm radio

### LINKED ISSUE #631: Hide FM Radio option when no hardware is plugged in
One can configure an FM Radio stream even if there is no hardware plugged in. This is a bit confusing, because it implies the AmpliPis have built-in FM radio receivers. Remove this option if there is no receiver plugged in.

rtertiaer:
This PR intends to hide the FM radio stream addition option when hardware is not present. This closes #631 .

It attempts to do this in a way that's relatively scalable and set us up for the future of #610 . However, `mypy` was quite displeased with my use of class variables and I couldn't make it less displeased. After spending quite some time on investigating why `__dir__()` won't show classvars, I've opted to punt on this and just wait til #610 lands (or solicit quick feedback here on being more rigorous.)
 
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * System
   * Add ability to display a special message for shipping on the eink display
   * Make eink display clear when the display process is stopped
+* Streams
+  * Hide FM Radio if hardware is not available
 
 ## 0.3.3
 * Web App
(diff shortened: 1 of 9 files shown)


rtertiaer:
Alright, this is ready for review once more. I've opted to ignore the super persnickety line's typing.

linknum23 (approved):
This seems to work well. I tested with adding and removing the rtl-fm dongle and everything behaved as expected. I also tested out the extra api info and it looked good.

rtertiaer:
There was a lot of code whiplash and unnecessary commits; I've squashed all 13 commits into one.

PR #648, opened 2024-02-28
Stop audio from echoing on certain internet radio streams

### LINKED ISSUE #625: Sources doubling up audio
Related to [this post](https://amplipi.discourse.group/t/streams-doubling-up-when-first-starting/283) in our discourse group

We've been hearing reports of various stream types "doubling up" (seemingly starting twice when loaded) since the 0.3.0 release
If anyone else has this issue, please collect logs via Settings->About->Logs and send them to us at [support@micro-nova.com](mailto:support@micro-nova.com) with the subject "RE: stream echo"

As for the actual issue documentation, streams are loading twice. This used to be an issue with just internet radio (#400) but it applies to multiple now, with reports of LMS and I personally have heard it happen with pandora

This _feels_ like it's related to frontend react state management, but I can't promise that since I haven't poked around yet

rtertiaer:
This PR fixes one bug where `restart_vlc()` is called unnecessarily. This fixes #625 in the case we've observed in-house. This often happened when opening a stream took a long time; this PR both extends the initial opening time from 2s to 10s, and also improves responsiveness for streams that open well before that 2s.

One case that seems odd is that linting should have caught at least one potential reason why `restart_vlc()` was called, triggering the echo. Even though it did not catch it, I've opted to lint for errors on these scripts as well.
 
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   * Hide FM Radio if hardware is not available
   * Make pandora stream album art use HTTPS urls to make sure it is rendered in the ios app
   * Make Pandora streams a bit more robust against failure
+  * Fix internet radio startup bug that caused echo
 
 ## 0.3.3
 * Web App
--- a/streams/runvlc.py
+++ b/streams/runvlc.py
@@ -79,6 +79,7 @@ def update_info(info) -> bool:
     log('Error: %s' % sys.exc_info()[1])
     return False
 
+log("Starting VLC instance.")
 instance = vlc.Instance(config.split())
 try:
   media = instance.media_new(args.url)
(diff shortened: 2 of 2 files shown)


streams/runvlc.py
```diff
@@ -137,13 +138,41 @@ def restart_vlc():
...
+    break
+  elif state in ['State.Opening', 'State.Buffering'] and opening_counter <= MAX_STREAM_OPENING_COUNTER:
+    # give it some time. Some streams take a while.
+    time.sleep(0.25)
+    opening_counter += 1
+    log(f"Stream still opening; waiting {MAX_STREAM_OPENING_TIME - STREAM_OPENING_BACKOFF*opening_counter} more seconds...")
+  elif opening_counter > MAX_STREAM_OPENING_COUNTER:
+    log("Stream took too long to open.")
+    # This is a hail mary; either we restart_vlc & try again, or exit. If we lived under a real process monitor
+    # like VLC, I'd probably exit(1) and let it handle graceful backoffs and restarts... but since we
```
linknum23, line 158: like systemd?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Streams
   * Hide FM Radio if hardware is not available
   * Make pandora stream album art use HTTPS urls to make sure it is rendered in the ios app
+  * Fix a bug where audio would echo
```
linknum23, line 14:
```suggestion
  * Fix internet radio startup bug that caused echo
```

linknum23 (approved):

PR #650, opened 2024-02-28
Ensure required fields are set before sending form

### LINKED ISSUE #572: Disallow saving Pandora stream configs without all fields completed
The UI will gleefully send off your request with no "name" field; the API rightfully rejects it. we can both handle the form better before it hits the API, and also handle the 422 Unprocessable Request response from the API.

rtertiaer:
This PR checks that every required field for a stream has a value before sending the update to the API. It doesn't do much else in the way of form validation; there's also a case to be made here for actually reading the status of the POST/PATCH for a particular stream, but that's probably a different issue entirely.

This PR fixes #572 .
 
### Checklist


klay2000 (changes requested):
This works but I think the text should be red when an error occurs or an [alert component](https://mui.com/material-ui/react-alert/) should pop up.

rtertiaer:
Screenshot from 2024-03-04 14-59-25

:sunglasses:

klay2000 (approved):

Issue #570, opened 2023-12-18
Put Pandora in a process monitor


rtertiaer:
For stability reasons, Pandora ought to run in a process monitor.

PR #657, opened 2024-03-05
Ensure Pandora streams deactivate under process_monitor.py

### LINKED ISSUE #656: Pandora deactivation leaves child process running
Some part of #570 broke this. Repro steps on 0.3.4:
* Create and start a pandora stream
* Stop said stream
* You will see `pianobar` inherited by the `systemd --user` process; only `alsaloop` is being managed by `process_monitor.py`.

rtertiaer:
This PR ensures all `pianobar` processes running under a `process_manager.py` get cleaned up when asked. This fixes #656 , which was introduced with #570.

I intend to add this as a hotfix to the 0.3.4 release, as it's still marked as a "prerelease" and this bug was found during QA of that release.
 
### Checklist

### DIFF
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -738,8 +738,18 @@ class Pandora(PersistentStream):
 
   def _deactivate(self):
     if self._is_running():
-      self.proc.kill()
-      self.proc.wait()
+      try:
+        self.proc.terminate()
+        self.proc.wait(timeout=4)
+      except:
+        # Likely a subprocess.TimeoutException, but we will handle all exceptions the same.
+        # Because this runs within a process_monitor.py, we need to handle the entire
+        # process group.
+        os.killpg(self.proc.pid, signal.SIGKILL)
+        # does the child still leave a zombie if we wait() the parent? pid space is not at
+        # a premium, so I'll leave that question unanswered.
+        self.proc.wait()
+
     self.proc = None
     self.ctrl = ''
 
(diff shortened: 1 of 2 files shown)


amplipi/streams.py
```diff
@@ -738,8 +738,18 @@ def _activate(self, vsrc: int):
...
+        self.proc.terminate()
+        self.proc.wait(timeout=4)
+      except:
+        # Likely a subprocess.TimeoutException, but we will handle all exceptions the same.
+        # Because this runs within a process_monitor.py, we need to handle the entire
+        # process group.
+        os.killpg(self.proc.pid, signal.SIGKILL)
+        # does the child still leave a zombie if we wait() the parent? pid space is not at
+        # a premium, so I'll leave that question unanswered.
+        self.proc.wait()
```
linknum23, line 751: Can this proc.wait block forever?

rtertiaer: I think it can... if `SIGKILL` fails to terminate a process. If my memory serves me, this can happen when blocking IO is occurring. FWIW I don't think we should engineer for that case (at least at present), because that's effectively very broken anyhow - chances are that blocking IO is communication is with our ALSA loopback.

We do this in other places in the code, using `communicate()` (`wait()` is communicate without the inputs & outputs); this is additionally how [the `subprocess` docs recommend accomplishing this](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate).

linknum23 (approved):

PR #661, opened 2024-03-07
Python-Logging

SteveMicroNova:
Add python's logging module and use it to replace random print() statements around everywhere
 
### Checklist

### DIFF
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -81,17 +81,8 @@ app.mount("/generated", StaticFiles(directory=GENERATED_DIR), name="generated")
 
 app.add_exception_handler(NotAuthenticatedException, not_authenticated_exception_handler)
 
-logger = logging.getLogger()
-logger.setLevel(logging.INFO)
-
-loghandler = logging.StreamHandler()
-loghandler.setLevel(logging.INFO)
-
-logformat = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
-
-loghandler.setFormatter(logformat)
-
-logger.addHandler(loghandler)
+logger = logging.getLogger('ampli-logger')
+logger.setLevel(logging.DEBUG)
 
 class SimplifyingRouter(APIRouter):
   """
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -81,7 +81,17 @@ app.mount("/generated", StaticFiles(directory=GENERATED_DIR), name="generated")
 
 app.add_exception_handler(NotAuthenticatedException, not_authenticated_exception_handler)
 
-logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+loghandler = logging.StreamHandler()
+loghandler.setLevel(logging.INFO)
+
+logformat = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
+
+loghandler.setFormatter(logformat)
+
+logger.addHandler(loghandler)
 
 class SimplifyingRouter(APIRouter):
   """
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -24,6 +24,8 @@ The FastAPI/Starlette web framework is used to simplify the web plumbing.
 
 import argparse
 
+import logging
+
 import os
 
 # type handling, fastapi leverages type checking for performance and easy docs
@@ -230,7 +232,7 @@ def lms_mode(ctrl: Api = Depends(get_ctrl)):
   """ Toggles Logitech Media Server mode on or off. """
   new_config: models.Status
   if ctrl.lms_mode:
-    print("turning LMS mode off...")
+    logging.info("turning LMS mode off...")
     try:
       os.remove(pathlib.Path(defaults.USER_CONFIG_DIR, "lms_mode"))
     except FileNotFoundError:
@@ -239,7 +241,7 @@ def lms_mode(ctrl: Api = Depends(get_ctrl)):
     Popen('sudo systemctl disable logitechmediaserver', shell=True)
     new_config = models.Status(**defaults.default_config(is_streamer=ctrl.is_streamer, lms_mode=False))
   else:
-    print("turning LMS mode on...")
+    logging.info("turning LMS mode on...")
     pathlib.Path(defaults.USER_CONFIG_DIR, "lms_mode").touch()
     Popen('sudo systemctl start logitechmediaserver', shell=True)
     Popen('sudo systemctl enable logitechmediaserver', shell=True)
@@ -268,15 +270,15 @@ async def subscribe(req: Request):
     try:
       while True:
         if await req.is_disconnected():
-          print('disconnected')
+          logging.info('disconnected')
           break
         if not msg_que.empty():
           msg = msg_que.get()
           yield msg
         await asyncio.sleep(0.2)
-      print(f"Disconnected from client {req.client}")
+      logging.info(f"Disconnected from client {req.client}")
     except asyncio.CancelledError as exc:
-      print(f"Disconnected from client (via refresh/close) {req.client}")
+      logging.info(f"Disconnected from client (via refresh/close) {req.client}")
       # Do any other cleanup, if any
       raise exc
   return EventSourceResponse(stream())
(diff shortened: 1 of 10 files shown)


SteveMicroNova:
I literally just searched for "print" in all files
Once I started getting outside of the main files and into seemingly ultra-specific firmware stuff (for example, generate_therm_tables.py in fw/preamp) or the tests I decided to stop replacing stuff

When reviewing this, make sure to verify that the logging level is appropriate (info, debug, warning, critical, exception, error)

amplipi/hw.py
```diff
@@ -393,14 +394,14 @@ def program_all(self, filepath: str, num_units: Optional[int] = None, baud: int
...
-    print(f'Programming up to {program_count} AmpliPi Preamps '
+    logging.info(f'Programming up to {program_count} AmpliPi Preamps '
           f'({previous_unit_count} currently detected)')
 
     # Attempt programming until program_count, but stop if an error occurs
     unit = 0
     success = True
     while success and unit < program_count:
-      print()
+      print() # PR COMMENT: Why the empty print?
```
rtertiaer, line 404: probably to get a newline.

the comment probably doesn't belong in the codebase, and instead as a comment in the PR like this one.

amplipi/mpris.py
```diff
@@ -143,22 +144,22 @@ def close(self):
 
     self.metadata_process.terminate()
     if self.metadata_process.wait(1) != 0:
-      print('Failed to stop MPRIS metadata process, killing', flush=True)
+      logging.info('Failed to stop MPRIS metadata process, killing', flush=True)
```
rtertiaer, line 147: I can't find a reference to a `flush=true` in anything logger related. Can you find the definition for me?

amplipi/streams.py
```diff
@@ -1335,16 +1336,16 @@ def _activate(self, vsrc: int):
...
   def _deactivate(self):
     if self._is_running():
       try:
         self.proc.terminate()
         self.proc.communicate()
       except Exception as e:
-        print(f"failed to terminate LMS stream {self.name}: {e}")
-        print("forcefully killing.")
+        logging.exception(f"failed to terminate LMS stream {self.name}: {e}")
+        logging.info("forcefully killing.")
```
rtertiaer, line 1348: these deserve to be the same log level. imagine setting log level to "INFO"; you'd only get an obtuse line saying "forcefully killing", without it specifying what it's killing or why.

amplipi/utils.py
```diff
@@ -377,19 +378,19 @@ def get_identity() -> dict:
       potential_identity = json.load(identity_file)
       identity.update(potential_identity)
   except FileNotFoundError as e:
-    print(f'Error loading identity file: {e}')
-    print('Creating an identity file from defaults.')
+    logging.exception(f'Error loading identity file: {e}')
+    logging.info('Creating an identity file from defaults.')
```
rtertiaer, line 382: good example of where using two different log levels is helpful; the second message informs a user what is going on, with reasonable context.

rtertiaer (changes requested):
The second part of #661 was:
> Set and manage the log level some place central, during app initialization.

you still need to add this bit

amplipi/hw.py
SteveMicroNova: I found that early on in the process and didn't expect to remember to actually comment on it
I assume the codebase thing is fine if I squash before merging when I remove it in a second

rtertiaer: definitely :sunglasses:

amplipi/streams.py
SteveMicroNova: I made it a single logging instance in the recent commit
`logging.exception(f"failed to terminate LMS stream {self.name}: {e} \nforcefully killing.)`
This seems like a good practice to me to retain the newline but without the risk of other prints / logging events happening between the two instances, \n for newline seems uncommon so I wanted to call attention to it in case that's intentionally avoided elsewhere

amplipi/mpris.py
SteveMicroNova: There's no such thing, I just missed that when converting the print to a log

amplipi/streams.py
rtertiaer: It'll look a little funny in the logs; normally we don't want to be concerned with formatting/rendering log lines and just emitting them, but it'll do.

amplipi/mpris.py
```diff
@@ -143,22 +144,22 @@ def close(self):
...
     if self.metadata_process.wait(1) != 0:
-      print('Failed to stop MPRIS metadata process, killing', flush=True)
+      logging.info('Failed to stop MPRIS metadata process, killing')
       self.metadata_process.kill()
 
     self.metadata_process = None
 
     if self.mpris:
-      print('disconnecting proxy', flush=True)
+      logging.info('disconnecting proxy', flush=True)
```
rtertiaer, line 153: there are a handful of these `flush`es.

amplipi/app.py
```diff
@@ -79,6 +81,7 @@
 
 app.add_exception_handler(NotAuthenticatedException, not_authenticated_exception_handler)
 
+logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
```
rtertiaer, line 84: could you give this logger a name, and import it in each of the files?

I was imagining something more along the lines of [this](https://docs.python.org/3/howto/logging.html#configuring-logging). This is actually what gives us the flexibility to move log lines around, increase logging on particular components as we need, direct certain things to files. You don't need to bother with any of the formatting or stream handler stuff - just creating a particular `logger` instance and getting it in each module is plenty.

rtertiaer (changes requested):

amplipi/app.py
```diff
@@ -81,7 +81,17 @@
 
 app.add_exception_handler(NotAuthenticatedException, not_authenticated_exception_handler)
 
-logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
+logger = logging.getLogger()
```
rtertiaer, line 84: can you give this logger a name?

amplipi/app.py
```diff
@@ -81,7 +81,17 @@
...
+logger.setLevel(logging.INFO)
```
rtertiaer, line 85: should probably match what we already use for now - DEBUG.

amplipi/app.py
```diff
@@ -81,7 +81,17 @@
...
+
+loghandler = logging.StreamHandler()
+loghandler.setLevel(logging.INFO)
```
rtertiaer, line 88: where do you use this?

amplipi/auth.py
```diff
@@ -2,7 +2,7 @@
 import json
 import time
 import secrets
-import logging
+from app import logger
```
rtertiaer, line 5: you probably actually wanna use `getLogger`: https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

rtertiaer (changes requested):

amplipi/app.py
SteveMicroNova: I've gotten the sense that DEBUG is meant to be used as needed and INFO is meant to be the default lowest level, but I'll totally follow your lead here

amplipi/app.py
SteveMicroNova: https://github.com/micro-nova/AmpliPi/pull/661#discussion_r1521765478
Nowhere
I admittedly read as far as "I was imagining something more along the lines of (link)" and just followed the link as a tutorial, forgot to come back to see you saying this was unnecessary for our needs

amplipi/app.py
```diff
@@ -79,6 +81,8 @@
 
 app.add_exception_handler(NotAuthenticatedException, not_authenticated_exception_handler)
 
+logger = logging.getLogger('ampli-logger')
+logger.setLevel(logging.DEBUG)
```
Lohrer, line 85: We shouldn't just hard-code DEBUG log level, this should ideally be a config option.

Lohrer:
If you want to dig deeper, the [display code](https://github.com/micro-nova/AmpliPi/blob/89860185bb6313ee0a8661b3b081889d6a4fa948/amplipi/display/display.py#L7) uses `loguru`, so searching for `print` won't catch those messages. They are all `log.info`/`log.debug` etc.

rtertiaer (approved):
:raised_hands:

PR #664, opened 2024-03-08
Lms metadata 3

SteveMicroNova:
Finally fix and implement the LMS metadata collection script I originally made just over a year ago
 
### Checklist


SteveMicroNova:
Currently it's incredibly efficient at finding the player when there isn't a specified server; python's socket module seems to be complaining whenever I try to resolve the hostname but also that's because the LMS server is on my windows machine and the metadata script runs in WSL and the hostname resolves to the WSL instance instead of windows; needs more testing outside of my super specific setup to verify that it's good with or without hostname being provided

Multiple servers/players on the same network seem to be a trouble right now as well, but that's a less common usecase and not the highest order concern while I figure out the single stream single server setup.

SteveMicroNova:
This feels ready for testing

To test, spin up multiple LMS streams with various settings on one AmpliPro after deploying this branch to it and play things to each instance and witness the metadata. If you need a unit to test on, lamplipi (the one on lincolns desk) currently has it deployed

I don't know how to / if I have to implement that into our pytests thing as the last checkbox of the PR template suggest

streams/lms_metadata.py
```diff
@@ -0,0 +1,180 @@
...
+    while not connected:
+      try:
+        if self.server is not None:
+          self.ip = socket.gethostbyname(self.server)
+        else:
+          find_lms_server = "bin/arm/find_lms_server"
+          if platform.machine() == "x86_64":
+            find_lms_server = "bin/x64/find_lms_server"
+          # Much faster method of connecting to the metadata server using code from: https://github.com/ralph-irving/squeezelite/blob/master/tools/find_server.c
+          ip_find = subprocess.run([find_lms_server], check=True, capture_output=True, text=True)
```
rtertiaer, line 90: i'd love to see this more effectively case-statemented out; I can see a world where we run both x86_64, arm, and arm64.

is there a fallback for when these executables cannot be run? it doesn't appear we actually handle the exception potentially generated from `check=true`; that feels important here.

rtertiaer (changes requested):
how did you generate the binaries for this?

SteveMicroNova:
> how did you generate the binaries for this?

They were compiled a year ago so I don't fully remember

I'm pretty sure I compiled the find_lms_server.c once on my local (x64) and once on an amplipi (arm)

SteveMicroNova:
I don't know how to compile C programs for other architectures so I've simply made it an error case if there's an unsupported architecture that tries to run the program
Ideally if we ever use a Pi with a different type of CPU, we'll just compile all of our binaries on it as well so we'll have the right ones available before sending it out into the world

streams/lms_metadata.py
SteveMicroNova: Now handled explicitly with `except subprocess.CalledProcessError` in recent commit

streams/lms_metadata.py
```diff
@@ -0,0 +1,209 @@
...
+            find_lms_server = "bin/arm/find_lms_server"
+          else:
+            self.meta.artist = "Unsupported CPU architecture for LMS MetaData"
+            self.meta.album = "Please contact AmpliPi Support:"
+            self.meta.track = "support@micro-nova.com"
+            logging.error("LMS metadata reader has detected an unsupported chipset")
+            abort = True
+
+          # Much faster method of connecting to the metadata server using code from: https://github.com/ralph-irving/squeezelite/blob/master/tools/find_server.c
+          ip_find = subprocess.run([find_lms_server], check=True, capture_output=True, text=True)
```
rtertiaer, line 113: what is the "less fast" way of connecting to the metadata server? is it an acceptable fallback for the unsupported CPU arch case? ie, is it pure Python?

SteveMicroNova: The less fast way is via manually scraping the whole network with python nmap; one that was prohibitively slow and also annoyed the hell out of every other device on the network by going door to door and asking "Are you an LMS server?"

Technically possible, doesn't seem worthwhile or sensible

IMO it'd be better to implement the ability to write down the IP as this script already supports that and is simply awaiting the LMS stream to add it in as an option during stream creation

rtertiaer: > IMO it'd be better to implement the ability to [input] the IP

after reading through the rest of this, I tend to agree. if you:
a) haven't provided a hostname or IP
b) the LMS server cannot be autodiscovered, with the binary provided
then you've already come to the end of the road. making (a) accept more inputs seems like a reasonable path here

SteveMicroNova:
I don't know why MyPy linting is complaining, the line it's currently complaining was just reverted to the way it's almost always been until like three commits ago

amplipi/streams.py
```diff
@@ -1331,7 +1341,9 @@ def _activate(self, vsrc: int):
           # NOTE: port 9000 is assumed
           server.replace('localhost', socket.gethostname())
         lms_args += ['-s', server]
-      # TODO: allow port to be specified with server (embedding it in the server URL does not work)
+
+      meta_args = ['python3', 'streams/lms_metadata.py', '--name', self.name, '--server', self.server]
```
rtertiaer, line 1345: it looks like the thing mypy is complaining about is the `Optional` bit; i think it may be because self.name and self.server are not guaranteed to be set here, which could indeed trigger a bug

SteveMicroNova:
I'd like to add port support during this PR, but I don't know how to do so
That requires changing the stream; I've found out how to add the spot in the form on the frontend but I don't know how to make that data actually be saved when it's input so far

amplipi/streams.py
```diff
@@ -1331,7 +1348,15 @@ def _activate(self, vsrc: int):
           # NOTE: port 9000 is assumed
```
SteveMicroNova, line 1348: This makes me feel like port should be used around here
I don't currently know where or how
I mean, it has to be appended like `lms_args += ['-s', server:port]`, but the use of "assumed" makes me feel that 9000 is input somewhere somehow automatically and I'll have to replace it first

SteveMicroNova:
This actually needs more work before merging
We have proper folders to store different metadata in, Lincoln told me about it verbally and I haven't implemented that yet because I got focused on making the branch actually work and forgot to make it work good
I'm currently on #667, I'll loop back to this in a bit

streams/lms_metadata.py
```diff
@@ -0,0 +1,217 @@
...
+  # parser.add_argument('--port', type=int, default=9000, help='The port the LMS server uses')
+  # parser.add_argument('--ref', type=int, default=2, help='The frequency of metadata refresh cycles')
+  # parser.add_argument('--debug', action='store_true', default=False, help='''debug mode, activates various console logs so that you can debug in the command line,
+  #                     also creates json dumps in the main directory: {player name}_track_raw.json and {player name}_song_raw.json''')
+  # args = parser.parse_args()
+
+  # There's a few args here that aren't currently used by the system, but can be helpful for debugging in the field
+  # some are also there so we can implement future features easier, or so the devs at home can do the same with a minimal lift
+  # LMSMetadataReader(args.name, args.server, args.port, args.ref, args.debug).connect()
+  LMSMetadataReader("LinkData", None, 9000, 2, True).connect()
```
SteveMicroNova, line 217: Left local test changes in on accident, these will be cleaned out in next commit

streams/lms_metadata.py
```diff
@@ -0,0 +1,218 @@
...
+          # faster relative to the original method, using NMAP to go door to door (ip to ip) and ask if self.name is home
+          ip_find = subprocess.run([find_lms_server], check=True, capture_output=True, text=True)
+          # Uses regex because ip_find spits out as '{Hostname}:{port} ({ip})'
+          self.server = re.search(r'\((.*?)\)', ip_find.stdout).group(1)
+
+        if self.debug:
+          logging.debug(f"Found LMS Server: {self.server}:{self.port}")
+
+        player_json = {"id": 1,	"method": "slim.request",	"params": [self.player_name, ["players", "-", 100, "playerid"]]}
+        player_info = requests.get(f'http://{self.server}:{self.port}/jsonrpc.js', json=player_json, timeout=5)
```
SteveMicroNova, line 132: image

For some reason if you don't populate self.server during __init__, this has a chance of never getting anything for that value and so polls for `http://:900/jsonrpc.js` which is a fail state

self.server should always be populated by this point though, if it wasn't solved by `resolve_host` then it should be solved by `find_lms_server`; if somehow neither of those happen it should error out before ever reaching this point in the code

SteveMicroNova:
The metadata script is incredibly clean and efficient, if noisy at the moment
I'm moving on to more fruitful endeavors for now as Ryan and I have identified an issue I've been having as inherent to our alsaloop implementation or the version of LMS we're running, either way something that is out of my power and knowledge and may need work done outside of this branch to solve

SteveMicroNova:
Issue #679 that was blocking this PR has been solved in #681, this branch now works and is safe to review and eventually merge :+1:

amplipi/models.py
```diff
@@ -83,6 +83,7 @@ class fields(SimpleNamespace):
   * Digital or Analog Stream ('stream=SID') where SID is the ID of the connected stream (rca inputs are now just the RCA stream type)
   * Nothing ('') behind the scenes this is muxed to a digital output
   """)
+  Port = Field(description='Port used by LMS server for metadata collection')
```
rtertiaer, line 86: Is it reasonable to set a default here?

amplipi/models.py
```diff
@@ -106,6 +107,7 @@ class fields_w_default(SimpleNamespace):
   GroupVolumeF = Field(default=MIN_VOL_F, ge=MIN_VOL_F, le=MAX_VOL_F,
                        description='Average output volume as a floating-point number')
   Disabled = Field(default=False, description='Set to true if not connected to a speaker')
+  # Port = Field(default=9000, description='Port used by LMS server for metadata collection')
```
rtertiaer, line 110:
```suggestion

```

amplipi/streams.py
```diff
@@ -1342,20 +1360,39 @@ def _activate(self, vsrc: int):
   def _deactivate(self):
     if self._is_running():
       try:
+        src_config_folder = f'{utils.get_folder("config")}/srcs/v{self.vsrc}'
+        os.system(f'rm -f {src_config_folder}')
         self.proc.terminate()
         self.proc.communicate()
+        if self.meta_proc is not None:
+          self.meta_proc.terminate()
```
rtertiaer, line 1368: can you communicate/poll/wait here to prevent zombies?

scripts/configure.py
```diff
@@ -158,6 +158,8 @@
   },
   'lms' : {
     'apt': ['libcrypt-openssl-rsa-perl', 'libio-socket-ssl-perl'], # needed for ShairTunes2W support
+    'copy' : [{'from': 'bin/ARCH/squeezelite', 'to': 'streams/squeezelite'},
```
rtertiaer, line 161: `squeezelite` is already installed on these boxes; can you clarify what versions we are going from and to with this?

rtertiaer (changes requested):
Couple of notes, no huge lifts I don't think. Can you also pre-squash this before the next review?

CHANGELOG.md
```diff
@@ -6,8 +6,11 @@
 * System
   * Add version number to config
   * Load arbitrary extra fields into /info from file
-  * Automatically mount usb storage devices as media drives in LMS
-  * Upgrade LMS to 8.5.1
+  * LMS
```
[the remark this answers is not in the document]
SteveMicroNova: We don't have an example of a subheader / subsection, it does seem like we prefer to make new headers rather than having subheaders within what would otherwise be the relevant section
Change made in upcoming commit

scripts/configure.py
SteveMicroNova: I continue to not remember why this line was added
I'm going to simply remove it as the specific LMS / squeezelite install is managed elsewhere, testing just now and removing this line changes nothing

SteveMicroNova:
Turning down the logging level of lms_metadata.py before anything else, logs are absolutely destroyed by that especially if you're in LMS mode with all 4 streams going

rtertiaer (changes requested):
can you squash this?

amplipi/streams.py
```diff
@@ -44,8 +44,12 @@
 
 # We use Popen for long running process control this error is not useful:
 # pylint: disable=consider-using-with
-logger = logging.getLogger('ampli-logger')
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+sh = logging.StreamHandler(sys.stderr)
```
rtertiaer, line 49:
```suggestion
sh = logging.StreamHandler(sys.stdout)
```

rtertiaer (changes requested):

rtertiaer (approved):

PR #666, opened 2024-03-11
Sourcemapping

SteveMicroNova:
Add sourcemap generation to Vite config
 
### Checklist

### DIFF
--- a/web/vite.config.js
+++ b/web/vite.config.js
@@ -32,6 +32,6 @@ export default defineConfig({
     },
   },
   build: {
-    sourcemap: "eval",
+    sourcemap: "inline",
   }
 })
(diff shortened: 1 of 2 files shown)


SteveMicroNova:
Simulated React minified error by instantiating a component that doesn't exist
The sourcemapped stuff only shows up in the console log, not the actual error screen
image

web/vite.config.js
```diff
@@ -31,4 +31,7 @@ export default defineConfig({
       },
     },
   },
+  build: {
+    sourcemap: "inline",
```
rtertiaer, line 35: can we build these not inline? it balloons the size of index.js from 500kb to 5mb :sweat_smile:

rtertiaer (changes requested):

web/vite.config.js
SteveMicroNova: I thought they didn't show up if they weren't inline but I was wrong, seems to work and it's down to 555 kb on my end again

PR #667, opened 2024-03-12
Pandora control

SteveMicroNova:
We forgot to add the pandora-specific controls of liking and disliking songs when we made webapp2
This PR readds them
 
### Checklist

### DIFF
--- a/amplipi/models.py
+++ b/amplipi/models.py
@@ -127,6 +127,19 @@ class BaseUpdate(BaseModel):
   name: Optional[str] = fields.Name
 
 
+class Rating(Enum):
+  # "rating" is a pandora-specific enum
+  # decode where text is common wording and ("text") is the pandora/pianobar terminology
+  # 0 default
+  # 1 liked ("loved")
+  # 2 disliked ("banned")
+  # 3 shelved ("tired")
+  DEFAULT = 0
+  LIKED = 1
+  DISLIKED = 2
+  SHELVED = 3
+
+
 class SourceInfo(BaseModel):
   name: str
   state: str  # paused, playing, stopped, unknown, loading ???
@@ -136,6 +149,7 @@ class SourceInfo(BaseModel):
   station: Optional[str]  # name of radio station
   img_url: Optional[str]
   supported_cmds: List[str] = []
+  rating: Optional[Rating] #Only used for pandora
 
 
 class Source(Base):
--- a/amplipi/models.py
+++ b/amplipi/models.py
@@ -127,7 +127,7 @@ class BaseUpdate(BaseModel):
   name: Optional[str] = fields.Name
 
 
-class Rating(Enum):
+class PandoraRating(Enum):
   # "rating" is a pandora-specific enum
   # decode where text is common wording and ("text") is the pandora/pianobar terminology
   # 0 default
@@ -149,7 +149,7 @@ class SourceInfo(BaseModel):
   station: Optional[str]  # name of radio station
   img_url: Optional[str]
   supported_cmds: List[str] = []
-  rating: Optional[Rating] #Only used for pandora
+  rating: Optional[PandoraRating] #Only used for pandora
 
 
 class Source(Base):
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -171,7 +171,6 @@ class BaseStream:
     """
     raise NotImplementedError(f'{self.name} does not support commands')
 
-
 class VirtualSources:
   """ Virtual source allocator to mind ALSA limits"""
 
@@ -779,6 +778,7 @@ class Pandora(PersistentStream):
             source.track = data[1]
             source.album = data[2]
             source.img_url = data[3].replace('http:', 'https:') # HACK: kind of a hack to just replace with https
+            source.rating = models.Rating(int(data[4]))
             source.station = data[5]
         return source
     except Exception:
@@ -795,10 +795,28 @@ class Pandora(PersistentStream):
     """
     try:
       if cmd in self.supported_cmds:
+        if cmd == "love":
+          src_config_folder = f'{utils.get_folder("config")}/srcs/v{self.vsrc}'
+          with open(f'{src_config_folder}/.config/pianobar/currentSong', 'r+', encoding='utf-8') as file:
+            data = file.read()
+            chunks = data.split(",,,")
+            if len(chunks) >= 5:
+              liked_state = chunks[4].strip()
+              if liked_state == "0":
+                chunks[4] = "1"
+              elif liked_state == "1":
+                chunks[4] = "0"
+            modified_data = ",,,".join(chunks)
+            file.seek(0)
+            file.truncate()
+            file.write(modified_data)
+            file.flush()
+
         with open(self.ctrl, 'w', encoding='utf-8') as file:
           file.write(self.supported_cmds[cmd]['cmd'])
           file.flush()
         expected_state = self.supported_cmds[cmd]['state']
+
         if expected_state is not None:
           self.state = expected_state
       elif 'station' in cmd:
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -795,17 +795,23 @@ class Pandora(PersistentStream):
     """
     try:
       if cmd in self.supported_cmds:
-        if cmd == "love":
+        if cmd == "love": # Pandora specific command
           src_config_folder = f'{utils.get_folder("config")}/srcs/v{self.vsrc}'
+          # Read the file that contains pandora metadata, flip an enum manually as it doesn't do so itself when sending this command
           with open(f'{src_config_folder}/.config/pianobar/currentSong', 'r+', encoding='utf-8') as file:
             data = file.read()
+            # This file contains information in a specific format, separated by 3 commas. Example below:
+            # Artist,,,Track,,,Album,,,Album_Art_URL,,,Liked_State,,,Station
             chunks = data.split(",,,")
-            if len(chunks) >= 5:
-              liked_state = chunks[4].strip()
-              if liked_state == "0":
-                chunks[4] = "1"
-              elif liked_state == "1":
-                chunks[4] = "0"
+            liked_state = chunks[4].strip()
+
+            # Chunk 5 (index 4) holds the liked state of a song for pandora
+            # Potential values are numbers 0-3, in order: Default, Liked, Disliked, Tired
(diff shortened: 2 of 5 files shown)


rtertiaer (changes requested):
this is great, glad it was an easy execute.

however, there's no feedback when I click "Love"; it's not clear if it's actually doing anything to the end user. that's definitely necessary; it feels like the button does nothing when i click it. is there maybe a quick effect you can put on the Icon to indicate a user actually clicked it? if it's a huge hassle to wait for the postCommand response, you probably don't need to do that.

SteveMicroNova:
> this is great, glad it was an easy execute.
> 
> however, there's no feedback when I click "Love"; it's not clear if it's actually doing anything to the end user. that's definitely necessary; it feels like the button does nothing when i click it. is there maybe a quick effect you can put on the Icon to indicate a user actually clicked it? if it's a huge hassle to wait for the postCommand response, you probably don't need to do that.

After this feedback I've been trying to CSS my way through this, but I think our present state management (and how we ask the server for data once a second and update state accordingly) disallows that
Here's what I was shooting for:
`    function Love() {
        const [clicked, setClicked] = React.useState(false);
        function onClick() {
            postCommand("love");
            setClicked(true);
            setTimeout(() => {setClicked(false);}, 5000);
        }
        if(true){
            return(
                <FontAwesomeIcon
                    style={{ color: clicked ? 'green' : 'white', transition: 'color 5s' }}
                    icon={faThumbsUp}
                    className={cmdToClassName("love")}
                    onClick={() => onClick}
                />
            )
        }
    }`
    
That way when you click like, it turns green and transitions to white again after 5 seconds
This doesn't work due to the sheer number of renders we experience, though; if you can think of a secondary option let me know but as it stands I think this request is larger scope than this PR has the right to correct for

rtertiaer:
I'd personally use [`onmousedown()` and `onmouseup()` handlers](https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_onmousedown) to modify the style; that way you don't need to handle any state. Not sure if that's the recommended way in React land, but that's probably a lot more straightforward if React can tolerate it (don't see why it wouldn't.)

klay2000:
what happens if you love or ban a song twice? If it toggles then we need a way to get this state directly from Pandora and change the button color. Furthermore, if this is possible putting the love/ban stuff in some state would be required which also solves your previous problem.

rtertiaer:
When you ban a song, it immediately transitions to the next song; it might still be a consideration for `love`'ing a song twice tho

linknum23:
This needs to be investigated via the pianobar interface.

SteveMicroNova:
Pianobar (the commandline app) does output the liked/disliked state in the song metadata stream; I can't seem to find it in our metadata stream and it seems that finding it would be a hell of a time atm
I swapped back to LMS metadata (#664) to avoid slamming my head into a brick wall all day today

SteveMicroNova:
I've just found that our current metadata scraper is fully capable of getting the liked status of a song, with an asterisk
That asterisk being that the metadata file only updates when the song changes, so liking a song won't show that it's liked until the next play
This suggests I'll need a way to store the liked status locally while also polling from the file, which could be messy but we'll see

web/src/components/MediaControl/MediaControl.jsx
```diff
@@ -19,6 +23,17 @@ import "./MediaControl.scss";
...
+
+    const [invertRating, setInverse] = React.useState(false); // Stores the clicked state of a song, that way it can be set to have a green thumb if the like is clicked as that metadata won't update in the backend until the song is played again
+    const [currentSong, setSong] = React.useState();
+
+    React.useEffect(() => {
+        if(currentSong != String(meta.track)){
+            setInverse(false);
+            setSong(String(meta.track));
+        }
+    }, [meta])
```
SteveMicroNova, line 36: useState on invertRating is subpar as leaving the page and reentering it, or simply entering it as a different user, will make it unclear if the song is liked or not when it was freshly liked

I do not know how to solve this as changing the internals of useStatusStore feels unwise as someone who doesn't understand it (and also it looks fragile)

SteveMicroNova: Klayton suggested looking into what the disable checkboxes do to store bool data for streams or zones, but those just update house.json once you hit "accept" or "submit"
I'm uncertain that what we want is to have the metadata file be manually rewritten by amplipi when we hit the like button, that sounds dangerous since that is managed by pianobar directly and should likely stay that way

Also I've looked into it and useStatusStore and other persistent storage methods seem to also be not quite what I want, creating a status store in App.jsx just to store temporary data for one stream seems bad too

amplipi/streams.py
```diff
@@ -789,10 +789,28 @@ def send_cmd(self, cmd):
...
+              liked_state = chunks[4].strip()
+              if liked_state == "0":
+                chunks[4] = "1"
+              elif liked_state == "1":
+                chunks[4] = "0"
+            modified_data = ",,,".join(chunks)
+            file.seek(0)
+            file.truncate()
+            file.write(modified_data)
+            file.flush()
```
SteveMicroNova, line 807: Inside the command handling function, intercept the love command and rewrite the metadata file with the opposite liked state
This is gross but also the only realistic way to have the liked status be visible, so oh well ig

amplipi/models.py
```diff
@@ -135,6 +135,13 @@ class SourceInfo(BaseModel):
   station: Optional[str]  # name of radio station
   img_url: Optional[str]
   supported_cmds: List[str] = []
+  rating: Optional[int]
+  # "rating" is a pandora-specific enum
+  # decode where text is common wording and ("text") is the pandora/pianobar terminology
+  # 0 default
+  # 1 liked ("loved")
+  # 2 disliked ("banned")
+  # 3 shelved ("tired")
```
rtertiaer, line 144: can you use Python 3's `enum` here?

web/src/components/MediaControl/MediaControl.jsx
```diff
@@ -19,6 +23,8 @@ import "./MediaControl.scss";
 const MediaControl = ({ selectedSource }) => {
     // const source = status.sources[selectedSource]
     const source = useStatusStore((s) => s.status.sources[selectedSource]);
+    const meta = useStatusStore((s) => s.status.sources[selectedSource].info);
+    console.log(meta);
```
rtertiaer, line 27:
```suggestion

```

rtertiaer (changes requested):

amplipi/models.py
SteveMicroNova: I don't really understand `enum` but adding it as I just did seems to work after testing so I guess that's fine to do
I left the comment block in as it still belongs to translate terminology as well as the use for the "rating" variable

rtertiaer (changes requested):
to use an `enum`, [first define a subclass of `Enum`](https://github.com/micro-nova/AmpliPi/blob/08cafa1da9e7a38bcda15c205fd433a416fb501b/amplipi/display/common.py#L32). then you can [reference its user friendly representation elsewhere](https://github.com/micro-nova/AmpliPi/blob/08cafa1da9e7a38bcda15c205fd433a416fb501b/amplipi/display/common.py#L75). It helps both the developer, and the type-checker. Your comment about what each numbered state means belongs next to the type definition.

amplipi/models.py
```diff
@@ -127,6 +127,19 @@ class BaseUpdate(BaseModel):
   name: Optional[str] = fields.Name
 
 
+class Rating(Enum):
```
rtertiaer, line 130: can you rename this to be a PandoraRating or something that makes it explicit what this type is?

rtertiaer (changes requested):

amplipi/streams.py
```diff
@@ -795,10 +795,28 @@ def send_cmd(self, cmd):
...
+          src_config_folder = f'{utils.get_folder("config")}/srcs/v{self.vsrc}'
+          with open(f'{src_config_folder}/.config/pianobar/currentSong', 'r+', encoding='utf-8') as file:
+            data = file.read()
+            chunks = data.split(",,,")
+            if len(chunks) >= 5:
+              liked_state = chunks[4].strip()
+              if liked_state == "0":
+                chunks[4] = "1"
+              elif liked_state == "1":
+                chunks[4] = "0"
```
rtertiaer, line 808: what does 0 and 1 mean here? maybe it makes sense to use the new Enum

amplipi/streams.py
```diff
@@ -795,10 +795,28 @@ def send_cmd(self, cmd):
...
+            modified_data = ",,,".join(chunks)
+            file.seek(0)
+            file.truncate()
+            file.write(modified_data)
+            file.flush()
```
rtertiaer, line 813: you open the file `r+` then truncate and write the file. [that's probably not what you want](https://docs.python.org/3/library/functions.html#open).

amplipi/streams.py
```diff
@@ -795,10 +795,28 @@ def send_cmd(self, cmd):
     """
     try:
       if cmd in self.supported_cmds:
+        if cmd == "love":
+          src_config_folder = f'{utils.get_folder("config")}/srcs/v{self.vsrc}'
+          with open(f'{src_config_folder}/.config/pianobar/currentSong', 'r+', encoding='utf-8') as file:
+            data = file.read()
+            chunks = data.split(",,,")
+            if len(chunks) >= 5:
```
rtertiaer, line 803: can you clarify in a comment what `len(chunks) >= 5` means in this context? there's a bunch of data parsing here but no description of the expected format or what you are attemping to scrape out. should someone else come along and have to debug this logic (for example, if the upstream format changes), this chunk of code is almost easier to scrap and rewrite than figure out what it intends to do.

rtertiaer (changes requested):

amplipi/streams.py
SteveMicroNova: r+ opens the file for reading, with no truncation. Using w+ would truncate the file immediately before I get a chance to read, making it so I cannot edit based on what data is already present

I read the file, see what there is, and make the changes I want to a local tempfile, then truncate and write afterwards as a means of overwriting the original file with the contents of the tempfile,. An alternative would be opening as r and then later opening as w but I feel that's likely less efficient than simply seeking and truncating manually

rtertiaer: ahhh gotcha, fair enough.

rtertiaer (approved):

PR #668, opened 2024-03-12
remove noisy console.log() from MenuBar

rtertiaer:
This PR removes a `console.log()` that prints a message like:
```
sourceInputType MenuBar: none MenuBar.jsx:48:12
```

every 1s in the browser. It is probably a leftover from some more active development. Explicitly not adding this change to the CHANGELOG.
 
### Checklist


klay2000 (approved):
ty, I kept meaning to track this down

PR #670, opened 2024-03-18
Robustify firmware programming

### LINKED ISSUE #587: Robustify firmware programming when expansion units present
Programming seems to succeed less often the more expanders that are connected.

Blocked on getting 2 expanders of the latest hardware.

Lohrer:

Fix #587.

* Provides a new preamp firmware version 1.8, which removes the debug printouts that cause bootloading to sometimes fail when expanders are present.
* Retries firmware flashing a few times if initial communication fails.
 
### Checklist


Lohrer:
@SteveMicroNova  I know you're working on logging infrastructure changes and `hw.py` is a bit of a black sheep since it's used largely for scripting the firmware update process. What I added here works for firmware programming, but could you just verify that it will work in the bigger picture of your new logging setup?

amplipi/hw.py
```diff
@@ -39,6 +40,8 @@
 
 PI_SERIAL_PORT = '/dev/serial0'
 
+log = logging.getLogger('ampli-hw')
```
SteveMicroNova, line 43: Change `ampli-hw` to `ampli-logger` and it should automatically plug in once both branches are put together
An optional change would be to change `log` to `logger` as I've been using `logger` in every file thus far; but that's a purely stylistic thing for consistency and not a functional change
Also I could swap `logger` to `log`, that's more down to what variable name we want to keep in common around the files

SteveMicroNova (changes requested):
I've got 1-2 changes to make to make this compatible with my stuff, otherwise lgtm :+1:

rtertiaer (approved):

amplipi/hw.py
rtertiaer: I can see this both ways. Cuz this is largely run as a script, it's not likely to run thru the existing logger anyhow... but I could also envision a world where we would run this not as a script but as some automated process orchestrated by the AmpliPi.

I don't have a problem with it as-is. It's an easy change if we do run it through AmpliPi, and we may not even want to do that change if we want to segregate the logging for hardware stuff.

SteveMicroNova (approved):

amplipi/hw.py
Lohrer: Sorry, the display code was using `log` so I stuck with what I had done before.

As for the logger name, I'm not super familiar with how this logger works but would it make sense/be possible to set one logger name in `if __name__ == '__main__':` (so if called as a script) and another if imported?

rtertiaer: That's definitely a sensible approach. Since we have no case right now where `__name__ != '__main__'`, it feels a little bit like extra credit to prepare for that case right now.

PR #680, opened 2024-04-16
Load arbitrary extra fields into /info from a file

rtertiaer:
This PR permits adding arbitrary extra fields to the `/info` endpoint, loaded from a JSON file on the local filesystem. This permits some amount of third party extensibility without tighter integration.

I've tested these cases:
* valid JSON is loaded into the `/info` endpoint as expected, including non-flat structures
* invalid JSON causes the `extra_fields` key to be omitted from the endpoint
* the file not existing causes the `extra_fields` key to be omitted from the endpoint
 
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   * Reverse order available sources are used in
 * System
   * Add version number to config
+  * Load arbitrary extra fields into /info from file
 
 ## 0.3.4
 * Web App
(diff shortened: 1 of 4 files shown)


linknum23 (commented):
Any chance this can be cached so we don't hit the file system on every request? Looking at this code we appear to do the same with version, oops! 

Also union[None,Stuff] is typically written as Optional[Stuff]

rtertiaer:
I even knew that bit about Optional (see the model changes), I just blanked :woozy_face: 

The way I'm reading it, only `reinit` hits this path, which doesn't happen on every request - am I mistaken?

linknum23 (approved):

linknum23:
Sorry I got confused about where that was being populated. Looks good!

PR #684, opened 2024-04-17
Add a tested checkbox to PR template

rtertiaer:
This updates the PR template to include an affirmation that the proposed changes have been tested.


SteveMicroNova (approved):
It may also make sense to have a "Did you include a guide on how to test these changes" thing as well while we're at it, or at least suggest that the PR author do such if not add it to the checkbox

PR #692, opened 2024-04-22
Pandora config

klay2000:
Adds infrastructure for song browsing and implements it with Pandora stations.


amplipi/streams.py
```diff
@@ -820,19 +864,73 @@ def send_cmd(self, cmd):
...
+          j = i.rstrip("\n").split(":")
+          image = ""
+          for i in pd_stations:
+            if(i[0].upper() == j[1].upper()):
+              image = i[1]
+              break
+
+          self.stations.append(models.BrowsableItem(name=j[1], playable=True, id=j[0], parent=False, img=image))
+
+    return self.stations
```
linknum23, line 928:
```suggestion
  def browse(self, _) -> List[models.BrowsableItem]:
    """ Browse the stream for items """

    if not self.set_stations:
      self.set_stations = True # TODO: retry on failure?
      pd_stations : Dict[str,str] = dict()
      try:
        pd_stations = {s.name.upper() : s.art_url for s in self.pyd_client.get_station_list()}
      except Exception as e:
        logger.exception(f'Error browsing for pandora stations: {e}')
        return []
      with open(self.pb_stations_file) as f:
        # try to match PianoBar's list of stations with those returned by the Pandora API
        # NOTE: duplicate station names will only match the last duplicate station returned by the Pandora API
        for l in f.readlines():
          sinfo = l.rstrip("\n").split(":")
          if sinfo.len >= 2:
            station_id = sinfo[0]
            name = sinfo[1]
            img = pd_stations.get(name.upper(), "")
            self.stations.append(models.BrowsableItem(name=name, playable=True, id=station_id, parent=False, img=img))
    return self.stations
```

linknum23:
This all looks good. Let me know when there is an amplipi with this running on it that I can test with.

linknum23 (approved):

PR #698, opened 2024-04-29
PandoraControls Width

SteveMicroNova:
I pushed an update without testing for UX design on multiple platforms, causing the new Pandora Controls to flow off the screen on mobile
This fixes that by making the buttons smaller

### DIFF
--- a/web/src/components/MediaControl/MediaControl.scss
+++ b/web/src/components/MediaControl/MediaControl.scss
@@ -8,6 +8,7 @@
   justify-content: center;
   margin-top: 1rem;
   margin-bottom: 1rem;
+  width: 90vw;
 }
 
 .media-inner {
@@ -18,10 +19,24 @@
 }
 
 .media-control {
-  font-size: 3.5rem;
   color: general.$controls-color;
   padding-left: 1.5rem;
   padding-right: 1.5rem;
+
+
+  @media (max-width: 365px) {
+    font-size: 1.5rem;
+    padding-left: 1.25rem;
+    padding-right: 1.25rem;
+  }
+
+  @media (min-width: 365px) and (max-width: 425px) {
+    font-size: 2.5rem;
+  }
+
+  @media (min-width: 425px) {
+    font-size: 3.5rem;
+  }
 }
 
 .media-control-disabled {
--- a/web/src/pages/Player/Player.jsx
+++ b/web/src/pages/Player/Player.jsx
@@ -72,14 +72,19 @@ const Player = () => {
                     <StreamBar sourceId={selectedSourceId} onClick={() => {setStreamsModalOpen(true);}}/>
                 </Chip>
             </div>
+            {/* There are many sub-divs classed player-inner here because formatting was strange otherwise */}
             <div className="player-inner">
                 <img src={img_url} className="player-album-art" />
+            </div>
+            <div className="player-inner">
                 <SongInfo
                     sourceId={selectedSourceId}
                     artistClassName="player-info-title"
                     albumClassName="player-info-album"
                     trackClassName="player-info-track"
                 />
+            </div>
+            <div className="player-inner">
                 <MediaControl selectedSource={selectedSourceId} />
             </div>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * Add firmware version for main and expansion units on About page
   * Limit length of displayed stream names with ellipsis
   * Add restart stream button to stream player and stream modal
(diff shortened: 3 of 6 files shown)


rtertiaer (changes requested):
can you update the changelog pls?

rtertiaer (changes requested):
also, per the PR checklist that you removed, please affirm that you have tested these changes and that they work, across multiple platforms (ios, android, firefox, chrome at a minimum.)

SteveMicroNova:
Chrome and Firefox are good, as well as my phone (mobile safari) and all of the default emulatable screen sizes for mobile devices in Chrome

SteveMicroNova:
Needs more testing with long metadata due to the div change before I'm willing to merge this
Will do that tomorrow

web/src/components/MediaControl/MediaControl.scss
```diff
@@ -18,12 +22,23 @@
...
   padding-right: 1.5rem;
 }
 
+@media screen and (max-width: 320px) {
+  .media-control {
+    font-size: var(--min-font-size);
+  }
+}
+
+@media screen and (min-width: 360px) {
```
rtertiaer, line 36: why are these two screen sizes different? perhaps I just don't understand CSS.

SteveMicroNova: That's because the space between is theoretically a space where it would transform percents at a time between the min and the max when I was trying to set the max to be the original 3.5 rem for a PC screen and 1 rem for the smallest of small, ideally keeping somewhere around the 2.5 rem that worked for most phones somewhere inbetween

That wasn't working so I made the max be 2.5 and so the scaling by percents here is kinda vestigial and unimportant, will remove just for the sake of clarity tomorrow

web/src/pages/Player/Player.jsx
```diff
@@ -72,14 +72,19 @@ const Player = () => {
...
+            </div>
+            <div className="player-inner">
                 <SongInfo
                     sourceId={selectedSourceId}
                     artistClassName="player-info-title"
                     albumClassName="player-info-album"
                     trackClassName="player-info-track"
                 />
+            </div>
+            <div className="player-inner">
```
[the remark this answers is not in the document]
SteveMicroNova: Un-divided album art from metadata, kept separation between controls and the others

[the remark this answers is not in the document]
SteveMicroNova: Re-divided those because album art doesn't scale down on sufficiently tiny screens, but still kept text in line with its leftmost side

rtertiaer (commented):
can you squash this PR pls?

web/src/pages/Player/Player.jsx
```diff
@@ -72,14 +72,19 @@ const Player = () => {
...
                 <SongInfo
                     sourceId={selectedSourceId}
                     artistClassName="player-info-title"
                     albumClassName="player-info-album"
                     trackClassName="player-info-track"
                 />
+                </div>
+                <div className="player-inner">
```
rtertiaer, line 87: you open a div here but don't close it?

rtertiaer (changes requested):

web/src/pages/Player/Player.jsx
SteveMicroNova:
```
<div className="player-inner">
                <img src={img_url} className="player-album-art" />
            </div>
            <div className="player-inner">
                <SongInfo
                    sourceId={selectedSourceId}
                    artistClassName="player-info-title"
                    albumClassName="player-info-album"
                    trackClassName="player-info-track"
                />
                </div>
                <div className="player-inner">
                <MediaControl selectedSource={selectedSourceId} />
            </div>
```
I close a preexisting div and open a new one, inheriting the original ones closing tag

rtertiaer: can you fix the indentation then?

web/src/components/MediaControl/MediaControl.scss
```diff
@@ -18,10 +19,24 @@
...
+}
+
+@media (min-width: 365px) and (max-width: 425px) {
+  font-size: 2.5rem;
+}
+
+@media (min-width: 425px) {
+  font-size: 3.5rem;
+}
 }
```
rtertiaer, line 40: is this correct? if it isn't the indentation needs to be corrected

SteveMicroNova: Only problem was a tabindex, fixed in recent commit

rtertiaer (approved):

PR #700, opened 2024-04-29
Fix: Zones playing audio on source used for announcement are not muted…

### LINKED ISSUE #699: Bug: Zones playing audio on source used for annoncement are not muted while the announcement is playing
### Context
AmpliPi has 4 sources that can each play audio. One of these sources needs to be used to make a PA announcement. If that source is in use during an announcement, the playback is stopped temporarily and restarted after the announcement finishes. 

By default announcements use source 4 and play to all zones at 50% volume. Optionally only some zones can be played to using the `zones` or `groups` field to specify them.

### Problem
A user (@rwarner) reported that when an announcement is played to a select few zones, other zones not specified also play the audio. Those unspecified zones also happen to be connected the the source that was playing audio on the source used for the announcement. See the detailed report here:  https://amplipi.discourse.group/t/announce-endpoint-zone-specific/304/11 

### Solution
Mute the effected zones during the announcement. Unmute them afterward.

linknum23:
… while the announcement is playing

Closes #699
 
### Checklist

* [x] Have you written new tests for your core features/changes?
* [x] If this is a bugfix, is there a test that fails before the fix is applied?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * Streams
   * Add DLNA metadata and control support
   * Add support for browsing Pandora stations
+* API
+  * Fix: Zones playing audio on source used for announcement are not muted while announcement is playing
 
 ## 0.3.5
 * Web App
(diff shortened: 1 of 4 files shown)


linknum23:
@SteveMicroNova Can you check if these changes make sense?

SteveMicroNova (approved):
Everything here makes sense to me, you check what zones need to be muted, which of those zones aren't muted and then mute said zones, then return them to previous state afterwards

PR #701, opened 2024-04-30
PandoraLoveFix

SteveMicroNova:
Fix failing pandora command tests
 
Please test for me, I was having issues where every test passed on my machine but not on others

### Checklist


amplipi/streams.py
```diff
@@ -827,26 +827,33 @@ def send_cmd(self, cmd):
...
+              chunks = data.split(",,,")
+              try:
+                liked_state = models.PandoraRating(int(chunks[4].strip()))
+              except:
+                liked_state = models.PandoraRating.DEFAULT
+
+              # Chunk 5 (index 4) holds the liked state of a song for pandora
+              # Potential values are numbers 0-3, in order: Default, Liked, Disliked, Tired
+              # This code flips it from Default to Liked and vice-versa when you hit the like button, as pianobar only updates metadata when songs start
+              if liked_state == models.PandoraRating.DEFAULT.value:
```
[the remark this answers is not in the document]
SteveMicroNova: We don't support tired and if you dislike a song it immediately skips it so there's no need to change the color of that button

This bit of code is only to flip the thumbs up button from white to green and back to white if it's unliked

linknum23:
I would add that to the comment here.

linknum23:
Tested that the 'love' tests work. I'm trying to use those tests gingerly so we don't blow through our hidden quota again.

amplipi/streams.py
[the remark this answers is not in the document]
linknum23: Looking at how this works again I see a potential race condition in how the pandora rating is stored. Imagine that you thumb up a song right as it is changing to the next song.

1. read currentSong
1. currentSong gets updated to new song
1. we update currentSong with the new Rating ( but the old song is written)
1. AmpliPis interface shows the old song until the pandora station advances to the next song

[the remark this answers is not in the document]
linknum23: To avoid this race condition we could keep a shadowed current song. The rating code would grab the song from the shadowed current song, if it matches current song it returns the shadowed rating.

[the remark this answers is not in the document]
SteveMicroNova: I think I fixed this in https://github.com/micro-nova/AmpliPi/pull/701/commits/ce61189bfac80557f12b39c6600990d8ac43b1c4, but I feel the change I did there was potentially too large for this scope

I couldn't find a way to easily grab the current metadata state from this spot so I made one. All tests passed when I tried them again.

amplipi/streams.py
```diff
@@ -801,14 +803,26 @@ def info(self) -> models.SourceInfo:
       with open(loc, 'r', encoding='utf-8') as file:
         for line in file.readlines():
           line = line.strip()
+          if self.track != data[1]: # When song changes, stop inverting state
```
linknum23, line 806: This read before assignment doesn't cause an exception?

linknum23:
Note: you should be able to test love with this line: `pytest tests/test_rest.py -vvv -k 'post_stream_cmd_live[clientnm1-love]' --capture=no`

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Streams
   * Add DLNA metadata and control support
   * Add support for browsing Pandora stations
+  * Bug fix on liking song on Pandora
```
linknum23, line 9:
```suggestion
  * Make Pandora like work and pass tests without metadata race condition
```

amplipi/streams.py
```diff
@@ -825,28 +839,8 @@ def send_cmd(self, cmd):
...
-              chunks[4] = models.PandoraRating(1)
-            elif liked_state == "1":
-              chunks[4] = models.PandoraRating(0)
-            modified_data = ",,,".join(chunks)
-            file.seek(0)
-            file.truncate()
-            file.write(modified_data)
-            file.flush()
+        if cmd == "love":
+            self.invert_liked_state = not self.invert_liked_state
```
linknum23, line 843:
```suggestion
            self.info() # make sure the liked state is synchronized to the current song
            self.invert_liked_state = not self.invert_liked_state
```

amplipi/streams.py
```diff
@@ -803,12 +805,24 @@ def info(self) -> models.SourceInfo:
...
+
+            # Pianobar doesn't update metadata after a song starts playing
+            # so when you like a song you have to change the state manually until next song
+            if self.invert_liked_state and int(data[4]) == 0:
+                source.rating = models.PandoraRating(1)
+            elif self.invert_liked_state and int(data[4]) == 1:
+              source.rating = models.PandoraRating(0)
+            else:
+              source.rating = models.PandoraRating(int(data[4]))
+
```
linknum23, line 825:
```suggestion
            initial_rating = models.PandoraRating(int(data[4]))
            source.rating = initial_rating
            if self.invert_liked_state:
              if initial_rating == models.PandoraRating.DEFAULT:
                source.rating = models.PandoraRating.LIKED
              elif initial_rating == models.PandoraRating.LIKED:
                source.rating = models.PandoraRating.DEFAULT
```

linknum23: I think this reads a tad better since it uses the actual Pandora rating types instead of magic/unnamed integers.

linknum23 (approved):
Looks good. Remember to squash your commits and rebase on main.

PR #708, opened 2024-05-07
SongInfoStyle

SteveMicroNova:
Add customizable style for SongInfo, right align metadata text on player card on Home page and limit the max size of metadata

Metadata on player cards can take up a lot of space, reduce that max space and wrap text as necessary

See images in first comment for a diff
 
### Checklist

* [ ] If this is a UI change, have you tested it across multiple mobile and web platforms?


SteveMicroNova:
Old:
image

New:
image

SteveMicroNova:
On mobile:

Old:
IMG_2119

New:
IMG_2120

SteveMicroNova:
I'll also note: These are largely stylistic choices, I know the maxwidth thing assisted with some fighting for space but the right text align just looks better on player cards imo so you don't have a blob of text floating in the middle of the card

SteveMicroNova:
Was testing on a unit capable of multi-zone play, found more bugs, fixed them
I've made it so that the zone names are dynamically moved around to make space for the longer ones as needed to not cover or otherwise impact metadata
Note that it is still possible to cover up metadata with a zone name, but only if you have a single extremely long word

Old:
image

New:
image

SteveMicroNova:
Looking into ways of potentially limiting max size in previous screenshots

SteveMicroNova:
> Looking into ways of potentially limiting max size in previous screenshots

Figuring this out is nontrivial and has minimal dividends so I am no longer pursuing it

rtertiaer (commented):
what github issue does this address?

not a huge fan of the wrapping change, and I dont' think you've necessarily made the case that a) this is a problem we should spend time solving, and b) this is the best solution to that problem. making these boxes variable height is the quickest way to make this look messy imo.

SteveMicroNova:
> what github issue does this address?

None to my knowledge, it was a change added in #703 that I added to attempt to prevent zones overlapping with metadata (generally the metadata is long, I tried and failed to do the same with a long zone)

> not a huge fan of the wrapping change, and I dont' think you've necessarily made the case that a) this is a problem we should spend time solving, and b) this is the best solution to that problem. making these boxes variable height is the quickest way to make this look messy imo.

I'm not sure I like the vertical resizing either, that was just the easiest way to go about fixing this. I could technically also do dynamic font scaling, but that sounded messy

web/src/components/CustomMarquee/CustomMarquee.jsx
```diff
@@ -0,0 +1,69 @@
...
+        setDataScroll(false);
+        setDataScrollOverride(true);
+
+        setDataOverrideTimeout(setTimeout(()=>{setDataScroll(true); setDataScrollOverride(false);}, 2000)); // When a scroll completes, pause for two seconds and restart (unless the timeout is cleared)
+    }
+
+    setTimeout(()=>{setDataScrollOverride(false);}, 2000) // used to kick off useEffect after two seconds of page load
+
+    return(
+        <Marquee play={dataScroll} speed={30} delay={2} onCycleComplete={() => {setOverride()}}>
```
rtertiaer, line 53: you dedicate a lot of state to managing a scrolling delay, but your upstream Marquee library already has a `delay` param. why is your delay state necessary? can it be removed, such that we don't trigger so many re-renders with changes in `useState()` things?

rtertiaer (changes requested):

web/src/components/PlayerCard/PlayerCardFb.jsx
```diff
@@ -75,7 +75,7 @@ const PlayerCardFb = ({ sourceId, setVol }) => {
                             <StreamerOutputBadge sourceId={sourceId} />
                         </div>
                     )}
-                    <SongInfo sourceId={sourceId} />
+                    <SongInfoMarquee sourceId={sourceId} style={{maxWidth: "50%", textAlign: "right"}} />
```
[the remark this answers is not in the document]
SteveMicroNova: There was a good reason back when SongInfo was used in both the player view and the home view, to not effect both of them
I split them out so this isn't actually necessary anymore

web/src/components/SongInfo/SongInfoMarquee.jsx
```diff
@@ -0,0 +1,48 @@
...
+    style,
+}) => {
+    const info = useStatusStore((state) => state.status.sources[sourceId].info);
+    const artistRef = React.useRef(null);
+    const albumRef = React.useRef(null);
+    const trackRef = React.useRef(null);
+
+    artistClassName = "artist-name " + artistClassName;
+    albumClassName = "album-name " + albumClassName;
+    trackClassName = "track-name " + trackClassName;
```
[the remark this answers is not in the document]
SteveMicroNova: I'm unclear as to why we do a += to the classnames, that was code that already existed in the original SongInfo that this was forked off of
The useRefs are just a React approved way of doing something along the lines of `document.getElementById()`

web/src/components/CustomMarquee/CustomMarquee.jsx
SteveMicroNova: https://www.react-fast-marquee.com/documentation#delay
The built-in delay only works on initial render, not on every cycle completion

web/src/components/MenuBar/MenuBar.scss
```diff
@@ -2,6 +2,7 @@
 
 .bar {
   position: fixed;
+  z-index: 1;
```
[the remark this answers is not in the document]
SteveMicroNova: image
Avoiding the Marquee landing above the MenuBar on scroll, adding comment to that effect in next commit

web/src/components/SongInfo/SongInfoMarquee.jsx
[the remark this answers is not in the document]
rtertiaer: where do all the `ref` values come from? I see you set them to null here; where are they populated?

what is the purpose of this line? 
```
  artistClassName = "artist-name " + artistClassName;
```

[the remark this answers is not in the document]
SteveMicroNova: `
    return (
        <div className="song-info song-info-marquee">
            <div ref={artistRef} className={artistClassName}>
                <CustomMarquee data={info.artist} containerRef={artistRef} />
            </div>

            <div ref={albumRef} className={albumClassName}>
                <CustomMarquee data={info.album} containerRef={albumRef} />
            </div>

            <div ref={trackRef} className={trackClassName}>
                <CustomMarquee data={info.track} containerRef={trackRef} />
            </div>
        </div>
    );
    `

The refs are populated by the elements being assigned a given ref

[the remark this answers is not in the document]
SteveMicroNova: As for the ClassNames, @klay2000 will have to answer for that as he (or Jonah) created those. I assume they're a way of passing external classes in while also defining a baseline class for each portion, though I don't know if there's any classes that are ever passed in in practice

rtertiaer (commented):
is the original `SongInfo.jsx` still in use, or can it be removed?

I'm still hesitant to approve this, because I see no less than 5 `useState`s that can cause re-renders and a `useEffect` that watches 3 of them. Is there any possible way to reduce the amount of state stored and reacted upon?

web/src/components/SongInfo/SongInfoMarquee.jsx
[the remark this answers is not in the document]
klay2000: I believe this was necessary to re-use song info on the stream card and the player view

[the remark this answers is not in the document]
SteveMicroNova: With this clarification I have removed them as this PR splits SongInfo.jsx into two files, one for the home screen player cards and another for the player view metadata

SteveMicroNova:
This is leaking off the rightmost edge of playercards atm, needs more work but I have no opportunity to dig that deep atm

web/src/components/CustomMarquee/CustomMarquee.jsx
```diff
@@ -0,0 +1,75 @@
...
+function CustomMarquee(props) {
+    const {
+        data,
+        containerRef,
+    } = props;
+
+    const windowSize = React.useRef(window.innerWidth);
+    const resizeCooldown = React.useRef(false);
+
+    const dataRef = React.useRef(null);
```
rtertiaer, line 15: where do you set this?

web/src/components/CustomMarquee/CustomMarquee.jsx
```diff
@@ -0,0 +1,75 @@
...
+        dataScrollOverride.current = true;
+
+        dataOverrideTimeout.current = setTimeout(()=>{dataScrollOverride.current = false; assessMarquee();}, 2000); // When a scroll completes, pause for two seconds and restart (unless the timeout is cleared)
+    }
+
+    setTimeout(()=>{dataScrollOverride.current = false; assessMarquee();}, 2000) // used to kick off useEffect after two seconds of page load
+
+
+    function handleResize(){
+        if(!resizeCooldown.current && window.innerWidth != windowSize.current){
```
rtertiaer, line 46: is the `innerWidth != windowSize.current` necessary? feels like you kinda want to fire on every resize where cooldown isn't doing the thing; probably will get rid of some extra state too

rtertiaer (commented):
what other solutions have you examined for this problem? i feel like you haven't convinced me that, out of the entire solution space, doing this in javascript, importing a library, and wrapping it ourselves with a ton of state is the correct solution.

web/src/components/CustomMarquee/CustomMarquee.jsx
SteveMicroNova: I'm using refs as a document.getElementById, which you can't use directly because react elements are too transient to have an ID at all times
https://react.dev/learn/manipulating-the-dom-with-refs#getting-a-ref-to-the-node

You send a null useRef to the value of a ref prop on an element and it magics it together to be a functional equivalent to document.getElementById for me to read the innerWidth of

rtertiaer: where do you set this tho? i don't see a `dataRef.current = [...]` anywhere

SteveMicroNova: I suppose that link wasn't a perfect example

You don't need to set the value of `current` directly at any point in some cases, this being one of them

I set the value of that ref in the div inside of the Marquee, and I use that to measure the width of the content vs the width of the marquee to decide if to should or shouldn't scroll

```

        <Marquee play={dataScroll} speed={30} onCycleComplete={() => {setOverride()}}>
            <div
                style={{marginLeft: "10px"}}
                className="marquee-text"
                ref={dataRef} // Here
            >
                {data}
            </div>
        </Marquee>
```

And then during assessMarquee(), I set the value of a const to the value of dataRef.current so I can use that ref as a reference to the element and measure the relative widths

SteveMicroNova:
> what other solutions have you examined for this problem? 

I answered this in a slack DM, but to answer it publicly as well:
I've tried doing it myself in CSS and that proved to have far too many levers that were all extremely sensitive to any little change and it made the animation look very chunky and didn't fully scroll like I wanted to (instead snapping back to the start position when the text was fully offscreen rather than wrapping back around to come to a stop), so this was the only alternative that I could find
The library I'm using is [depended on by 34k+ repos](https://github.com/justin-chu/react-fast-marquee/network/dependents) so I had the feeling it was a good route to go down even if it lacks a small amount of tooling that I've managed to create with that state handling
I can pare down the state a little more yet, that windowSize ref is uneccessary and was only there because I blindly turned almost all states into refs instead when it can realistically be just the window.innerWidth at all times

web/src/components/CustomMarquee/CustomMarquee.jsx
```diff
@@ -0,0 +1,72 @@
+import React from "react";
+import Marquee from "react-fast-marquee";
+
+import PropTypes from "prop-types";
+
+function CustomMarquee(props) {
+    const {
+        data,
```
rtertiaer, line 8: can you please use `Children` for encapsulating react components like this?

web/src/components/CustomMarquee/CustomMarquee.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+    function assessMarquee() {
+        const container = containerRef.current;
+        const scroll = dataRef.current;
+
+        if(container && scroll){
+            if(container.offsetWidth < scroll.scrollWidth && !dataScrollOverride.current){
+                setDataScroll(true);
+            } else if(container.offsetWidth >= scroll.scrollWidth && dataScrollOverride.current){ // dataScroll is only able to be set to false when a scroll has been completed to avoid randomly jumbled text
+                clearTimeout(dataOverrideTimeout.current);
```
rtertiaer, line 28: i'd like to be a bit more rigorous here. there are a couple of unhandled cases:
* `container.offsetWidth < scroll.scrollWidth && dataScrollOverride.current`
* `container.offsetWidth >= scroll.scrollWidth && !dataScrollOverride.current`

i see here that the data/control flows thru two different paths:
* one with `setDataScroll()`: you call `setOverride()` thru `onCycleComplete()`, which sets `dataScrollOverride`
* and another with `clearTimeout()`, which cancels `dataOverrideTimeout` and thus never sets `dataScrollOverride.current = false`

is there any way you can either shore up this conditional to handle the other cases, or flatten the control flow such that the actions triggered by this conditional aren't behind a couple of layers of indirection that need to be correctly executed to function the next time `assetssMarquee()` is called?

rtertiaer (changes requested):

web/src/components/CustomMarquee/CustomMarquee.jsx
SteveMicroNova: I feel that those unhandled cases can remain unhandled, and simply continue to do whatever it was already doing do to how dataScrollOverride is handled:
dataScrollOverride is set to true at the end of every marquee cycle, and then reassesses whether it should be true or false after a two second delay. If it is left true then any event that could turn it false triggers a reassessment of the state
I would move the dataScrollOverride check into the play={} state of the Marquee element to make reading easier, but doing so makes it so that making the screen wider suddenly can stop the marquee in the wrong spot as soon as the screen is wide enough to fit it.

As for the data flow concern, I don't think it's that bad? There's a binary state: playing or not. That relies on two other binaries, one being "does the info fit inside the space given to it" and the other is "is the scrolling paused", if the scrolling is paused it gets unpaused after 2 seconds max, or doesn't unpause if the screen is now wide enough to fit the content. It's done this way to ensure the content always stops in the correct position, getting rid of the scrollOverride and instead simply using the playing state would cause the text to stop in weird spots instead of having a well defined stopping point and not having timers makes the data scroll constantly, which looks bad.

I feel like graphing the flow out would make it look nicer / less complicated, but I'm not doing that unless I'm at the whiteboard and I'm not able to spend the time to do that today.

rtertiaer (approved):
:sunglasses: :+1: 

thanks for your hard work here. it's gorgeous.

PR #709, opened 2024-05-07
Invert stream list

SteveMicroNova:
Invert the order of listed streams, so when you add a new stream it appears at the bottom instead of the top while retaining the reason that it was inverted in the first place in #674 
 
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Upcoming Release
 * Web App
   * Add initial support for song browser (only Pandora stations for now)
+  * Add streams to the stream list on the homepage in the order they were added
 * Streams
   * Add DLNA metadata and control support
   * Add support for browsing Pandora stations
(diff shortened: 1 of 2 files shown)


SteveMicroNova:
Before:
image

After:
image

rtertiaer (commented):
What Github Issue does this address?

I'm fine with it. I think you should seek other approval here too though, perhaps from @klay2000 .

SteveMicroNova:
> what github issue does this address?

None to my knowledge, it was a change added in #703 that I added to fix a nitpick of mine, as it was a whole 5 characters of change I've brought it forward into a separate PR. I found it slightly awkward feeling for new streams to be added by a button at the bottom but then get placed on the top, that's all this addresses

klay2000 (approved):
As previously discussed, this shouldn't break the stream ordering.

PR #715, opened 2024-05-10
Manual QR Code Update

SteveMicroNova:
Update QR Codes in manual to have branded features, add new QR codes for the upcoming video

### Checklist


SteveMicroNova:
Requesting review from Jason because he has opinions on the manual branding, requesting review from Lincoln and Ryan because they both have strong opinions about images in git

SteveMicroNova:
> Requesting review from Jason because he has opinions on the manual branding, requesting review from Lincoln and Ryan because they both have strong opinions about images in git

Rerequesting the same three for the same reasons now that the QRs have been updated once more

rtertiaer (changes requested):

rtertiaer (commented):
is there a higher quality AmpliPi logo you can use?

Is there a compelling reason why we should include the imaging instructions? even in hackable DIY products I've purchased, that's not something prominently featured in the in-box documentation (and if it were, I would not have the highest expectation of quality.)

rtertiaer (commented):
I like this rendition. You've asked for other feedback too, so I'll leave it to those folks to click "approve"

linknum23:
How are we creating the pdf? The `script/typeset_manual` is broken

SteveMicroNova:
> How are we creating the pdf? The `script/typeset_manual` is broken

Is it? I just right click docker-compose.yaml with docker desktop open and hit "Compose up" and I get a PDF

linknum23:
Looks like a couple of things are wrong with my setup. Give me a little bit to get back up and running. 

While I'm having errors it looks like there is still one generated file that needs to be added to .gitignore and removed:
- `docs/manual/manual.fls`

SteveMicroNova:
> Looks like a couple of things are wrong with my setup. Give me a little bit to get back up and running.
> 
> While I'm having errors it looks like there is still one generated file that needs to be added to .gitignore and removed:
> 
> * `docs/manual/manual.fls`

https://github.com/micro-nova/AmpliPi/blob/ManualQRUpdate/.gitignore
It's actually in there, at the bottom. I've having issues with some files ignoring their presence on the gitignore, am I maybe doing something obviously wrong for them to be continuously uploaded regardless?

linknum23:
You have to delete the file with ``git rm docs/manual/manual.fls`. After that git will respect the ignore file.

linknum23:
It looks like the video link is redirected to amplipi's GitHub. I assume that's intended?

linknum23:
Here's my build of the latest PDF. 
[manual.pdf](https://github.com/micro-nova/AmpliPi/files/15326825/manual.pdf)

linknum23 (approved):
Looks good to me. I'll have to add a pr to fix scripts/typeset_manual

rtertiaer (approved):

gorski123 (approved):
I like the QR codes, The quick start video QR code appears to be broken but I'm guessing this is more a Ryan / Andrew question.

SteveMicroNova:
> I'm guessing this is more a Ryan / Andrew question.

It's a Steve question, I made the cloud flare reroute, generated the QR codes, and added the image they've got on them so I've got it on lock
You'll have to define "doesn't work" for me however; does it not scan, or does it not go where expected?
I'll hop on this tomorrow

rtertiaer:
@gorski123 do we have a quickstart video to point to? right now it redirects me to the manual but that's intentional.

PR #716, opened 2024-05-10
Password Security

SteveMicroNova:
Minor security improvement
The password is still plaintext in the backend, but it is now not copyable from the textbox and the textbox converts it into dots as password fields should

image

### Checklist

### DIFF
--- a/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
+++ b/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
@@ -14,13 +14,14 @@ const RESTART_DESC = "Sometimes the stream gets into a bad state and neds to be 
 
 // We're already using mui, why are we reinventing the wheel? https://mui.com/material-ui/react-text-field/
 // if it's a matter of className control on the underlying components, that still works with the mui textfield with the InputLabelProps prop and other componentProps
-const TextField = ({ name, desc, defaultValue, onChange }) => {
+const TextField = ({ name, desc, type="text", defaultValue, onChange }) => {
+
     return (
         <>
             <div className="stream-field">
                 <div className="stream-field-name">{name}</div>
                 <input
-                    type="text"
+                    type={type}
                     defaultValue={defaultValue}
                     onChange={(e) => {
                         onChange(e.target.value);
@@ -35,6 +36,7 @@ const TextField = ({ name, desc, defaultValue, onChange }) => {
 TextField.propTypes = {
     name: PropTypes.string.isRequired,
     desc: PropTypes.string.isRequired,
+    type: PropTypes.string,
     defaultValue: PropTypes.string,
     onChange: PropTypes.func.isRequired,
 };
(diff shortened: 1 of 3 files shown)


web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -6,20 +6,21 @@ import Alert from "@mui/material/Alert";
...
 import StreamTemplates from "../StreamTemplates.json";
 import ModalCard from "@/components/ModalCard/ModalCard";
+import { none } from "ramda";
 
 const NAME_DESC =
   "This name can be anything - it will be used to select this stream from the source selection dropdown";
 const DISABLED_DESC = "Don't show this stream in the input dropdown";
 
 // We're already using mui, why are we reinventing the wheel? https://mui.com/material-ui/react-text-field/
 // if it's a matter of className control on the underlying components, that still works with the mui textfield with the InputLabelProps prop and other componentProps
```
SteveMicroNova, line 16: https://mui.com/material-ui/react-text-field/
I forgot about us recreating the TextField manually
I still hate it, but this time around I'm gonna make an issue out of that and solve it myself

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -190,18 +192,35 @@ const StreamModal = ({ stream, onClose, apply, del }) => {
...
-                                    }}
-                                />
-                            );
+                            if(field.name == "password"){
+                                return (
+                                    <TextField
+                                        key={field.name}
+                                        name={field.name}
+                                        desc={field.desc}
+                                        type="password"
```
rtertiaer, line 201: can you just toggle this field on & off instead of duplicating this many lines of code and having a conditional return value?

rtertiaer (changes requested):
why just pandora?

SteveMicroNova:
> why just pandora?

Not just pandora, pandora is just the only one I know of that has a password and so I named the branch after that, it still works with all parts labeled "password" on the stream create and edit modal

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -190,11 +191,16 @@ const StreamModal = ({ stream, onClose, apply, del }) => {
                     streamTemplate.fields.map((field) => {
                         switch (field.type) {
                         case "text":
+                            let type = "text";
+                            if(field.name == "password"){
+                                type = "password";
```
rtertiaer, line 196: i think i'd probably prefer to change the `type` of the password fields, instead of behaving magically with specifically named `text` fields. this also creates a new component - it's perhaps somewhat more LOC, but more maintainable, and you don't have to change the interface of some underlying component.

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -34,6 +34,7 @@ const TextField = ({ name, desc, defaultValue, onChange }) => {
 TextField.propTypes = {
     name: PropTypes.string.isRequired,
     desc: PropTypes.string.isRequired,
+    type: PropTypes.string.isRequired,
```
rtertiaer, line 37: either this is incorrect, or you have a bug in the other text field

rtertiaer (changes requested):

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -228,6 +217,20 @@ const StreamModal = ({ stream, onClose, apply, del }) => {
                                     }}
                                 />
                             );
+                        default:
```
rtertiaer, line 220: I don't think I like falling through to a `default` like this. We should handle two types - text, and password.

rtertiaer (changes requested):

rtertiaer (approved):
thanks~ pardon it was so much back & forth

PR #719, opened 2024-05-15
Eink status

mjustian:
This adds two new pieces of text to the e-ink display: Serial Number and Status. The serial number is taken off of the EEPROM. The status defaults to showing whether the AmpliPi is playing or not, but a custom status can be sent using a function as well. A numerical status is interpreted as an error code.

### Checklist


rtertiaer (approved):
I think my requests are entirely nitpicks, besides the early-return `if` statement stuff I'd like to see change into an `elif` structure in `ctrl.py` as we talked about in person. Also tbf my eyes glaze over a little bit with the eeprom changes, but they looked fine; I also tend to trust these changes if @Lohrer had a hand in it.

amplipi/display/einkdisplay.py
```diff
@@ -15,15 +20,25 @@
 
 from amplipi.display import epd2in13_V3
 from amplipi.display.common import DefaultPass, Display
+from amplipi.display.einkinterface import set_custom_eink_status
+from amplipi.display.einkinterface import status_filename
+from amplipi import models
+from amplipi import rt
 
-SysInfo = namedtuple('SysInfo', ['hostname', 'password', 'ip'])
+import requests
```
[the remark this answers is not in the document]
mjustian: Moved under the logger line

amplipi/display/einkdisplay.py
```diff
@@ -132,13 +156,36 @@ def update_display(self, info: SysInfo = SysInfo(None, None, None), draw_base=Fa
...
+        draw.text((0, start + 3 * interval), f'Status:   {info.status_code}', font=self.font, fill=0)
+        if info.status_code != startup_msg and self._boot:
+          self._boot = False
+          draw_base = True
+      else:
+        draw.text((0, start + 3 * interval), f'ERROR! - Code: {info.status_code}', font=self.font, fill=0)
+
+      draw.text((0, start + 4 * interval), f'Pass: {info.password}', font=self.pass_font, fill=0)
+      # if info.password:
+      #  draw.text((0, start + (password_level+1) * interval), info.password, font=self.pass_font, fill=0)
```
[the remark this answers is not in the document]
mjustian: Removed the commented section and added an if info.password before the actual password

amplipi/display/einkdisplay.py
```diff
@@ -150,17 +197,157 @@ def update_display(self, info: SysInfo = SysInfo(None, None, None), draw_base=Fa
...
+      set_custom_eink_status(17, None)
+      return 0
+  except Exception as e:
+    log.exception(f'Exception getting number of expanders: {e}')
+    set_custom_eink_status(16, None)
+    return 0
+
+
+def get_status(url: str) -> Tuple[Union[str, int], Optional[int], int]:
+  """Get the status from the file if there is one, otherwise base it off of API state"""
```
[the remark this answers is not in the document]
mjustian: Changed to:
"""Returns the system's status code as either a str or int, a serial number, and how many expanders are hooked up"""

amplipi/display/einkinterface.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
+"""E-Ink Display Interface"""
+
+from loguru import logger as log
```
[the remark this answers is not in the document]
mjustian: The entire E-Ink project uses Loguru, so it would be outside the scope of this task to switch it over

Lohrer:
Before I review this, can you confirm that:

- [ ] The TFT display on older units still functions properly.
- [ ] An AmpliPro/Streamer with an unprogrammed EEPROM (all 1's) shows an error
- [ ] The Streamer EEPROM can still be programmed with `hw/tests/config_streamer.bash`
- [ ] Tests for any API changes are added (I believe `tests/test_rest.py` is the correct spot)

amplipi/display/einkinterface.py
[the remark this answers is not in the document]
linknum23: While you are in this you may want to fast follow with a PR that updates the logging here.

amplipi/ctrl.py
```diff
@@ -541,9 +547,35 @@ def _sync_stream_info(self) -> None:
...
+        if bi is not None:
+          self._serial = bi.serial
+      else:  # -1 means there is no serial number available from any EEPROM
+        self._serial = -1
+      self.status.info.serial = str(self._serial)
+
   def get_state(self) -> models.Status:
     """ get the system state """
     self._update_sys_info()
+    self._update_serial()
```
linknum23, line 578: This looks like it is called every get(). I realize that internally the serial number is cached so that isn't actually what happens.  Why isn't this called in reinit instead? That is where we read a similar property set, the firmware versions.

mjustian: It didn't work in reinit since other things were trying to reach the EEPROM at the same time

Lohrer: That's not true, it didn't work in reinit at first because you put it before the preamp's micro was assigned an I2C address, so it couldn't communicate. We didn't realize that until after you had already moved the serial reading to here. There's no reason you couldn't move it back, it just needs to be after the line that calls `rt.Rpi()` to setup the micro/firmware.

amplipi/eeprom.py
```diff
@@ -76,166 +131,123 @@ class UnsupportedFormatError(Exception):
...
-        raise EEPROMWriteProtectError("Write enable failed")
+    # try:
+    #   val = self._i2c.read_i2c_block_data(self._address, WRITE_CHECK_ADDRESS, 1)[0]
+    #   self._i2c.write_i2c_block_data(self._address, WRITE_CHECK_ADDRESS, [val + 1])
+    #   sleep(0.1)
+    #   check = self._i2c.read_i2c_block_data(self._address, WRITE_CHECK_ADDRESS, 1)[0]
+    #   self._i2c.write_i2c_block_data(self._address, WRITE_CHECK_ADDRESS, [val])
+    #   sleep(0.1)
+    #   if check != val + 1:
+    #     raise EEPROMWriteProtectError("Write enable failed")
```
linknum23, line 176: Whats the plan here?

amplipi/eeprom.py
```diff
@@ -76,166 +131,123 @@ class UnsupportedFormatError(Exception):
...
     except OSError as exception:
       raise EEPROMWriteError("Write failed") from exception
     sleep(0.1)
 
-  def _read(self, address: int, length: int) -> List[int]:
+  def _read(self, address: int, length: int, unit_addr=0x08) -> bytes:
     """Read data from address in EEPROM."""
     try:
-      return self._i2c.read_i2c_block_data(self._address, address, length)
+      if self._bus() == 1:
```
linknum23, line 229: Should we add an enumeration for the different i2c bus #'s?

amplipi/eeprom.py
```diff
@@ -76,166 +131,123 @@ class UnsupportedFormatError(Exception):
...
-            self._read_letter(BOARD_REV_ADDR + 1))
-
-  def get_board_info(self) -> BoardInfo:
-    """Get board info from EEPROM."""
-    return BoardInfo(self.get_serial(),
-                     self.get_unit_type(),
-                     self.get_board_type(),
-                     self.get_board_rev())
-
+  # TODO: Do we need this? If so implement like the _read() function above. If not remove.
```
linknum23: If we remove this, it probably makes sense to keep this pretty comment somewhere

amplipi/display/einkdisplay.py
```diff
@@ -150,17 +196,159 @@ def update_display(self, info: SysInfo = SysInfo(None, None, None), draw_base=Fa
...
+    update_logs(st)
+    result_status = st
+
+  # Check if API is running
+  api_on = os.system("systemctl --user is-active amplipi.service")
+  if api_on != 0:
+    return 0, None, 0
+
+  if url is None:
+    return 10, None, 0
```
Lohrer, line 291: All of these exception codes should be an enum. For an example see `eeprom.py`'s `class UnitType(IntEnum)`.

docs/manual/EINK_ERROR_CODES.md
```diff
@@ -0,0 +1,14 @@
+# AmpliPro E-Ink Display Error Codes
+
+The following numerical codes are displayed on the AmpliPro's E-Ink Display when the corresponding error occurs. Note that these codes do not appear on older AmpliPi/AmpliPro devices that use TFT touchscreen displays.
```
Lohrer, line 3: Is there a reason we're not supporting showing error codes on older units? Most of the units in the field have TFT displays at this point.

hw/tests/config_streamer.py
```diff
@@ -8,15 +9,15 @@
 
 # Add the amplipi directory to PATH
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../..')))
-from amplipi.eeprom import BoardInfo, BoardType, EEPROM, UnitType
 
-BOARD_REV_RE = re.compile(r"(\d+)([A-z])") # accept lower or uppser case change designator
-                                           # we will uppercase later
+BOARD_REV_RE = re.compile(r"(\d+)([A-z])")  # accept lower or uppser case change designator
```
Lohrer, line 13:
```suggestion
BOARD_REV_RE = re.compile(r"(\d+)([A-z])")  # accept lower or upper case change designator
```
While we're changing this anyway...

hw/tests/config_streamer.py
```diff
@@ -36,8 +37,14 @@ def main() -> None:
 
   try:
     print(f"Writing streamer config to eeprom: serial={serial}, board_rev={board_rev[0]}{board_rev[1]}")
-    eeprom = EEPROM(0, BoardType.STREAMER_SUPPORT)
-    eeprom.write_board_info(BoardInfo(serial=serial, unit_type=UnitType.STREAMER, board_type=BoardType.STREAMER_SUPPORT, board_rev=board_rev))
+    eeprom = EEPROM(BoardType.STREAMER_SUPPORT)
+
+    br = BoardRev(board_rev[0], board_rev[1])
+    if serial == -1:
+      eeprom.deprogram_board()
```
Lohrer, line 44: How about just calling this function `erase()`?

amplipi/ctrl.py
mjustian: I tried moving it back to reinit where the firmware is and it stopped working on the regular sized amplipi

amplipi/eeprom.py
```diff
@@ -76,166 +140,128 @@ class UnsupportedFormatError(Exception):
...
-      except OSError:
-        pass
-
-    return devices
+  def _address(self) -> int:
+    """Returns the I2C address of the EEPROM, based on the BoardType"""
+    # TODO: make board a full class to abstract this away.
+    return int(self._board) & 0x7F  # Lowest 7 bits are the I2C address
 
+  # TODO: Is this necessary?
```
Lohrer:
```suggestion
```

amplipi/eeprom.py
Lohrer: These are old, remove them

linknum23:
As discussed: In the future we may want to add a slim barcode that can be scanned to quickly link to our troubleshooting guide when there is an error.

amplipi/eeprom.py
```diff
@@ -76,166 +140,128 @@ class UnsupportedFormatError(Exception):
...
   def _enable_write(self) -> None:
-    """Enable write to EEPROM."""
+    """Enable write to EEPROM. Only required for streamers."""
     try:
+      GPIO.setmode(GPIO.BCM)
+      GPIO.setup(WP_PIN, GPIO.OUT)
       GPIO.output(WP_PIN, GPIO.LOW)
     except NameError:
       pass
```
Lohrer, line 165: Since this should only ever be done on the Streamers, how about
```suggestion
  if self._board == BoardType.STREAMER_SUPPORT:
      try:
        GPIO.setmode(GPIO.BCM)
        GPIO.setup(WP_PIN, GPIO.OUT)
        GPIO.output(WP_PIN, GPIO.LOW)
      except NameError:
        pass
```
And then you can remove that check in the `write_board_info()` function.

amplipi/display/einkdisplay.py
```diff
@@ -1,29 +1,43 @@
...
 from amplipi.display import epd2in13_V3
 from amplipi.display.common import DefaultPass, Display
+from amplipi.display.einkinterface import set_custom_eink_status
+from amplipi.display.einkinterface import status_filename
+from amplipi import models
+from amplipi import rt
 
-SysInfo = namedtuple('SysInfo', ['hostname', 'password', 'ip'])
+
+SysInfo = namedtuple('SysInfo', ['hostname', 'password', 'ip', 'status_code', "serial_number", "ext_count"])
```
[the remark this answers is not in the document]
mjustian: mypy does not like this for whatever reason

amplipi/ctrl.py
```diff
@@ -541,9 +547,39 @@ def _sync_stream_info(self) -> None:
...
+  def _update_serial(self) -> None:
+    expanders = []
+    # Preamp Boards
+    eeprom_preamp = EEPROM(BoardType.PREAMP)
+    eeprom_streamer = EEPROM(BoardType.STREAMER_SUPPORT)
+    if eeprom_preamp.present():
+      bi = eeprom_preamp.read_board_info()
+      if bi is not None:
+        self._serial = int(bi.serial)
+        for board in [0x10, 0x18, 0x20, 0x28, 0x30]:
```
Lohrer, line 559: I would recommend
```suggestion
        for board in rt._DEV_ADDRS[1:]: # Check for expanders
```
Although maybe eeprom.py should be handling the unit number -> I2C address conversion...

And either way I guess _DEV_ADDRS shouldn't be marked as private anymore, or moved to a getter.

amplipi/display/einkdisplay.py
```diff
@@ -1,29 +1,43 @@
...
 from PIL import Image, ImageDraw, ImageFont
 from loguru import logger as log
+import requests
 
 from amplipi.display import epd2in13_V3
 from amplipi.display.common import DefaultPass, Display
+from amplipi.display.einkinterface import set_custom_eink_status
+from amplipi.display.einkinterface import status_filename
+from amplipi import models
+from amplipi import rt
```
Lohrer, line 26:
```suggestion
from amplipi.display.einkinterface import set_custom_eink_status, status_filename
from amplipi import models
```

amplipi/display/einkdisplay.py
```diff
@@ -73,20 +89,27 @@ def init(self) -> bool:
...
+
   def run(self):
     self._ok = True
     try:
       default_pass = DefaultPass()
-      info = SysInfo(None, None, None)
+      info = SysInfo(None, None, None, None, None, None)
 
       display_change_counter = self.refresh_interval
+      # set_custom_eink_status("test!", expiration=(datetime.datetime.now() + datetime.timedelta(seconds=30)))
```
Lohrer, line 108: I assume this can be removed?

amplipi/display/einkdisplay.py
```diff
@@ -1,29 +1,43 @@
...
 
-SysInfo = namedtuple('SysInfo', ['hostname', 'password', 'ip'])
+
+SysInfo = namedtuple('SysInfo', ['hostname', 'password', 'ip', 'status_code', "serial_number", "ext_count"])
+
+startup_msg = "Starting Up"
 
 
 class EInkDisplay(Display):
   """ Display system infomation on EInk Panel"""
```
Lohrer, line 35: While we're in here...
```suggestion
  """ Display system information on EInk Panel"""
```

amplipi/eeprom.py
```diff
@@ -3,58 +3,122 @@
 These are used to identify features and device types among other things."""
 
 from dataclasses import dataclass
-from enum import Enum
+from enum import IntEnum
+from struct import pack, unpack, calcsize
 from time import sleep
-from typing import List, Tuple
-import smbus2 as smbus
+from typing import List, Optional, Union, Tuple
```
Lohrer, line 9: Is your setup showing you unused imports?
```suggestion
from typing import List, Optional, Union
```

amplipi/display/einkinterface.py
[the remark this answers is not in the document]
mjustian: If another process has an error code and wants to set a status to the E-Ink display, it will need this function to do so.

docs/manual/EINK_ERROR_CODES.md
mjustian: As of this pull request there still isn't a field showing error codes on TFT displays. That sounds like it is outside of the scope of this pull request, but it would make sense to add at some point.

Lohrer: I discussed with Ryan and we will keep it out of this PR but it will be the next task.

amplipi/display/einkdisplay.py
```diff
@@ -150,17 +194,159 @@ def update_display(self, info: SysInfo = SysInfo(None, None, None), draw_base=Fa
...
-        if font.getlength(password) <= max_length:
+        if font.getlength("Pass: " + password) <= max_length:
           return font
     except Exception as exc:
       raise Exception(f'Failed to load {self.fontname} font') from exc
     return ImageFont.truetype(self.fontname, 10)
 
 
-def get_info(iface, default_pass) -> SysInfo:
+def update_logs(status) -> None:
```
Lohrer, line 209: What is the purpose of this? It looks like you're reinventing logging here. All the rest of our logs are captured by `journald` (it's just capturing `stdout` and `stderr`, nothing fancy) and are viewable with `journalctl -f` or by going to Settings->About->Logs in the AmpliPi web UI which pulls from the same journal service.

Also I tested it and the only log message I'm seeing in `display-logs.json` is "Starting Up". I tried changing the status and no other statuses show in that log file.

rtertiaer: This was one possible path I mentioned on day 2 as a persistent store of statuses to display. It may be vestigial now, particularly if we don't have external things reading or writing status, to this file or elsewhere - everything is simply represented as a state from the API.

rtertiaer:
One ask here would be to surface these same statuses on the TFT display; you can borrow the unit on my desk for this if you'd like.

amplipi/display/einkdisplay.py
mjustian: It would make a lot more sense to move the log file's information to the actual logs. I'm making that change now.

Lohrer (approved):
Much better! One last request would be to add `display-status.txt` to the `.gitignore` file so that git doesn't try to track it.

I think it would be awesome in the future to define these error codes as a {Short Name, Description} instead of just a number. That way we can display the short name on the display and the description in the logs and maybe webUI?

rtertiaer (approved):
two nitpicks, but otherwise heck yes :sunglasses:

hw/tests/config_streamer.py
```diff
@@ -8,15 +9,15 @@
 
 # Add the amplipi directory to PATH
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../..')))
-from amplipi.eeprom import BoardInfo, BoardType, EEPROM, UnitType
 
-BOARD_REV_RE = re.compile(r"(\d+)([A-z])") # accept lower or uppser case change designator
-                                           # we will uppercase later
+BOARD_REV_RE = re.compile(r"(\d+)([A-z])")  # accept lower or upper case change designator
+# we will uppercase later
```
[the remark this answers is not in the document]
Lohrer: Out of curiosity, what do you feel this is obscuring? I suggested this change in the first place so I'd like to know for my own future reference. I felt that since we were already making changes in this area of the file, since the auto-formatter added a space on this line already, and since in the commit you can clearly see it's just a change in the comment, that it was worth fixing the typo now.

Issue #723, opened 2024-05-16
Support for Power Board Rev 6


Lohrer:
The latest revision of the power board slightly improves the filtering on the supplies, commonizes parts for easier production, and adds more "Power Good" signals available for reading by the Preamp's microcontroller.

We need to update the microcontroller's firmware to pass these new signals up to the Pi.

PR #724, opened 2024-05-16
Support for Power Board Rev 6

Lohrer:

Complete #723

### Checklist

* [x] Have you updated preamp_i2c_regs.md?
* [x] Gave you updated the fw/preamp/CHANGELOG?


.vscode/settings.json
```diff
@@ -11,6 +11,11 @@
     "editor.rulers": [100]
   },
   "C_Cpp.default.cStandard": "c23",
+  "C_Cpp.default.defines": [
+    "false=0",  // Defining true and false here is required since the C/C++ extension's
+    "true=1",   // intellisense does not support this feature of c23 yet.
```
linknum23, line 16: This seems risky. I thought true was !false

linknum23: I guess this is not for compilation so it probably doesn't matter.

Lohrer: Exactly, I just got annoyed about the red squiggles. I am updating the comments around these to better indicate their temporary nature and clarify that they aren't used for compilation.

As for true=!false, I assume that would work here but from the C++ spec:
> A prvalue of type bool can be converted to a prvalue of type int, with false becoming zero and true becoming one.

So I figured defining true=1 was the easiest.

Lohrer:
Gonna have to make some changes here, PG_9V is inconsistent depending on the hardware setup. The hardware revision with Power Board 3.B is reporting a bad PG_9V, which makes sense since that signal doesn't exist and the current firmware implementation assumes all older hardware has that signal.

Lohrer:
This appears to work on amplipi256 (preamp v4, power v6), amplipi197 (preamp v3, power v4), and lamplipi (preamp v3, power v3). I still need to test on the original units (preamp v2 and power v2) and the intermediate combo of preamp v4 and power v4.

linknum23 (approved):

PR #726, opened 2024-05-17
Validate stream name

npunniyamoorthy:
Validate fields when streams are created and reconfigured for AirPlay, Spotify, Pandora, LMS, and Internet Radio

### Checklist


npunniyamoorthy:
Need to: 
-have validate_stream take in a dictionary instead of multiple arguments. 
-Create a type of exception for streams so we can have consistent error messages

amplipi/streams.py
```diff
@@ -86,6 +86,12 @@ def uuid_gen():
   # Generic UUID in case of failure
   return '39ae35cc-b4c1-444d-b13a-294898d771fa'
 
+class StreamError(Exception):
```
linknum23, line 89: Would this work better as an InvalidField exception? That way the code could always print the message with the same formatting.

amplipi/ctrl.py
```diff
@@ -984,8 +988,9 @@ def set_stream(self, sid: int, update: models.StreamUpdate) -> ApiResponse:
       stream.reconfig(**changes)
       self._sync_stream_info()
       return ApiResponse.ok()
-    except Exception as exc:
```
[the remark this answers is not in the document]
npunniyamoorthy: Probably not safe to assume so

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -133,19 +133,6 @@ InternetRadioSearch.propTypes = {
     onChange: PropTypes.func.isRequired,
 };
 
-function validateInput(streamFields, streamTemplate) {
-    streamTemplate['fields'].filter((f) => f.required).map( (f) => {
```
linknum23, line 137: Does this move all required field validation to the api?

npunniyamoorthy: yes

linknum23:
As far as I can tell all of the previous review comments have not been addressed. 

In addition since this is an API change we should probably update the stream tests in tests/test_rest.py:
- [ ] Make test_patch_stream_rename to check for success or failure depending on a list of good or bad names. 
- [ ] Make test_create_pandora check for success or failure based on good/bad names
- [ ] Add a similar test_create_X for the other stream names being validated

amplipi/ctrl.py
```diff
@@ -83,6 +88,11 @@ def error(msg: str):
     """ create an error response """
     return ApiResponse(ApiCode.ERROR, msg)
 
+  @staticmethod
```
linknum23, line 91: :heart:

docs/amplipi_api.yaml
```diff
@@ -1,12047 +0,0 @@
-openapi: 3.0.2
```
linknum23, line 1: Why was this deleted?

amplipi/streams.py
```diff
@@ -1167,6 +1200,14 @@ def send_cmd(self, cmd):
     except Exception:
       pass
 
+  def validate_stream(self, **kwargs):
+    url_regex = r'^https?://[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)$'
+    logo_regex = r'^https?://[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)\.(?:jpg|gif|png)$'
```
linknum23, line 1205: We probably don't need to specify the type of image here. Web browsers support a wide variety of image formats and I assume that they don't require an extension.

amplipi/streams.py
```diff
@@ -685,6 +704,11 @@ def send_cmd(self, cmd):
     except Exception as e:
       raise Exception(f"Error sending command {cmd}: {e}") from e
 
+  def validate_stream(self, **kwargs):
+    regex = r"^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$"
```
linknum23, line 708: I'm having a hard time digesting this regex. This is the one that detects a hostname-like name right?

amplipi/streams.py
linknum23: I assume both fields could just use url_regex right?

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -151,33 +155,31 @@ const InternetRadioSearch = ({ onChange }) => {
...
     ); // set streamFields to copy of stream
 
     const [errorMessage, setErrorMessage] = React.useState("");
 
+    const [errorField, setErrorField] = React.useState("");
+
     const streamTemplate = StreamTemplates.filter(
         (t) => t.type === stream.type
     )[0];
+    // HACK slight hack to ensure name is always present
```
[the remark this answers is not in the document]
linknum23: This is because the API  Stream's base class requires a name right?

rtertiaer (changes requested):

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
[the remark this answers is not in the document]
klay2000: kinda, I'm gonna clean this up, it's a little overengineered

amplipi/streams.py
linknum23: once this regex is cleaned up a potential better name for the `regex` variable might be `HOSTNAME_LIKE`

rtertiaer (approved):
:+1:

amplipi/ctrl.py
```diff
@@ -69,9 +70,13 @@ class ApiCode(Enum):
...
   """ Ctrl Api Response object """
 
-  def __init__(self, code: ApiCode, msg: str = ''):
+  def __init__(self, code: ApiCode, msg: str = '', data: Optional[Dict] = None):
     self.code = code
     self.msg = msg
+    if data:
+      self.data = data
+    else:
+      self.data = {}
```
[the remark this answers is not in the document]
linknum23: I prefer the original to avoid pitfalls with mutable default arguments. Your suggestion should produce a linter warning.

amplipi/streams.py
```diff
@@ -144,7 +156,7 @@ def restart(self):
       self.send_cmd('stop')
     except:
       logger.info(f'Stream {self.name} does not have a stop response')
-    last_src = self.src # Disconnect sets self.src to none, so temp variable used to keep track
+    last_src = self.src  # Disconnect sets self.src to none, so temp variable used to keep track
```
linknum23, line 159: Whats going on with the formatter here? I'm seeing a fair amount of lines that are just spacing changes.

linknum23: Looks like you are using vscode's autoformatting. We should probably do this to our entire codebase. PR made for this here: https://github.com/micro-nova/AmpliPi/pull/773

amplipi/ctrl.py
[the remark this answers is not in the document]
rtertiaer: TIL. surprised this is the first time I've seen this :woozy_face:

PR #728, opened 2024-05-20
put QR code to eink display at end of mfg

### LINKED ISSUE #718: Make quickstart QR code valid; put to e-ink display during manufacturing

rtertiaer:
This PR puts the QR code to the e-ink display at the very end of manufacturing. This has to happen here, because earlier than this it will be overwritten, and at the end of this script the appliance is shut down and boxed up.

This closes #718 .

### Checklist


linknum23 (approved):
Side Note: the cleanup script needs to be run with an ip address in our normal network since the hostname changes to amplipi.

PR #730, opened 2024-05-21
Tests: Limit the amount of requests sent to pandora api

linknum23:

This limits the amount of requests sent to the Pandora api to 1/5. It also adds some delays to further reduce the request intensity. 

### Checklist

### DIFF
--- a/tests/test_rest.py
+++ b/tests/test_rest.py
@@ -14,6 +14,10 @@ import tempfile
 import os
 from copy import deepcopy # copy test config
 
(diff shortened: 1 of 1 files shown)


linknum23:
NOTE: we may need to limit these requests further by hiding them behind a command line argument. That would be an extra step that may lead to us missing test failures. Lets try this for now and see what kind of mileage we can get out of it.

rtertiaer (approved):
it's gentle, simple... seems good for now.

PR #732, opened 2024-05-22
Make preamp tests find aptester

rtertiaer:

For some reason aptester commands need a bigger timeout. This puts a bandaid on it without fixing the underlying problem.

### Checklist

* [x] Have you tested your changes and ensured they work? (Lincoln did)


linknum23 (approved):
:1st_place_medal:

PR #734, opened 2024-05-29
Refresh button

mjustian:
There was no easy way to reset a broken stream before. This adds a button to the player view and the modal in settings to do that. The buttons interface with a stream command on the REST API that resets a stream. IT supports both BaseStream and PersistentStream.

### Checklist

* [X] If this is a UI change, have you tested across multiple viewport sizes (ie. desktop versus mobile)?

Web View:
reset0

iPad:
reset_ipad

iPhone 12 Mini:
reset_iphone12mini

iPhone SE
reset_iphonese

Stream Modal:
reset_modal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   * Add streams to the stream list on the homepage in the order they were added
   * Add firmware version for main and expansion units on About page
   * Limit length of displayed stream names with ellipsis
+  * Add restart stream button to stream player and stream modal
 * System
   * Add serial number to eink display
   * Add ability to display status on eink display
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -87,6 +87,7 @@ logger.setLevel(logging.DEBUG)
 sh = logging.StreamHandler(sys.stdout)
 logger.addHandler(sh)
 
+
 class SimplifyingRouter(APIRouter):
   """
   Overrides the route decorator logic to:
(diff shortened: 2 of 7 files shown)


rtertiaer (changes requested):
if you absolutely must wholesale run files thru a linter, can you at least put those in separate commits so I can easily review what's functionally changed? (not a huge fan of that anyhow, unless you're substantially rewriting a file anyhow.)

linknum23:
Opinions below. I'll let others chime in with their thoughts and see if we can reach a consensus.

- I really like being able to reset the stream from the stream modal since it takes a couple of clicks to get there and has an explanation as to what it does and why.
- My instinct is that the actual reset button on the player is too easy and takes up too much space for an operation that shouldn't be needed under most scenarios.
- After talking with Michael on this I like the word restart for this instead of reset. I think it is a less ambiguous name and clearly explains what it does in one work.

gorski123:
The restart button should be shown by the currently playing stream. I dont think we need to write out reset or restart, rather just a button with a logo (reset logo or the current recycle logo you are using) should suffice. I was thinking something like this:
image
image

linknum23:
I don't think the main view with all of the streams needs a reset button.
It doesn't even have the much more used play/pause/next song button.

mjustian:
If the main view had a button like that, it would be really hard to tap on a mobile device.

mjustian:
Also, would the end user know the recycle symbol would mean "restart stream"? Next to the icon it isn't really clear that it's a button.

gorski123:
Hows about just putting it next to all the other buttons (play pause etc)? Thats prolly where it belongs...

mjustian:
That would make sense, but on views such as Pandora which has 5 buttons already, it would be a bit long, especially on a mobile device. I'm not 100% sure if that would be worth making the rest of the buttons much smaller to fit the reset button.

gorski123:
Can the buttons wrap to a second line if they dont fit? The pandora buttons already go off the screen in mobile, perhaps implementing some code to wrap to a second line could squash that bug as well.

mjustian:
I think it would look odd if the thumbs up and thumbs down buttons are on different levels. Also, putting it with the rest of the controls conflicts with the "having a restart button so easily accessible means it will break often" opinion. It might be better to make the controls smaller on mobile, rather than wrapping to a second line.

rtertiaer (approved):
one last request, but otherwise you have my approval. could you detail in a comment in the code why you need a `src = self.src` in `BaseStream`'s `restart()`, but not in `PersistentStream`?

rtertiaer:
oh, and rebasing + squashing this is very preferred too.

mjustian:
Reason for src = self.src in BaseStream:
the _disconnect function sets self.src to None, but it is also needed to run the connect function. The PersistentStream does not have any arguments when activating, so it's not needed.

PR #736, opened 2024-05-31
Added an expiration time to display status related to expanders

mjustian:
There was a bug on the display where some statuses would not clear when the underlying issue stopped, particularly when calculating the number of expanders. To fix that, I added an expiration time of 10 seconds.

### Checklist

*


rtertiaer (approved):
if you affirm that this current rendition is tested, looks good to me :sunglasses:

Issue #702, opened 2024-05-01
LMS stream choppy and distorted after not being used for a few days


sumnerboy12:
I have an LMS client stream configured on my AmpliPi, connected to my LMS server running elsewhere on my network. After a few days of no use (with nothing playing and no zones linked) when I try to play something on the LMS stream, the playback is distorted, crackly, and sounds terrible.

Deleting the input using the LMS client stream and recreating makes no difference, the playback is still terrible.

Deleting the input and stream and recreating both appears to resolve the issue.

rtertiaer:
I last touched this chunk of code [in Dec 2023](https://github.com/micro-nova/AmpliPi/commit/05f0f01f8f7a91ea4d2ce2fe24226e2a2833b6d4), part of our initial 0.3.0 release. To note - that's also the moment I introduced squeezelite as an actual .deb package, which may have updated the version we are running to one with a regression.

 @sumnerboy12 have you noticed this in other 0.3.x releases as well?

sumnerboy12:
I have implemented a nightly restart of all my streams (using NodeRED) and I have not noticed any issues since then. Sorry I can't provide more information but I am happy things are stable at the moment :).

rtertiaer:
Reproduced this in-house after running an LMS stream for 6 days. Some notes:
* there are not duplicate `squeezelite` clients
* Modifying AmpliPi controlled volume or mute does not change this.
* One `squeezelite`+LMS player combo was playing Beyonce for 6 days and had this issue. Killing the Beyonce playlist from within LMS and playing a Detroit jazz internet radio station actually cleared this for that player alone; `squeezelite` did not get restarted.

linknum23:
One solution to this problem may be to build the latest pianobar release from source. Currently the deb package is at least several years out of date.

rtertiaer:
I'm testing upgrades to these packages in #743 ; we'll see if they make a difference.

rtertiaer:
We've received some feedback that this issue has not yet reoccurred for an LMS user running 0.3.6, with the above LMS & squeezelite updates merged.

rtertiaer:
And this issue has re-occurred for this user.
Given those old suspicious squeezelite issues above, I am increasingly wondering if there's some underlying squeezelite/LMS dependency  issue. All those reports are many years old and have seemingly evaporated - nobody complains of this anymore, except our users. This may be solved long term by #471 + #472 + upgrading to a later distro (which unlocks #415). 

[...]

PR #737, opened 2024-06-03
clean up LMS client termination

rtertiaer:
This PR makes the LMS client termination code way cleaner and more straightforward. Prior, if the metaproc handling exceptioned, only the main proc would get fully killed. This change properly splits out that logic. 

This change was spurred by reexamining stuff related to #702 ; after running this branch for 5 days, I am unable to replicate that particular issue. It's hard to say if that means I didn't replicate it, or this fixed it. My gut feeling is that this isn't the fix for that issue and I'll keep trying to replicate it, but in the meantime this is still a valid fix.

### Checklist

### DIFF
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -1519,16 +1519,25 @@ class LMS(PersistentStream):
         src_config_folder = f'{utils.get_folder("config")}/srcs/v{self.vsrc}'
         os.system(f'rm -f {src_config_folder}')
         self.proc.terminate()
-        self.proc.communicate()
-        if self.meta_proc is not None:
-          self.meta_proc.terminate()
-          self.meta_proc.communicate()
-          self.meta_proc = None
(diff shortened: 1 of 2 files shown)


amplipi/streams.py
```diff
@@ -1498,16 +1498,25 @@ def _deactivate(self):
...
         self.proc.terminate()
-        self.proc.communicate()
+        self.proc.communicate(timeout=10)
+      except Exception as e:
+        logger.exception(f"failed to gracefully terminate LMS stream {self.name}: {e}")
+        logger.warning(f"forcefully killing LMS stream {self.name}")
+        os.killpg(self.proc.pid, signal.SIGKILL)
+        self.proc.communicate(timeout=3)
+
+      try:
```
linknum23, line 1508: Suggestion move the if above the try to show the logical flow better

linknum23 (approved):

PR #742, opened 2024-06-10
Make upgrades a bit more robust

rtertiaer:
This PR increases the service restart retrying period during upgrades. I've seen this fail a couple of times when it shouldn't have (the service eventually restarted, but wasn't up within the 5s/20s); I also know I've personally added some ['lengthy' hardware checks](https://github.com/micro-nova/AmpliPi/pull/646/files#diff-b099f52a8996d7f5e64a578e2d43f942e75b5c095abacae37278e5c1af448577R1236) over time that probably make this more necessary, in addition to our software simply becoming more complex.

I don't feel like our service increasing startup time from & to these values is massively problematic. We're working towards getting newer Python which should increase our startup speed a ton; this is a decent change in the interim to increase 'stability' of upgrades, without necessarily kicking some performance can down the road ceaselessly.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   * Add serial number to eink display
   * Add ability to display status on eink display
   * Better logging around failed upgrades
+  * Make upgrades more stable
 * Display
   * Add serial number
   * Add status code field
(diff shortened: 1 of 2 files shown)


linknum23 (approved):
😎

PR #745, opened 2024-06-13
expose stream types in SourceInfo for better icon handling

### LINKED ISSUE #713: Wrong Stream type icons
https://github.com/micro-nova/AmpliPi/pull/708#issuecomment-2099164493

As seen in the linked screenshot, icons can be wrong for some streams
For instance, that is a Pandora stream

rtertiaer:
This PR adds a new field to `SourceInfo` - `type`, used to instantiate with the `stream_type` directly from the class. This allows us to robustly handle stream icons in the frontend. I've set this as an optional field; that feels reasonably safe to me, largely because we actually handle the fall-through undefined or incorrectly set cases in the frontend already.

Before, we were simply parsing the type out of the stream name, and along the way we broke that convention multiple times.

I'd generate the new spec with this, but there's a circular import preventing that. I'll attempt to fix that in the next PR.

This PR fixes #713 .

the after pic: 
image

### Checklist


linknum23 (approved):
Looks good. Thanks for removing my jenky `-` based parsing that was propagated from the old web app.

PR #747, opened 2024-06-13
Player No-Zone obviation

### LINKED ISSUE #637: When a player has no zones attached the volume/mute should not be shown and the lack of zones should be more obvious

SteveMicroNova:

Remove volume and mute controls when there is nothing playing on the player card and player page

Aims to close #637 by removing anything that suggests audio could be playing

What is specifically achieved is removing the volume bar and mute button from both the player card on the home page and the player page when a source has no zones, as well as making the "Add Zones" badge on the player card flash slowly when there aren't any zones attached

### Checklist

### DIFF
--- a/web/src/components/Chip/Chip.scss
+++ b/web/src/components/Chip/Chip.scss
@@ -1,5 +1,11 @@
 @use "src/general";
 
+:root {
+  --start-shadow: 0px 0px 25px 0px rgba(0, 0, 0, 1);
+  --pulse-shadow: 0px 0px 25px 3px rgba(0, 149, 255, 1);
+}
+
+
 .chip {
   display: flex;
   align-items: center;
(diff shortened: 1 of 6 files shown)


SteveMicroNova:
https://github.com/micro-nova/AmpliPi/pull/747/commits/0dec1c61558b0c73ad509edadba3d0e0195fe44b
I do not like the amount of prop drilling I had to do to make this work, and I especially do not like making a new prop for a single usecase
we need to be able to override classNames and styles in parent elements in general, this comes up a lot

SteveMicroNova:
https://github.com/micro-nova/AmpliPi/assets/114415937/48df28e7-8e21-40e2-8371-4487aface936

See video for demonstration

linknum23:
Mind editing the checklist and selecting whats been done?

linknum23:
I didn't noticed the pulsing in the video until the 4th watch. I think this needs to be a little more distinctive. 
I also tested this with groove salad as a background and was unable to notice the blinking.

Would changing the background color instead of alpha help here?

linknum23:
Also I got lamplipi back up and running should be good to test on!

web/src/components/Chip/Chip.scss
```diff
@@ -9,3 +9,33 @@
   margin: 0.25rem;
   @include general.highlight-hover;
 }
+
+
+@keyframes shakeBackground {
+  // Do not edit the steps of this animation unless you intend to not use box-shadow
```
linknum23, line 15: You may want to update this comment with a message about why we want to shake the background. Something like "Get the users attention but don't be obnoxious about it. Used For things that aren't errors, but need some user attention like a source with no zones "

linknum23 (approved):
I like how simple this is! Nice work.

PR #749, opened 2024-06-14
Correct power output detail for the every channel used case

rtertiaer:
This PR attempts to add clarity around our all-channel peak output power (am I even saying that right?)

I'd like feedback on how this is represented; I will go through the process of stamping out a new rendered manual and hosting it upstream once the verbiage here is correct.

### Checklist

### DIFF
--- a/docs/manual/MAIN_CONTROLLER_SPECS.md
+++ b/docs/manual/MAIN_CONTROLLER_SPECS.md
@@ -23,6 +23,7 @@
 |||
 |Speaker Power     | 55 WPC @ 4 Ω |
 |                  | 32 WPC @ 8 Ω |
+|                  | 23 WPC all-channel driven power |
 |||
 |Speaker Impedance | 4 - 8 Ω |
 |||


docs/manual/MAIN_CONTROLLER_SPECS.md
```diff
@@ -23,6 +23,7 @@
 |||
 |Speaker Power     | 55 WPC @ 4 Ω |
 |                  | 32 WPC @ 8 Ω |
+|                  | 23 WPC all-channel driven power |
```
Lohrer: I'm not sure how you want to format this in here, but these measurements are for **<0.1% THD+N** and taken at **1kHz**. Both of those are important because they strongly influence the result. For example the much higher 10%THD (note the lack of +N which also makes this weaker) of the amp chip spec results in a _much_ higher output wattage, but when tested I had to clip the signal to get 10% THD.

rtertiaer: I'm putting this in front of you largely cuz I have no idea how to properly represent that data; this is not my area of expertise.

Could you provide a recommendation on how you would represent it?

linknum23:
@Lohrer where is this at? Should we just try to merge this as-is?

PR #750, opened 2024-06-14
Removed stop option from Pandora

mjustian:
Sending a stop command to Pandora closes Pianobar, requiring a restart. This has no benefits and would cause problems to those using alternative user interfaces, which would consider Pandora on AmpliPi to have a stop command and make it visible.

### Checklist

* [X] If applicable, have you updated the documentation/manual?


amplipi/streams.py
```diff
@@ -1719,4 +1714,4 @@ def stream_types_available() -> List[str]:
     stypes.append(FMRadio)
   # the below line is not type checked because mypy isn't smart enough to see this is a relatively
   # constrained set of types, and instead evaluates this as a `list[type]`
-  return [s.stream_type for s in stypes] # type: ignore
+  return [s.stream_type for s in stypes] # type: ignore
```
linknum23, line 1717: Weird that this doesn't have a newline, editorconfig?

linknum23: Fix the neline and I think this PR makes sense.

mjustian: Removed the newline

linknum23 (approved):

PR #751, opened 2024-06-14
Sync streams info without causing a recursive import

linknum23:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # AmpliPi Software Releases
 
+## Upcoming Release
+* System
+  * Fix broken API generation
+
 ## 0.3.6
 * Web App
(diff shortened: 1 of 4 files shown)


rtertiaer (approved):
i mean, if feels acceptable to you, it works for me. one some level it feels like maybe removing the `_` prefix is acknowledging the state of the world better, but as-is I'm fine too cuz if it works it works :sunglasses: 

I used these changes to generate a new schema in #745 , so it definitely functions for that use case.

linknum23:
@klay2000 Can you make sure the Pandora station syncing that was needed still works. It worked for me, but I don't know all of your test cases.

klay2000 (approved):
Appears to work just fine when tested with Pandora!

PR #756, opened 2024-06-17
Player, Player Card synchronization

SteveMicroNova:
Adds marquee to player page, getting rid of font scaling issue Ryan and I found while I was showing off #747 related to using the same .scss file for both SongInfo and SongInfoMarquee

I've added the marquee to the player page, and as such SongInfoMarquee has fully replaced the original SongInfo and taken its name

I've also added MUI Grid support to the player card and player page as a fast method of using CSS grid styling to ensure that metadata always starts and ends in the same spots, rather than being a dynamically sized mess (see images below)

### Checklist


SteveMicroNova:
Old:
image
image


New:
image
image

Note that the new stuff is the marquee and so doesn't fit, but does scroll

The intended fix was the font size of the home page player cards on smaller screens, synchronizing the marquee to the player page as well (cutting down on total number of components, removing text wrapping that increases page height) was just a bonus

SteveMicroNova:
Temporarily converting to draft to show that this isn't ready actually

normal screen size doesn't work well on player page atm

image

SteveMicroNova:
Getting rid of the SongInfo may not have been the move, perhaps it continued to exist for a reason

I'll push more on this tomorrow, it's work party time

web/src/components/PlayerCard/PlayerCardFb.jsx
```diff
@@ -70,17 +70,30 @@ const PlayerCardFb = ({ sourceId, setVol }) => {
...
+                                    <StreamerOutputBadge sourceId={sourceId} />
+                                </div>
+                            )}
+                        </Grid>
+                        <Grid item xs={4} md={6}> {/* Spacer */} </Grid>
+                        <Grid item xs={4} md={3}>
+                            <SongInfo sourceId={sourceId}/>
+                        </Grid>
+
+                    </Grid>
```
SteveMicroNova, line 96: I've added support for the MUI Grid component, which is a fast way of implementing CSS grid formatting support. It has built in breakpoints (xs, sm, md, lg, xl) for different screen sizes, then you identify what relative size you want everything to be, interpreted as a percentage of adding all the same breakpoints up together

I always have my breakpoints add up to 12 as that's the tutorial on the mui website, though that isn't necessary. In this case, at the xs breakpoint everything is sized at 4/12 (33%) and at md and larger it's sized [3/12, 6/12, 3/12]  [25%, 50%, 25%]

The primary reason I went to grid formatting is this:

Old:
image


New:
image

SteveMicroNova: I've since changed these particular breakpoints a little, but the underlying explanation is still good for those who wish to know how the Grid breakpoints work

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
   * Make update available badge show up on both Settings page and the menu bar
   * Fix stream icons
   * Make it clearer when a stream needs a zone added on the home screen
+  * Enforce breakpoint styling to ensure that the UI looks the same between mobile, desktop, tablet viewports
```
rtertiaer, line 15: can you rebase this PR and put this under a `Next Release` heading?

rtertiaer (changes requested):
one change in the CHANGELOG but otherwise looks good to me. just to be explicit - you've tested this latest grid + breakpoint setup on different viewports/platforms yea?

SteveMicroNova:
> just to be explicit - you've tested this latest grid + breakpoint setup on different viewports/platforms yea?

Chrome, Firefox, and mobile safari yes
Chrome and Firefox both got window resizing treatments during testing, I can't do that on mobile safari but it seemed fine

rtertiaer (approved):

PR #759, opened 2024-06-18
PlayStatusVisibility1

SteveMicroNova:
Very basic attempt at #605, not so much expecting to solve the issue so much as fuel further conversation into it and give a basic proof of concept of an idea I had
I do not like this implementation, it's just something I could whip up in 5 minutes to point to during conversation, using tricks I learned since we last talked about this issue as a group

I've removed the PR checklist as this has no intention of being merged in


SteveMicroNova:
https://github.com/micro-nova/AmpliPi/assets/114415937/b22d4d82-50aa-4546-845b-72a1fa5d2f10

I've realized that I never made a video for this PR

SteveMicroNova:
I don't really like this option as much anymore as the wiggle might say "Click me!" to users, which means they're gonna mute it and then be confused why it's still shaking when they unmute it

PR #763, opened 2024-06-19
ResponsiveConfigPage

### LINKED ISSUE #580: No confirmation dialog for options on config page
When uploading a new config, resetting the box back to factory config, performing a HW reset, etc, It would probably be a good idea to first show a pop-up dialog confirming the choice. Then there should be some type of spinner while the action is performed, followed by some type of confirmation that the selected action has been completed.

SteveMicroNova:
Closes #580 by adding confirmation modals, loading spinners, success notifications

Also atomized/componentized the individual parts of the config page while I was at it, now it will be easier to add more sections (or edit them all at once) in the future

### Checklist


SteveMicroNova:
This is a draft PR due to the StatusBar component not working as expected
Figuring that out tomorrow, after that I suspect this ought to be a good change to find its way to main, unless we find the ConfigPanel component to be too awkward (it may well be)

web/src/pages/Settings/Config/ConfigPanels/LMSMode.jsx
```diff
@@ -0,0 +1,69 @@
...
+                element.click();
+            });
+        });
+    };
+
+    function Base(props) {
+        const { children } = props;
+        return(
+            <div>
+                Lyrion Media Server (LMS) Mode
```
SteveMicroNova, line 34: Calling attention to this

I noticed that it's called Lyrion in the changelog, thought we should change it here while I'm working here anyways

SteveMicroNova:
My own early opinion is that this currently relies on too many component templates that are impossible to explain anywhere, this isn't like a pydantic Base schema validation, I need to handle a lot of prop passing up and down

I'm going back to the drawing board a little bit to make that make sense for literally anyone who comes in after me

web/src/pages/Settings/Config/ConfigPanels/ConfigTemplates/ConfigModal.jsx
```diff
@@ -0,0 +1,34 @@
...
+import DialogContent from "@mui/material/DialogContent/DialogContent";
+import Button from "@mui/material/Button/Button";
+import PropTypes from "prop-types";
+
+export default function ConfigModal(props) {
+    const {
+        body,
+        confirm,
+        open,
+        setOpen,
```
rtertiaer, line 13: nit: I would probably rename this something like `onApply`. someone might wanna do something more than simply closing the modal here. it might also obviate `confirm()` too; looks like there's no calling one without the other. my gut feeling is that in this case flexibility is handy versus being too rigorous.

rtertiaer (approved):
I can see this a couple of ways; we sacrifice flexibility for some rigor and immediate ease around adding new config items. It's perhaps a little over-engineered for what it is, but it comes correct and addresses the bugs well.

rtertiaer (changes requested):
actually, one last thing - since there are visual changes here (modals & such), my approval is conditioned on seeing what those changes look like.

SteveMicroNova:
https://github.com/micro-nova/AmpliPi/assets/114415937/dc2d5f3f-a49b-4fac-b84f-14c8dac3e37f

See here for video of what these changes are + how they work

Now that I've made this video, I've noticed that we don't need the "Are you sure?" modal for config upload, I feel like since there's already 3 clicks to get to that point you don't need a 4th to make sure

linknum23:
First off these changes look great. I added something similar to the original app, but these look even better.

Second: I realize that I could actually deploy this, but what does a failure look like? I assume a junk config file would generate an error?

SteveMicroNova:
> Second: I realize that I could actually deploy this, but what does a failure look like? I assume a junk config file would generate an error?

No idea, I haven't seen a case of these erroring out before
I should test that "This isn't a config file" case, I feel like that's the only one that is realistically failable

SteveMicroNova:
> No idea, I haven't seen a case of these erroring out before and just included error modes as an insurance I should test that "This isn't a config file" case, I feel like that's the only one that is realistically failable

The answer is that the reinit function takes in a JSON file and simply figures it out. If there's data it doesn't understand, it just ignores it and dutifully processes the request as if you said "Set all config options to null"

the `ctrl.reinit()` function is too large for me to be curing in this PR, I'd think. Seems like it might be better to make a separate issue and separate PR, even after adding try-catching to `@api.post("/api/load/", tags=["config"])` I couldn't get it to fail (`response.ok = false` on the frontend) with the wrong type of JSON or even with an outright invalid JSON file

linknum23:
Ooops I forgot that wasn't a good candidate. We should probably have some sanity checks somewhere in the loader. Ideally they would actually come before reinit in the app.py.

I'm much more curious about how this fails generically. Like what happens when a reboot takes too long?

SteveMicroNova:
https://github.com/micro-nova/AmpliPi/assets/114415937/e298ac68-a04b-4eba-9b3c-995a1281519a

Added frontend error catching in the recent commit, now the frontend can error out without needing to await the status of `response.ok` from the backend
This video shows what it looks like when you comment out the if statement that checks for `response != null` without commenting out the code it contains, as an example of an error notification

rtertiaer (approved):

PR #764, opened 2024-06-21
Temporary File Player Stream for External Media

mjustian:
This change adds a new API endpoint, /api/play, which is used to create a temporary external media stream. This stream, titled "External Media" on the AmpliPi UI, is controlled using VLC similar to how the Internet Radio stations work. Additionally, Announcements also make use of temporary streams in the event of AmpliPi not resetting to the original configuration after the announcement plays. Finally, it is now possible to pause a FilePlayer stream.

### Checklist


amplipi/ctrl.py
```diff
@@ -1095,7 +1115,6 @@ def exec_stream_command(self, sid: int, cmd: str) -> ApiResponse:
         stream.restart()
       else:
         stream.send_cmd(cmd)
-      self.sync_stream_info()
```
linknum23, line 1098: Why was this removed?

mjustian: Was not supposed to be removed, put back

amplipi/streams.py
```diff
@@ -1373,13 +1399,43 @@ def wait_on_proc(self):
...
       if self.bkg_thread:
         self.bkg_thread.join()
-    self.proc = None
     self._disconnect()
+    self.proc = None
+
+  def _is_running(self):
+    if self.proc:
+      return self.proc.poll() is None
+    return False
```
linknum23, line 1438: Why is this duplicated from the base class?

amplipi/streams.py
```diff
@@ -31,7 +31,7 @@
 import threading
 import re
 import logging
-
+import random
```
linknum23, line 34: Is this used anymore?

amplipi/streams.py
mjustian: This function originally had additional behavior, and removing that left it with a copy of the base class version. Removing.

amplipi/streams.py
mjustian: No

amplipi/streams.py
```diff
@@ -1373,13 +1399,43 @@ def wait_on_proc(self):
...
+    if cmd in self.supported_cmds:
+      if cmd == 'next':
+        if self._is_running():
+          self.proc.kill()
+          if self.bkg_thread:
+            self.bkg_thread.join()
+        self.proc = None
+      if self.command_file_path is not None:
+        if cmd == 'pause':
+          logger.info('PAUSE')
```
linknum23, line 1411: Whats the point of UPPER_CASE here? Should we log what is getting paused?

streams/fileplayer.py
```diff
@@ -0,0 +1,263 @@
...
+  'state': 'stopped',
+}
+if args.song_info:
+  if not update_info(cur_info):
+    sys.exit(1)
+
+restarts: List[float] = []
+
+
+def restart_vlc():
```
linknum23, line 98: I'm unsure how useful the restart logic is in this script. We can leave it in for now though

amplipi/ctrl.py
```diff
@@ -1372,3 +1394,72 @@ def announce(self, announcement: models.Announcement) -> ApiResponse:
...
+        return resp0
+      stream = resp0
+
+    zones = []
+    for zone in self.status.zones:
+      if zone.source_id == media.source_id and zone not in zones:
+        zones.append(zone)
+
+    if len(zones) == 0:
+      return ApiResponse.error("No zones playing on this source")
```
rtertiaer, line 1429: why is this an error?

amplipi/streams.py
```diff
@@ -42,6 +41,7 @@
 
 from amplipi import models
 from amplipi import utils
+from amplipi import app
```
rtertiaer, line 44: pretty sure this re-introduces the bug fixed in https://github.com/micro-nova/AmpliPi/pull/751 ; can you try to run `./scripts/create_spec` and if it doesn't work rearrange your imports?

tests/test_rest.py
```diff
@@ -1270,6 +1270,24 @@ def test_load_preset(client, pid, unmuted=[1, 2, 3]):
...
+  # yes source
+  rv = client.post('/api/play', json={'media': nasa_audio, 'source_id': 0})
+  assert rv.status_code == HTTPStatus.OK, print(rv.text)
+  # db volume
+  rv = client.post('/api/play', json={'media': nasa_audio, 'source_id': 0, 'vol': -40})
+  # relative volume
+  rv = client.post('/api/play', json={'media': nasa_audio, 'source_id': 0, 'vol_f': 0.5})
+  assert rv.status_code == HTTPStatus.OK, print(rv.text)
+  rv = client.post('/api/play', json={'media': nasa_audio, 'source_id': 0, 'vol': -40, 'vol_f': 0.5})
+  assert rv.status_code == HTTPStatus.OK, print(rv.text)
```
rtertiaer, line 1291: not sure how reasonable this is, but it would be really nice if we could test that a stream was truly temporary.

amplipi/ctrl.py
```diff
@@ -651,6 +653,24 @@ def _get_source_config(self, sources: Optional[List[models.Source]] = None) -> L
...
+    for stream in self.status.streams:
+      if stream.type == 'fileplayer' and stream.temporary and stream.timeout is not None:
+        time = datetime.datetime.strptime(stream.timeout, '%Y-%m-%d %H:%M:%S.%f')
+        if time < datetime.datetime.now():
+          temp_streams.append(stream)
+
+    for source in self.status.sources:
+      for stream in temp_streams:
+        if source.input[7:].isdigit() and int(source.input[7:]) == stream.id:
+          temp_streams.remove(stream)
```
linknum23, line 668: Does this make the removal more obvious?
```suggestion
    conn_streams = {src.get_stream() for src in self.status.sources}
   temp_streams -= conn_streams
```

linknum23: temp_streams would have to be a set of stream ids

amplipi/ctrl.py
```diff
@@ -651,6 +653,24 @@ def _get_source_config(self, sources: Optional[List[models.Source]] = None) -> L
...
     return src_cfg
 
+  def _delete_unused_temporary_streams(self):
+    """Removes temporary file players if they are disconnected and have no connected sources"""
+    temp_streams = []
+    for stream in self.status.streams:
+      if stream.type == 'fileplayer' and stream.temporary and stream.timeout is not None:
+        time = datetime.datetime.strptime(stream.timeout, '%Y-%m-%d %H:%M:%S.%f')
+        if time < datetime.datetime.now():
+          temp_streams.append(stream)
```
linknum23, line 663: I feel like this is better accessed via the specific stream class with a method like expired()

amplipi/ctrl.py
```diff
@@ -651,6 +653,24 @@ def _get_source_config(self, sources: Optional[List[models.Source]] = None) -> L
...
+
+    for source in self.status.sources:
+      for stream in temp_streams:
+        if source.input[7:].isdigit() and int(source.input[7:]) == stream.id:
+          temp_streams.remove(stream)
+
+    for stream in temp_streams:
+      logger.info(f'Deleting unused temporary stream {stream.id}')
+      self.delete_stream(stream.id)
```
linknum23, line 672:
```suggestion
      self.delete_stream(stream.id, internal=True)
```

amplipi/ctrl.py
```diff
@@ -1372,3 +1394,69 @@ def announce(self, announcement: models.Announcement) -> ApiResponse:
...
+        update.vol = None
+        update.mute = False
+      else:
+        update.vol = zone.vol
+        update.vol_f = zone.vol_f
+        update.mute = zone.mute
+
+      update.source_id = media.source_id
+      update.name = zone.name
+      self.set_zone(zone.id, update, True, False)
```
linknum23, line 1449:
```suggestion
      if media.vol is not None or media.vol_f is not None: 
        update = models.ZoneUpdate(vol=media.vol, vol_f=media.vol_f, mute=False)
        self.set_zone(zone.id, update, internal=True)
```

amplipi/ctrl.py
```diff
@@ -1372,3 +1394,69 @@ def announce(self, announcement: models.Announcement) -> ApiResponse:
...
     return resp4
+
+  def play_media(self, media: models.PlayMedia) -> ApiResponse:
+    """Play media to a file player on a specified source"""
+    stream = None
+    # First, check if we already have a media player here
+    cur_stream = None
+    for source in self.status.sources:
+      if source.id == media.source_id and len(source.input) > 7:
+        cur_stream = int(source.input[7:])
```
linknum23, line 1405: I would suggest using source.get_stream() function from the source model

amplipi/ctrl.py
```diff
@@ -1372,3 +1394,69 @@ def announce(self, announcement: models.Announcement) -> ApiResponse:
...
+    if stream.id is not None:
+      self.streams[stream.id].reconfig(url=media.media, temporary=True,
+                                       timeout=str(datetime.datetime.now() + datetime.timedelta(seconds=5)))
+      self.sync_stream_info()
+      self.mark_changes()
+
+    src_update = models.SourceUpdate()
+    src_update.name = None
+    src_update.input = f'stream={stream.id}'
+    self.set_source(media.source_id, src_update, True, True)
```
linknum23, line 1460:
```suggestion
    self.set_source(media.source_id, src_update, internal=True, force=True)
```

amplipi/ctrl.py
```diff
@@ -1372,3 +1394,69 @@ def announce(self, announcement: models.Announcement) -> ApiResponse:
...
+    stream = None
+    # First, check if we already have a media player here
+    cur_stream = None
+    for source in self.status.sources:
+      if source.id == media.source_id and len(source.input) > 7:
+        cur_stream = int(source.input[7:])
+
+    logger.info(f'Stream found is {cur_stream}')
+    for tmp_stream in self.status.streams:
+      if tmp_stream.id == cur_stream and tmp_stream.type == 'fileplayer':
```
linknum23, line 1409: using self.streams might simplify this. Something like:
```
if instanceof(self.streams.get(tmp_stream.id), streams.FilePlayer)

**Review by linknum23 — CHANGES_REQUESTED** [review]
Getting closer. Needs a little cleanup in play_media()

File: amplipi/streams.py [hunk 16]
```diff
@@ -1346,21 +1358,43 @@ def reconfig(self, **kwargs):
...
-    if not self.mock:
+    if self.mock:
+      logger.info(f'{self.name} connected to {src}')
+      self.state = 'playing'
+      self.src = src
+      self.bkg_thread = threading.Thread(target=self.wait_on_proc)
+      self.bkg_thread.start()
+      return
+
+    if src is not None:
```
linknum23, line 1376:
```suggestion
    if not self.mock:
```

amplipi/streams.py
```diff
@@ -1346,21 +1358,43 @@ def reconfig(self, **kwargs):
...
+      ]
+      logger.info(f'running: {self.vlc_args}')
+      self.proc = subprocess.Popen(args=self.vlc_args, preexec_fn=os.setpgrp)
+
+      # make a thread that waits for a couple of secends and returns after setting info to stopped
+      self.bkg_thread = threading.Thread(target=self.wait_on_proc)
+      self.bkg_thread.start()
+
+      self.state = 'playing'
+      self.src = src
```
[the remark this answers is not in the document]
linknum23: These changes make the flwo fimilar to before

amplipi/streams.py
```diff
@@ -1787,7 +1846,13 @@ def build_stream(stream: models.Stream, mock=False) -> AnyStream:
   if stream.type == 'aux':
     return Aux(name, disabled=disabled, mock=mock)
   if stream.type == 'fileplayer':
-    return FilePlayer(name, args['url'], disabled=disabled, mock=mock)
+    if 'temporary' not in args.keys():
+      args['temporary'] = None
+    if 'timeout' not in args.keys():
+      args['timeout'] = None
+    if 'has_pause' not in args.keys():
+      args['has_pause'] = True
```
linknum23, line 1854: can you use the `args.get("KEY", default_value)` paradigm instead here?

linknum23 (approved):
I've tested this and it works well. Please address the small suggestions before merging.

Issue #498, opened 2023-08-16
Make settings tabs more consistent style wise


SteveMicroNova:
(Ex: every settings page has an X in the top right to go back to the settings page, but the Updates page has [<- Back to App] button in the top left instead, Configuration has the wrong font)

SteveMicroNova:
Since writing this issue, I've been made aware that the Updates page may be separate and different on purpose, I'd like to affirm whether or not we want to actually change it @linknum23

Changing it would require either merging it into our current react app so it can make use of the same stylings, or making it into its own react app and copying a lot of the same style info over

linknum23:
The updater is a completely separate app. Migrating it would be a fair amount of work.
 I like the idea of making it look pretty like our current UI, but I think we may want to wait on our changes to the updater for imaging to avoid doing extra work.

SteveMicroNova:
> I think we may want to wait on our changes to the updater for imaging to avoid doing extra work.

Got it, splitting this issue out and adding the Updater page portion to #780

PR #767, opened 2024-06-24
Settings UI Fixes

### LINKED ISSUE #554: Settings mobile view
image
The tops of every settings page are cut off on mobile (at least on my iPhone 8)
This makes all of the settings titles be cut in half as shown in the image

SteveMicroNova:
Working on a handful of issues relating to setting page UI concerns, as seen in the 0.4.0 project board UI/UX section

Intends to close #554 and #498

### Checklist


web/src/components/PageHeader/PageHeader.scss
```diff
@@ -1,7 +1,8 @@
 @use "src/general.scss";
 
 .page-header {
-  margin: 0.5rem;
+  padding-top: 2rem;
+  margin: 1rem;
```
SteveMicroNova, line 5: This fixes #554 by pushing the top of the page down, needs to be tested in the app as the issue we were having is entirely related to mobile safari, but this change will effect the app as well

I've only tested with safari thus far as I can locally host a webserver, but I can't advertise that webserver to the app afaik

SteveMicroNova: image

I've deployed to Lamplipi and it does not negatively effect the mobile layout

SteveMicroNova: Note that this is on an iPhone 8, the big gap at the top of the screen between the time and battery life info isn't part of the screen, but is where the camera and phone speaker go

SteveMicroNova: The true root of this issue was the List components that were custom built having weird styling that pushed stuff up for some reason

I've undone these changes and have not regained the problem they aimed to fix, so :+1:

web/src/pages/Settings/Settings.jsx
```diff
@@ -77,102 +94,85 @@ const Settings = ({ openPage }) => {
...
-                                        UnsaveURL();
-                                    }
-                                    setIsSavedUrl(e.target.checked);
-                                }}
-                            />
-                        </div>
-                    )}
-                </List>
-            </div>
+            <List style={{width: "400px", maxWidth: "100%"}}>
```
SteveMicroNova, line 97: This inline styling is an approximation of the previous styling, I used an approximation as I couldn't find what was making the old list be as wide as it was when the new one wasn't like that by default

It looks the exact same on my computer screen, but will never be too large to fit the screen either. Should be the same on mobile, too; testing shortly

SteveMicroNova:
This PR is ready for review pending an answer to [this question](https://github.com/micro-nova/AmpliPi/issues/498#issuecomment-2197275098) on #498

SteveMicroNova:
This code is currently deployed to lamplipi, if you wish to see it while reviewing. You should not see a difference when switching between inhouse and lamplipi on the settings pages now., see image at top of the thread to see what was solved by replacing the List and ListItem components

amplipi/updater/static/index.html
```diff
@@ -18,13 +18,19 @@
...
+      }
+    </script>
   </head>
 
   <body>
 
     <main role="main" class="container">
       <p class="">
-        <a id="back-to-app" class="btn btn-primary btn-sm text-center" role="button" href="/" onclick="javascript:event.target.port=80"><i class="fa fa-arrow-left"></i> Back to App</a>
+        <a id="back-to-app" class="btn btn-primary btn-sm text-center" role="button" href="/" onclick="redirectToSettings()"><i class="fa fa-arrow-left"></i> Back to App</a>
```
SteveMicroNova, line 33: This change causes tests to fail and stops config upload from working for some reason
Looking into this now

SteveMicroNova: It specifically breaks `test_zeroconf` with an assertion error as the last line in the file asserting that `5000 == 9898` and not liking that

amplipi/updater/static/index.html
```diff
@@ -24,7 +24,7 @@
 
     <main role="main" class="container">
       <p class="">
-        <a id="back-to-app" class="btn btn-primary btn-sm text-center" role="button" href="/" onclick="javascript:event.target.port=80"><i class="fa fa-arrow-left"></i> Back to App</a>
+        <a id="back-to-app" class="btn btn-primary btn-sm text-center" role="button" href="/#/settings" onclick="javascript:event.target.port=80"><i class="fa fa-arrow-left"></i> Back to App</a>
```
SteveMicroNova, line 27: Solving the zeroconf error was much simpler than I thought

rtertiaer:
I think this PR introduced #777 ; can you please stylize these consistently? inconsistent font families makes the app feel unpolished.

rtertiaer (changes requested):
I think this PR introduced https://github.com/micro-nova/AmpliPi/issues/777 ; can you please stylize these consistently? inconsistent font families makes the app feel unpolished.

SteveMicroNova:
> I think this PR introduced #777 ; can you please stylize these consistently? inconsistent font families makes the app feel unpolished.

Could you point to an example of this issue within this branch? The fonts look the same to me, maybe just with a different weight for the header vs the others but no outright different fonts afaik

web/src/App.jsx
```diff
@@ -18,8 +18,10 @@ import PropTypes from "prop-types";
 export const usePersistentStore = create(
     persist(
         (set) => ({
-            selectedSource: 0,
+            selectedSource: null,
             setSelectedSource: (selected) => {
+                console.log("SELECTED:");
+                console.log(selected);
```
rtertiaer, line 24:
```suggestion
```

web/src/pages/Settings/Groups/Groups.jsx
```diff
@@ -31,27 +32,31 @@ const GroupListItem = ({ group, zones }) => {
...
+                        zones={zones}
+                        onClose={() => {
+                            setModalOpen(false);
+                        }}
+                        del={deleteGroup}
+                        apply={editGroup}
+                    />
+                )}
+            </ListItem>
+            <Divider component="li" />
```
rtertiaer, line 58: honest question; does it look better without the divider and just bigger margins? that's what seemed to work well on the stream page, but maybe these cases are different?

rtertiaer (changes requested):
can you squash this? looks pretty good otherwise!

web/src/pages/Settings/Groups/Groups.jsx
SteveMicroNova: > that's what seemed to work well on the stream page

The stream page also has these dividers, they're added here due to the issue being "Make all the errant stylings in the settings tabs fall in line"
I think you'll have to clarify somewhat when you're back to show me what you mean

web/src/App.jsx
SteveMicroNova: Good catch, this seems to have made it in due to me switching branches after testing #788

web/src/pages/Settings/Groups/Groups.jsx
rtertiaer: personally I think dividers should separate sections, not individual items in a list. imo visually it reads more like a list if there isn't separation. for example, [material design doesn't use a divider but only whitespace between list items](https://m3.material.io/components/lists/overview).

SteveMicroNova: That page has 1 instance of being whitespace separated:
image

And two instances of divider-separation:
image
image


So it seems to be a matter of personal preference, I may say we bring this up in todays meeting to poll for group preference

rtertiaer: i think i'd prefer to see a before and after of AmpliPi with & without dividers, if we're gonna compare anything. it looked better in the one case where I removed them; not convinced that's true here too, which is why i asked instead of suggested.

SteveMicroNova: Before:
image

After:
Screenshot 2024-07-22 130914

linknum23: I like the lines!

rtertiaer: lines it is then :sunglasses:

web/src/App.jsx
```diff
@@ -18,7 +18,7 @@ import PropTypes from "prop-types";
 export const usePersistentStore = create(
     persist(
         (set) => ({
-            selectedSource: 0,
+            selectedSource: null,
```
rtertiaer, line 21: why did you do this? if this is a bug, it's not explained.

0 is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), and it appears you just moved the setting logic to something in PlayerCardFb

SteveMicroNova: This looks like I was attempting to do an early crack at what I ended up doing in #788, being that this handles autoselection and is in a branch that predates the successfully merged autoselection functionality

rtertiaer (approved):
does this fix https://github.com/micro-nova/AmpliPi/issues/833 too? :sunglasses: :clap:

SteveMicroNova:
> does this fix https://github.com/micro-nova/AmpliPi/issues/833 too?
Let me deploy this and check real quick, if it does I'll connect that issue before merging so it gets closed as well

SteveMicroNova:
> Let me deploy this and check real quick, if it does I'll connect that issue before merging so it gets closed as well

No

Issue #766, opened 2024-06-23
Constant "Error: 'playlist_loop', trying again in 2 seconds..." messages when LMS playlist is empty


sumnerboy12:
I am running LMS on a different server (192.168.10.112). I noticed recently that if I clear the playlist on LMS, for the AmpliPi client I have configred (via the LMS web UI) I start see the following messages in the Amplipi logs;

```
"2024-06-23T22:49:05.868+12:00","amplipi","DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): 192.168.10.112:9000"
"2024-06-23T22:49:05.872+12:00","amplipi","DEBUG:urllib3.connectionpool:http://192.168.10.112:9000 ""POST /jsonrpc.js HTTP/1.1"" 200 240"
"2024-06-23T22:49:05.876+12:00","amplipi","Error: 'playlist_loop', trying again in 2 seconds..."
"2024-06-23T22:49:05.877+12:00","amplipi","INFO:__main__:Error: 'playlist_loop', trying again in 2 seconds..."
"2024-06-23T22:49:06.685+12:00","amplipi","INFO:     192.168.10.21:59830 - ""GET /api HTTP/1.1"" 200 OK"
"2024-06-23T22:49:07.885+12:00","amplipi","DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): 192.168.10.112:9000"
"2024-06-23T22:49:07.889+12:00","amplipi","DEBUG:urllib3.connectionpool:http://192.168.10.112:9000 ""POST /jsonrpc.js HTTP/1.1"" 200 239"
"2024-06-23T22:49:07.893+12:00","amplipi","Error: 'playlist_loop', trying again in 2 seconds..."
"2024-06-23T22:49:07.893+12:00","amplipi","INFO:__main__:Error: 'playlist_loop', trying again in 2 seconds..."
```

As soon as I add something to the LMS playlist these message stop, and then immediately start again if I clear the playlist. Seems very reproducible.

rtertiaer:
Some of this is definitely as a result of setting the log level to DEBUG; it's exposed urllib3's DEBUG  level logging as well, which is definitely less than ideal.

sumnerboy12:
Yeah not too worried about the `DEBUG` logging, although agree it would be nice to quieten it down a bit.

It is the `Error:` logging that concerns me - is that an error being returned from LMS or generated in the AmpliPi LMS client code?

sumnerboy12:
[...]

PR #768, opened 2024-06-24
Reduce noise from LMS automated media device mounting script, service

### LINKED ISSUE #733: Fix broken LMS client connection, or stop spamming logs
It's not clear what's prompting [this to happen](https://github.com/micro-nova/AmpliPi/pull/682#issuecomment-2135447077), but it ought not. Either correct the invocation involved here, or gracefully handle the disconnect case instead of spamming the logs with tracebacks.

Related to #682.

SteveMicroNova:
Closes #733 by making fewer logs and only running the noisy part when it's meant to

### Checklist

### DIFF
--- a/scripts/edit_media_directories.py
+++ b/scripts/edit_media_directories.py
@@ -5,6 +5,8 @@ import requests
 import logging
 import sys
 import subprocess
+import urllib3
+import urllib3.exceptions
 
 def CheckLMSMode():
     try:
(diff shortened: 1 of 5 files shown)


SteveMicroNova:
Needs to be tested with an actual USB drive during LMS mode
I do not currently have a USB drive

rtertiaer:
might make sense to roll #766 into this as well

rtertiaer:
I'll review this when you've tested it

SteveMicroNova:
> Needs to be tested with an actual USB drive during LMS mode
> I do not currently have a USB drive

I recovered my USB drive at end of day yesterday, after some further tweaks it now works just fine, even caught some other noisy errors that were nonissues caused by a race condition during startup (will point those out in a self review shortly)

scripts/edit_media_directories.py
```diff
@@ -20,25 +31,42 @@ def get_usb_drives(logger: logging.Logger):
...
+        for drive in usb_drives:
+          drives += f"{drive}, "
+        logger.info(f"Detected drives: {drives}")
+        logger.info("If you wish to mount these to the internal LMS server, please go to Config and set LMS Mode")
+  except urllib3.exceptions.MaxRetryError:
+    # This error is extremely common during startup, but isn't actually broken
+    # It seemingly only occurs on startup, leading me to believe it is a race condition with other things starting up
+    logger.error("LMS Automated Drive Mounting has encountered an error due to a race condition during startup, if this persists (or happened outside of startup) please contact support@micro-nova.com with a copy of these logs")
+  except Exception as e:
+    logger.error(f"LMS Automated Drive Mounting has encountered an error: {e}")
```
SteveMicroNova, line 68: Note the comment here, the requests.post above will throw errors if it attempts to run while LMS is starting, but will be fine once LMS is finished loading

scripts/edit_media_directories.py
```diff
@@ -5,11 +5,22 @@
 import requests
 import logging
 import sys
+import subprocess
+import urllib3
+import urllib3.exceptions
+
+def CheckLMSMode():
```
linknum23, line 12: CamelCase is typically used only for classes in python. Is there a specific reason this and the LMSMode variable below are in camel case?

scripts/edit_media_directories.py
```diff
@@ -20,25 +31,43 @@ def get_usb_drives(logger: logging.Logger):
...
-  for d, drive in enumerate(usb_drives):
-    data[f"pref_mediadirs{d}"] = drive
-    data[f"pref_ignoreInAudioScan{d}"] = 1
-
-  logger.info("Adding drives to LMS settings...")
-  requests.post(url, data=data, verify=False, timeout=5)
-  logger.info("Drives added, LMS scanning now.")
+  try:
+    usb_drives = get_usb_drives(logger) # Get list of available drives
+    LMSMode = CheckLMSMode()
```
linknum23, line 36: This assignment reads confusing it looks like you are assigning a class/type from an instantiated class.

scripts/edit_media_directories.py
SteveMicroNova: CamelCase is simply how I write code, back when I was taught in school originally that's how we did everything and that practice has stuck with me

scripts/edit_media_directories.py
```diff
@@ -20,25 +31,43 @@ def get_usb_drives(logger: logging.Logger):
...
+      logger.info("Drives added, LMS scanning now.")
+    else:
+      if len(usb_drives) > 0:
+        drives = ""
+        for drive in usb_drives:
+          drives += f"{drive}, "
+        logger.info(f"Detected drives: {drives}")
+        logger.info("If you wish to mount these to the internal LMS server, please go to Config and set LMS Mode")
+  except (urllib3.exceptions.MaxRetryError, urllib3.exceptions.NewConnectionError, requests.exceptions.ConnectionError):
+    # This error seemingly only occurs on startup, leading me to believe it is a race condition
```
linknum23, line 64: Have we made any attempt to delay this service? Maybe make it dependent on the Logitech media server?

SteveMicroNova: I've made a few attempts in my last few commits, between creating the function that checked if lms was running to the systemd dependency attempts, it gives the same results of having this error print out 4 times during startup:
>Jun 25 15:30:10 lamplipi udisks2-listener.sh[551]: LMS Automated Drive Mounting has encountered an error: HTTPConnectionPool(host='localhost', port=9000): Max retries exceeded with url: /settings/server/basic.html (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x75db4390>: Failed to establish a new connection: [Errno 111] Connection refused'))

If there's another way to do this that I'm not aware of, please let me know

linknum23:
I think the best way would be to add a dependency in the service file. I
think you can use the after keyword:
Here's an example that waits for the  network to be up and ssh-keygen to be
running

    After=network.target sshd-keygen.service

I assume you could do the same thing for logitechmediaserver.service

You may also want to check out using the Wants and Requires keywords as
well to specify the LMS dependency.

SteveMicroNova:
I think that https://github.com/micro-nova/AmpliPi/pull/768/commits/8fc3525765ec69d12c063f06d13499afbb06b043 should solve the disjointed startup times, will test when I have access to a unit to test on tomorrow

SteveMicroNova:
I've attempted many permutations of "Only start when logitechmediaserver.service is running" between python code and systemd scripting but none have kept the script from winning the race condition and printing exactly 4 copies of the same error into the logs, it seems that the logitechmediaserver.service advertises itself as good to go a few seconds before that's true

Unless there's some other way to hook it in, I'll suggest to either:
1. live with exactly 4 error messages on every boot cycle
2. suppress the errors by catching without reraising
3. have a time.sleep(8) at the start of the script so that LMS fully loads rather than getting 4 of the same error with 2 seconds between each one

SteveMicroNova:
I contacted Ryan separately about my concerns and he sent me this: [https://www.slingacademy.com/article/python-requests-auto-retry-for-failed-http-requests/](https://www.slingacademy.com/article/python-requests-auto-retry-for-failed-http-requests/)

Which solved the issue entirely. No more random LMS errors in the console.

rtertiaer (approved):
:sunglasses:

PR #770, opened 2024-06-24
make only streams that aren't already playing selectable in StreamsModal

### LINKED ISSUE #501: Add a check for if a stream is already being played before it can be added to another source
Add a check for if a stream is already being played before it can be added to another source so that you can't accidentally steal music from others (Bonus points if you make the chosen zone tune in to the source already playing the stream)

rtertiaer:
This PR makes modifies the StreamsModal, which is what launches when you attempt to start a new stream. It filters out all already-playing streams from the selection dialog. This is a bit of an inverted way to close #501 , but is possibly a bit more intuitive and helps avoid messy logic around keeping streams running on the same source with zero interruption and just adding zones to the source, which is what the ticket details. Instead, we shift left and make the user do this if this is their intention.

In the non-webapp context, one can still steal streams; that's explicitly supported functionality and lives within the `ctrl.py` code.

I'm pretty sure the `input` field is always either `'None'` or starts with `stream=`, but that's one way this is brittle, for sure.

### Checklist

* [x] If this is a UI change, have you tested it across multiple browser platforms? Firefox & Chromium

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * Enforce breakpoint styling to ensure that the UI looks the same between mobile, desktop, tablet viewports
   * Groups containing disabled zones now behave as though those zones don't exist
   * Groups containing only disabled zones no longer selectable
+  * Only show not-playing streams in the new stream modal
 * Streams
   * Remove stop command (only accessable through API) from Pandora
   * Disconnect zones from sources when they are disabled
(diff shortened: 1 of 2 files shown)


SteveMicroNova (approved):
We love quick and easy ways of solving issues. Props to bringing this up in the meeting as an "Is this intended or no?" situation, only removing the issue from the frontend and leaving it as a feature in the backend

PR #788, opened 2024-06-28
PlayerCardSelection

### LINKED ISSUE #529: Make Home Page Player Selection Intuitive
This issue was added due to [customer feedback](https://amplipi.discourse.group/t/home-assistant-integration-error-with-api-url/261/14?u=amplipi)

Make the players on the home page obviously selectable
OR
Make it so that there is never a time where there isn't a player selected either by means of setting the default selected player to the first, or by using cookies to determine which was the last selected player on that device

SteveMicroNova:
Closes #529 
How to close that issue is pending, presently this is a draft with the first idea that came to mind so we can talk about it

### Checklist

* [x] Does your submission pass linting & tests? You can test on localhost using `./scripts/test`* [x] If this is a UI change, have you tested it across multiple browser platforms?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
   * Migrate more inputs to MUI
   * Fix internet radio searching during stream creation
   * Download appliance config when toggling LMS mode off
+  * Add hoverover effects on unselected player cards
+  * Handle missing source selection
+  * Autoselect newly created sources
 * System
   * Reduce noisy logging from lms_metadata.py
   * Fix install of FM radio software `redsea`
--- a/web/src/utils/selectActiveSource.jsx
+++ b/web/src/utils/selectActiveSource.jsx
@@ -1,0 +1,40 @@
+import { useStatusStore, usePersistentStore } from "@/App";
+
+// Import and use this function on every page that relies on selected sources
+// Currently, that means the Home and Player pages
+export default function selectActiveSource(){ // Selects an active source, if the selected source is inactive
+    const sources = useStatusStore((s) => s.status.sources);
+    const selectedSource = usePersistentStore((s) => s.selectedSource);
+    const setSelectedSource = usePersistentStore((s) => s.setSelectedSource);
+    const autoselectSource = usePersistentStore((s) => s.autoselectSource);
+    const setAutoselectSource = usePersistentStore((s) => s.setAutoselectSource);
+    let run = false;
+
+    // If selected source is not active, select the source with the lowest ID
+    if(!sources || !sources[selectedSource] || !sources[selectedSource]["input"]){
+        // While testing this script along with the one that autoselects the created source when adding a new source,
+        // I found that there were times that a source could reach an undefined variable
+
+        // Those won't reach prod on my watch, but this catch exists so that you don't need to go around commenting out
+        // every instance of sources[x]["input"] in every file if you manage to reach that spot in your own development
+
+        run = true;
+        console.log("Source data missing, selecting a new source");
+    }
+    if(autoselectSource){
+        if(run || sources[selectedSource]["input"] === "None"){
+            for(let i = 0; i < sources.length; i++){
+                if(sources[i]["input"] != "None"){
+                    setSelectedSource(i);
+                    break;
+                }
+            }
+        }
+    } else if (sources[selectedSource]["input"] !== "None"){
+        // If presently selected source is valid/loaded, sources can be autoselected
+
+        // autoselectSource is only set to false if the source that's been selected is currently loading,
+        // so sources[selectedSource]["input"] will equal "None" for a limited time
+        setAutoselectSource(true);
+    }
+}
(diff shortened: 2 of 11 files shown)


SteveMicroNova:
One failure with this present implementation is that it doesn't work on mobile, which makes it a bit of a nonstarter imo but it was my initial thought, I've found it's better to come to the table with a baseline idea for people to say yes or no to rather than ask them to pick something out of the ether themselves when it comes to UI stuff.

https://github.com/micro-nova/AmpliPi/assets/114415937/3f620f9f-f5a1-487e-8e4f-5713a7a6a901

After having a short talk with Lincoln just now, his preference is to solve the other half of this issue instead, simply make one of the cards selected by default; I prefer the other version as making them obviously selectable will make it so that we don't have a "How do I change what the controls page is controlling" question in the future.

Though on the other hand, perhaps having one of them always selected and having the users notice that it is selected on the player page will lead to them realizing things are selectable in the first place, I feel like that's entirely possible within the [user story](https://amplipi.discourse.group/t/home-assistant-integration-error-with-api-url/261/15?u=amplipi) that underpins this issue

linknum23:
I think the hover functionality works great for desktop but it has no equivalent for mobile. My vote is to keep the hover but also select the first? card if none is selected.

web/src/utils/selectActiveSource.jsx
SteveMicroNova, line 1: This is the file that has the majority of the meat of this PR, the rest is just calling this function and catching formatting mistakes as well as the CSS thing shown near the top of the thread

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
   * Migrate more inputs to MUI
   * Fix internet radio searching during stream creation
   * Download appliance config when toggling LMS mode off
+  * Add hoverover effects on player cards
+  * Enforce source selection
```
SteveMicroNova, line 9: I was uncertain how to word this without being too verbose, I feel that this has the opposite issue of being too low detail to explain "If there is a source that exists, you will always have a source selected"

SteveMicroNova:
The easiest way to test this, if you want to see it work, is have the same unit open in two windows and close the source that the other window has selected

Previously, that would lead to being sourceless; now it autoselects a different source (if available)


https://github.com/user-attachments/assets/980f4ee5-b2f9-44b0-812f-6bff8d13ede1


This was somewhat difficult to show in a concise video without a voiceover, hopefully me showing the two different units across 4 windows and the difference of what happens when you close a selected source gets the point across

SteveMicroNova:
I've requested review from @linknum23 as well, as Klayton is out of office today

Lincoln, it's alright if you don't have time for this PR, I've primarily requested your review since you gave early thoughts on this PR and asked for a feature that was added along the way

CHANGELOG.md
linknum23: Maybe handle missing source selection. Current wording is fine btw.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
   * Migrate more inputs to MUI
   * Fix internet radio searching during stream creation
   * Download appliance config when toggling LMS mode off
+  * Add hoverover effects on unselected player cards
+  * Hanlde missing source selection
```
linknum23, line 9: typo

web/src/utils/selectActiveSource.jsx
```diff
@@ -2,18 +2,39 @@ import { useStatusStore, usePersistentStore } from "@/App";
...
-        for(let i = 0; i < 4; i++){
-            if(sources[i]["input"] != "None"){
-                setSelectedSource(i);
-                break;
+    if(!sources || !sources[selectedSource] || !sources[selectedSource]["input"]){
+        // While testing this script along with the one that autoselects the created source when adding a new source,
+        // I found that there were times that a source could reach an undefined variable
+
+        // Those won't reach prod on my watch, but this catch exists so that you don't need to go around commenting out
+        // every instance of sources[x]["input"] in every file if you manage to reach that spot in your own development
```
SteveMicroNova, line 19: Those times that sources[selectedSource]["input"] were caused by my testing, and can only be caused while mutating how we handle selection. There are no such cases in the current head of main.

web/src/utils/selectActiveSource.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+        run = true;
+        console.log("Source data missing, selecting a new source");
+    }
+    if(autoselectSource){
+        if(run || sources[selectedSource]["input"] === "None"){
+            for(let i = 0; i < 4; i++){
```
linknum23, line 26: Should this use the js equivalent of len(sources)  in place of 4, just in case we ever have a different number of sources.

SteveMicroNova: https://github.com/micro-nova/AmpliPi/pull/788/commits/7f0e10e65a270cb4f732a59a8334d20020b3326a

linknum23 (approved):
This is working great. I think it is a nice improvement to the UX.

PR #790, opened 2024-07-03
AmpliPro Purchase QR Codes

SteveMicroNova:
Add two new QR codes that redirect to the new purchase site for AmpliPro hardware

There are no code changes and so I've removed the PR Template, these files are not used here but this seems to be where all the other QRs have gone so far and so they are being added here as well
One of these will be made defunct shortly, and will then be removed from this PR afterwards. The decision between keeping one or the other is what makes this a draft PR at this moment


SteveMicroNova:
Requesting review from Ryan as I asked his opinion on the redirect strategy, requesting review from Jason as he's the brand manager
I will not merge until both approve

gorski123:
These probably belong in the marketing repo

SteveMicroNova:
> These probably belong in the marketing repo

I'm inclined to agree, though some are required here to build the manual and I don't know how to go about managing that.
Either we have some things in the wrong place (what we currently have) , some things in multiple places (which is the right version?), or have to check two places for things that are otherwise the exact same (increasing the amount of tribal knowledge to keep in mind when we hire more people). Perhaps @rtertiaer or @linknum23 have some sort of answer for how to solve that conundrum.

SteveMicroNova:
Rerequesting review from @gorski123 and @rtertiaer now that these images have largely been removed (though some changes still remain)

Ryan and I came to the conclusion that it's ok to keep these two QR codes in two places, and added a README to both places to ensure that they're upkept properly

rtertiaer (changes requested):
can you squash this?

SteveMicroNova:
> can you squash this?

Done

rtertiaer:
why are there still 5 commits?

rtertiaer (approved):

PR #799, opened 2024-07-08
properly check the API during updates if it requires authentication

### LINKED ISSUE #761: Password protected APIs fail to upgrade

rtertiaer:
This PR properly checks URLs during the update process in the case where they requrie authentication. This closes #761 .

### Checklist

### DIFF
--- a/scripts/configure.py
+++ b/scripts/configure.py
@@ -15,6 +15,7 @@ import time
 import re
 import sys
 import requests
+import json
 
 # pylint: disable=broad-except
 # pylint: disable=bare-except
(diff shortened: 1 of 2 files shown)


scripts/configure.py
```diff
@@ -823,8 +824,29 @@ def _copy_old_config(dest_dir: str) -> None:
       pass
 
 
+def _api_key() -> Union[bool, str]:
```
linknum23, line 827: Its probably a little more pythonic to return Optional[str] here.

linknum23 (approved):
LGTM, assuming this was tested in an update with and without a key

scripts/configure.py
linknum23: Oddly that change doesn't even effect the downstream logic.

rtertiaer:
Given #812 , I may re-work this idea and instead just modify the auth to not be necesasry if the `Host` is localhost.

rtertiaer:
actually, simply checking the `Host` header invites us to [header smuggling attacks](https://www.intruder.io/research/practical-http-header-smuggling). I'm gonna implement that same methodology for the display server instead.

Issue #604, opened 2024-01-05
Internet radio stream creation is a little rough


linknum23:

rtertiaer:
Make the list of searched radio stations responsive.

Also make it evident when you have clicked a station that it has auto-filled out fields for you. Leave the station highlighted when clicked.

rtertiaer:
This seemingly got worse with #763 - the fields no longer auto fill out 😬

PR #800, opened 2024-07-09
Fixes to the stream create/update dialog

rtertiaer:
This PR improves the stream update/create dialog. It migrates every last input to an MUI component, adds some margins to permit easy use on mobile, makes the enter button search in the search field, and a bit more.

This was prompted by the internet radio search functionality no longer auto-filling fields, referenced in #604. That bug is fixed in this PR as well. However, this does not close out that ticket.

Before:
Screenshot from 2024-07-09 14-25-56

After:
Screenshot from 2024-07-09 14-31-02

After on my iOS mobile:
signal-2024-07-09-150238

### Checklist

### DIFF
--- a/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
+++ b/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
@@ -1,7 +1,5 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { CircularProgress } from "@mui/material";
-import { Divider } from "@mui/material";
 import Alert from "@mui/material/Alert";
 import "./StreamModal.scss";
 import StreamTemplates from "../StreamTemplates.json";
@@ -9,6 +7,11 @@ import ModalCard from "@/components/ModalCard/ModalCard";
 import TextField from '@mui/material/TextField';
 import Button from '@mui/material/Button';
 import Checkbox from '@mui/material/Checkbox';
+import List from '@mui/material/List';
+import ListItemAvatar from '@mui/material/ListItemAvatar';
+import ListItemText from '@mui/material/ListItemText';
+import ListItemButton from '@mui/material/ListItemButton';
+import Avatar from '@mui/material/Avatar';
 
 const NAME_DESC =
   "This name can be anything - it will be used to select this stream from the source selection dropdown";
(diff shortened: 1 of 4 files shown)


web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -7,62 +7,63 @@ import "./StreamModal.scss";
...
     return (
         <>
             <div className="stream-field">
                 <TextField
+                    className="stream-field-input"
                     type={type}
                     label={name}
                     error={error}
-                    defaultValue={defaultValue}
+                    value={value ? value : ''}
```
rtertiaer, line 28: https://stackoverflow.com/questions/68480218/material-ui-textfield-label-does-not-move-up-when-the-value-for-textfield-is-set

rtertiaer:
Alright, I just buckled down and did the full rendition. The latest commit fixes #604 by bringing in a bunch of MUI components that handle reactivity for us. A little bit of styling, move some rendering around, and this oughta be kosher.

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -96,9 +102,10 @@ const InternetRadioSearch = ({ onChange }) => {
     const [host, setHost] = React.useState("");
     const [results, setResults] = React.useState([]);
     const [query, setQuery] = React.useState("");
+    const [selectedUuid, setSelectedUuid] = React.useState("");
 
     const search = (name) => {
-        setResults([<CircularProgress key={name} />]);
+        setResults([{name: "Loading..."}]);
```
rtertiaer, line 108: Because keeping the state for which item was selected requires dynamically rendered items from the actual data structure itself, the spinner was removed. I've tried a couple of ways to get the `Avatar` component to do anything a little more sophisticated here and a couple of different ways of rendering, but I quickly approached a point of diminished returns on my time and landed on this instead. Let me know if it's a killer feature and I can re-introduce it with fresh eyes.

rtertiaer:
what the latest commit looks like: 
image

klay2000 (approved):
Looks good, if you want you should try making the two buttons on the bottom look better too, imo they've always been the most ridiculous looking part of this dialog.

rtertiaer:
I themed all MUI components using the [grey](https://mui.com/material-ui/customization/color/#2014-material-design-color-palettes) palette. I also put in a minor performance fix for probably unintended searches.

image

PR #802, opened 2024-07-10
Correct loglevel in lms_metadata to reduce logging noise

rtertiaer:
This PR corrects our log level handling in `lms_metadata.py`. Prior, it would always reset logging to `logging.DEBUG`, which is problematic for the underlying `urllib` which has its own `sys.stdout` stream handlers. This change was prompted by multiple bugs and troubleshooting scenarios not having enough logging because of the noisy journal being rotated often.

### Checklist


klay2000 (approved):
:+1:

PR #804, opened 2024-07-10
move state out of software directory

### LINKED ISSUE #785: Stop storing state next to software
This may be just storing house.json elsewhere, or storing both in places other than /home, or state in /home/pi/.config/amplipi

rtertiaer:
This PR moves all state out of the software directory. I'd like to run this a handful more times, but it ran fine the first time. I'm also seeking feedback on it earlier rather than later, particularly the `get_folder` changes. I added some tests around it to make it a smidge less brittle, hopefully. Unfortunately, most of the changes live in `configure.py` instead, which is extremely stateful and hard to test in an automated fashion.

This PR closes #785 

### Checklist

* [x] Have you written tests for your changes?


rtertiaer:
This has functioned as expected a couple of times now. Let me know if that's insufficient.

linknum23 (commented):
Looks like all of the config ported over nicely, even the memory mapped directories:

```
amplipi/config    102400       8    102392   1% /home/pi/.config/amplipi/srcs
amplipi/web       102400       0    102400   0% /home/pi/.config/amplipi/web/generated
```

A couple of things got missed:
- streams/lms_metadata.py (still references the old location, why was this hardcoded?)
- scripts/cleanup (tries to remove the old location)

I would also check the metadata from the streams to ensure they still work with these changes. I don't see anything else obvious in `streams/` though

It probably also makes sense to update the project.toml to include all files left in the repos config directory, currently each one needs to be whitelisted.

rtertiaer:
* I updated cleanup
* Unless I'm not seeing what you're seeing, the LMS metadata script has the new path
* I changed the config/ packaging to not be a whitelist, which more accurately matches the default `configure.py` during upgrade case.

pyproject.toml
```diff
@@ -30,13 +30,7 @@ include = [
...
   "requirements.txt",
   "AUTHORS",
-  "config/*.conf",
-  "config/bluetooth/*",
-  "config/*.rules",
-  "config/boot_config.txt",
-  "config/support_tunnel_crontab",
-  "config/support_group_sudoers",
-  "config/crontab",
+  "config/**",
```
linknum23, line 33: :ring_buoy:

scripts/cleanup
```diff
@@ -64,7 +64,7 @@ fi
 
 echo -e "reverting amplipi back to stock configuration"
 ssh $user_host "systemctl --user stop amplipi"
-ssh $user_host "rm -f amplipi-dev/house.json*"
+ssh $user_host "rm ~/.config/amplipi/*"  # intentionally avoids deleting folders used as mountpoints
```
linknum23, line 67: Will this remove authentication?

linknum23 (approved):

Issue #798, opened 2024-07-08
Version-checking in both the updater and main app report incorrect latest


rtertiaer:
My 0.4.0 box insists there's a new update, 0.4.0 - but only upon refresh. 

The backend reports the latest is still 0.3.6 in the /api endpoint. This might just take a second, but I'd like to not lose track of that so I'm reporting it here as well.

linknum23:
At the moment the latest release is checked once a day at 5am EST. We could probably increase the frequency on this. 

To actually fix this specific issue, we may want to run the `check_release` script after an update.

PR #813, opened 2024-07-12
check-release on reboot

rtertiaer:
This PR runs `check-release` on reboot, which accomplishes running post-upgrade too. Funnily enough, the crontab wasn't being packaged up by pyproject; this fixes this as well.

This tackles one half of #798 ; it may prevent all of it, but it's hard to tell when running on a non-release branch at the moment. I'll track that bug separately.

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ include = [
   "config/boot_config.txt",
   "config/support_tunnel_crontab",
   "config/support_group_sudoers",
+  "config/crontab",
   "debs/**/*",
   "fw/*.sh", # no need to add full source tree
   "fw/bin/*",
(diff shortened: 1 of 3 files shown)


klay2000 (approved):
wow, that's not how I thought getting the latest release worked...

pyproject.toml
```diff
@@ -36,6 +36,7 @@ include = [
   "config/boot_config.txt",
   "config/support_tunnel_crontab",
   "config/support_group_sudoers",
+  "config/crontab",
```
linknum23, line 39: Wait how was this working before?

rtertiaer: `pyproject.toml` only really gets used for `./scripts/deploy` to ship things to the AmpliPi - in that case, it simply didn't. However, regular upgrades download a tarball from github that do not do this whitelisting, and thus any updates from a crontab would still get intsalled. (Don't look too closely at who last edited the cronttab before these changes :eyes: )

PR #814, opened 2024-07-12
Fix display service on authenticated APIs

rtertiaer:
This PR fixes the display service on authenticated APIs, fetching the first user's session key directly from storage and using it in requests to the API. This methodology will have to change when we graduate beyond using a single user, but so will many other things around authentication as well. 

This has been tested against both e-ink and tft displays. I added a function to a shared module, so I've revisited the tests as well and souped them up.

### Checklist


klay2000 (approved):
lgtm,
Sorry, started reviewing this earlier, got pulled away, and just remembered it.

PR #816, opened 2024-07-15
StreamTitleMarquee

### LINKED ISSUE #555: Player page stream/source title
image
image

This is more of a nitpick issue
The source title on the player page is sometimes multiple lines, which pushes all the content down and you can't see the volume bar without scrolling.
My understanding of our purpose as we designed the webapp was to have absolute minimal scrolling, with the only instance of that on the player page being if you've expanded the volume dropdown. I'd suggest having dynamic font sizes so that the title always takes up the same amount of space
This may also demand a character limit for those titles so that you aren't in a position where text is wholly illegible due to text size

SteveMicroNova:
Add CustomMarquee for stream names on player page

Closes #555

### Checklist

### DIFF
--- a/web/src/components/SongInfo/SongInfo.scss
+++ b/web/src/components/SongInfo/SongInfo.scss
@@ -15,7 +15,6 @@
 
 .album-name {
   margin: 0.1rem;
-  font-size: 0.75rem;
   @include general.header-font;
   color: general.$text-color;
   font-size: 0.75rem;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   * Add hoverover effects on unselected player cards
   * Handle missing source selection
   * Autoselect newly created sources
+  * Add marquee support to stream names in player view
 * System
   * Reduce noisy logging from lms_metadata.py
   * Fix install of FM radio software `redsea`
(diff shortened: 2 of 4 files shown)


SteveMicroNova:
Through the power of having multiple tabs open, I've created a single video for both before and after.
Before comes before after, which is after the before.
https://github.com/user-attachments/assets/8200cb9a-dc59-4177-9a25-b754b3efc9c7

This isn't solving our issue of needing to scroll on that page, but we're a touch closer now

web/src/components/SongInfo/SongInfo.scss
```diff
@@ -15,7 +15,6 @@
 
 .album-name {
   margin: 0.1rem;
-  font-size: 0.75rem;
   @include general.header-font;
   color: general.$text-color;
   font-size: 0.75rem;
```
SteveMicroNova, line 20: Noticed that we were defining the same font-size twice

SteveMicroNova:
I'm being reminded that a byproduct of marquee is making the containing div twice as wide as it needs to be and then left aligning text, which now causes shorter stream names to look a little off:
image

On mid length and long names, it looks fine:
image
image

SteveMicroNova:
Lincoln has signaled that he both doesn't have time nor a deep enough understanding of React to give this a proper review, requesting review from Ryan to get to when he's back from vacation

rtertiaer (approved):
seems legit :sunglasses:

PR #820, opened 2024-07-17
Added support for mounting and reading USB sticks

mjustian:
### Checklist

image
image


amplipi/streams.py
```diff
@@ -1321,6 +1321,216 @@ def info(self) -> models.SourceInfo:
     return source
 
 
+MUSIC_EXTENSIONS = ('.mp3', '.wav', '.aac', '.m4a', '.flac', '.aiff')
+
+
+class USBStick(BaseStream, Browsable):
```
linknum23, line 1327: I would suggest making this a persistent stream so that an announcement can be played over it.

linknum23:
I know we added some support for USB with LMS does this have any conflicts with that? You may want to check with @SteveMicroNova

tests/test_rest.py
```diff
@@ -1278,7 +1278,7 @@ def test_play_media(client):
   assert rv.status_code != HTTPStatus.OK, 'Should require media field'
   # default no source
   rv = client.post('/api/play', json={'media': nasa_audio})
-  assert rv.status_code != HTTPStatus.OK, print(rv.text)
+  assert rv.status_code == HTTPStatus.OK, print(rv.text)
```
linknum23, line 1281: You may want to make these separate fixes for play into another PR

SteveMicroNova:
> I know we added some support for USB with LMS does this have any conflicts with that? You may want to check with @SteveMicroNova

Murphy talked to me about this and piggybacked off of that work, that's the udisks2-listener.service and udisks2-listener.sh files that were edited
He even went and fixed that implementation, when we got rid of the the desktop we lost the default installed service I was depending on (udisks2) so he's added that to our apt installs in configure.py and recreated the settings for it

SteveMicroNova (commented):
I'd suggest changing the vocabulary away from "USB Stick" and towards "Media Device" to cover the bases of having something like an external hard drive (not a usb 'stick' directly) or a NAS (once we get to that point)

amplipi/streams.py
```diff
@@ -1321,6 +1321,216 @@ def info(self) -> models.SourceInfo:
...
+    except Exception:
+      pass
+
+    if self.state == 'playing' and self.playing in self.song_list and self.ended and self.song_index < len(self.song_list) - 1 and self._ended_timeout < datetime.datetime.now():
+      self.next_song()
+    elif self.ended and (self.song_index >= len(self.song_list) - 1 or self.playing not in self.song_list) and self._ended_timeout < datetime.datetime.now():
+      self.state = 'paused'
+    return source
+
+  def browse(self, parent=None) -> List[models.BrowsableItem]:
```
linknum23, line 1507: How does this work when two people are browsing the same stream?

mjustian: It does not support that at this time, and supporting that would require the ability to track which session the browse request came from.

amplipi/streams.py
```diff
@@ -1488,7 +1486,7 @@ def info(self) -> models.SourceInfo:
                                img_url=img,
                                supported_cmds=self.supported_cmds,
                                type=self.stream_type)
-    src_config_folder = f"{utils.get_folder('config')}/srcs/{self.src}"
+    src_config_folder = f"{utils.get_folder('config')}/srcs/v{self.src}"
```
linknum23, line 1489: I think this config should be moved to the corresponding virtual source directory.

mjustian:
> I'd suggest changing the vocabulary away from "USB Stick" and towards "Media Device" or "Media Player" to cover the bases of having something like an external hard drive (not a usb 'stick' directly) or a NAS (once we get to that point)

I changed it to "Media Device"

linknum23:
Is this ready for review again?

mjustian:
> Is this ready for review again?

I believe so

amplipi/ctrl.py
```diff
@@ -533,11 +533,21 @@ def _check_latest_release(self) -> str:
...
+      if '/media' in partition.mountpoint and 'rootfs' not in partition.mountpoint:
+        usb_drives.append(str(partition.mountpoint))
+    return usb_drives
+
   def _update_sys_info(self, throttled=True) -> None:
     """Update current system information"""
     if self.status.info is None:
       raise Exception("No info generated, system in a bad state")
     self.status.info.online = self._online_cache.get(throttled)
+    self.status.info.connected_drives = self.get_usb_drives()
```
linknum23, line 550: Can you use a TimeBasedCache for this? This function gets called for every GET...

linknum23: See these for more info:

```
 self._online_cache = utils.TimeBasedCache(self._check_is_online, 5, 'online')
    self._latest_release_cache = utils.TimeBasedCache(self._check_latest_release, 3600, 'latest release')
```

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "amplipi"
-version = "0.4.0"
+version = "0.4.1"
```
linknum23, line 3: oops

amplipi/streams.py
```diff
@@ -1321,6 +1323,216 @@ def info(self) -> models.SourceInfo:
...
+      self.url = kwargs['url']
+      reconnect_needed = True
+    if reconnect_needed and self.is_activated():
+      self.restart()
+
+  def _activate(self, vsrc: int, remake_list: bool = True):
+    """ Connect a short run VLC process with audio output to a given audio source """
+    logger.info(f'connecting {self.name} to {vsrc}...')
+
+    if not self.mock and vsrc is not None:
```
linknum23, line 1365: vsrc is known to be not none as declared by the function signature. Is there a reason to doubt it?

amplipi/streams.py
```diff
@@ -1321,6 +1323,216 @@ def info(self) -> models.SourceInfo:
...
+        source.track = self.playing.split('/')[-1]
+        self.ended = data['state'] == 'ENDED'
+    except Exception:
+      pass
+
+    if self.state == 'playing' and self.playing in self.song_list and self.ended and self.song_index < len(self.song_list) - 1 and self._ended_timeout < datetime.datetime.now():
+      self.next_song()
+    elif self.ended and (self.song_index >= len(self.song_list) - 1 or self.playing not in self.song_list) and self._ended_timeout < datetime.datetime.now():
+      self.state = 'paused'
+    return source
```
linknum23, line 1504: Does this assume that this info method is polled at a specific interval? It is reasonably possible for this not to have any clients polling.

amplipi/streams.py
```diff
@@ -1321,6 +1323,216 @@ def info(self) -> models.SourceInfo:
...
+      self.restart()
+
+  def _activate(self, vsrc: int, remake_list: bool = True):
+    """ Connect a short run VLC process with audio output to a given audio source """
+    logger.info(f'connecting {self.name} to {vsrc}...')
+
+    if not self.mock and vsrc is not None:
+      # Make all of the necessary dir(s)
+      src_config_folder = f"{utils.get_folder('config')}/srcs/v{vsrc}"
+      os.system(f'mkdir -p {src_config_folder}')
```
rtertiaer, line 1368:
```suggestion
      os.makedirs(src_config_folder)
```

amplipi/streams.py
```diff
@@ -1321,6 +1323,216 @@ def info(self) -> models.SourceInfo:
...
+    for directory in self.directory_list:
+      browsables.append(models.BrowsableItem(name=directory.split('/')[-1], playable=True, parent=False, id=id, img='static/imgs/folder.png'))
+      id += 1
+    for song in self.song_list:
+      browsables.append(models.BrowsableItem(name=song.split('/')[-1], playable=True, parent=False, id=id, img='static/imgs/note.png'))
+      id += 1
+    return browsables
+
+  def play(self, id):
+    logger.info(f'id is {id}, local directory is {self.local_directory} and top directory is {self.directory}')
```
rtertiaer, line 1521:
```suggestion
    logger.debug(f'id is {id}, local directory is {self.local_directory} and top directory is {self.directory}')
```

rtertiaer (commented):
my security spidey senses are tingling; is it possible to traverse directories outside of `/media` using relative pathing (ie `../../../etc/passwd`)? I didn't see any place that prevents that at the moment - I'd love to see a check to ensure the resolved absolute path is still in `directory` (or maybe it's `local_directory`.) this is important both for the `make_song_list` case and also for the playing a song case.

(some folks would probably implement that closer to the API, but I personally think designing this backend so it doesn't have to be "used correctly" to be safe is a better bet.)

amplipi/models.py
```diff
@@ -996,7 +997,8 @@ class Config:
                 "git_dirty": False,
               }
             ],
-            'stream_types_available': ['bluetooth', 'fmradio']
+            'stream_types_available': ['bluetooth', 'fmradio'],
+            'connected_drives': [],
```
rtertiaer, line 1001: might be helpful to show something here. from this alone, it's unclear if it will show the mountpoint, or the device itself (i suspect i know the answer, but that's not clear.)

rtertiaer (commented):
sorry, i submitted too early.

where'd you get the icons? how are they licensed?

mjustian:
> sorry, i submitted too early.
> 
> where'd you get the icons? how are they licensed?

These are material icons, same as the rest of the UI.

amplipi/streams.py
```diff
@@ -1321,6 +1323,224 @@ def info(self) -> models.SourceInfo:
...
+                               supported_cmds=self.supported_cmds,
+                               type=self.stream_type)
+    src_config_folder = f"{utils.get_folder('config')}/srcs/v{self.src}"
+    loc = f'{src_config_folder}/currentSong'
+    try:
+      with open(loc, 'r', encoding='utf-8'):
+        source.track = self.playing.split('/')[-1]
+    except Exception:
+      pass
+    return source
```
linknum23, line 1507: As we discussed IRL the next track playback depends on the system being polled. Shouldn't this happen in a separate thread or process?

linknum23 (commented):
These changes look good, but I think you are missing the more significant change we discussed on Friday: moving metadata/command/next track handling in a separate thread or process. This would decouple the actual playback from the rest.

linknum23:
Alright. I'm looking at this again because I think this statement is probably wrong. This stream does not work how I was expecting but it does appear to work. I'll check through this again.

> These changes look good, but I think you are missing the more significant change we discussed on Friday: moving metadata/command/next track handling in a separate thread or process. This would decouple the actual playback from the rest.

amplipi/streams.py
linknum23: Disregard this for now.

scripts/99-udisks2.rules
linknum23, line 1: I would expect 10-udisks.pkla and this to be in the config directory.  These are not directly executable...

linknum23: Also can you add a newline to this file?

linknum23 (approved):
This latest iteration looks good. Please wait on  @rtertiaer to merge.

amplipi/app.py
```diff
@@ -512,43 +512,48 @@ def exec_command(cmd: models.StreamCommand, ctrl: Api = Depends(get_ctrl), sid:
...
 
 
-@api.get('/api/streams/{sid}/{pid}/browse', tags=['stream'])
-def browse_stream_child(ctrl: Api = Depends(get_ctrl), sid: int = params.StreamID, pid: int = params.ParentID) -> models.BrowsableItemResponse:
-  """ Browse the children of a media item in the current stream """
+@api.post('/api/streams/browser/{sid}/browse', tags=['stream'])
+def browse_stream(selection: Optional[models.BrowserSelection] = None, ctrl: Api = Depends(get_ctrl), sid: int = params.StreamID) -> models.BrowsableItemResponse:
+  """ Browse the top level children of the current stream's media """
   stream = ctrl.streams[sid]
+  logger.info("SOMEONE CALLED BROWSE_STREAM!!!")
```
rtertiaer, line 531:
```suggestion

```

amplipi/models.py
```diff
@@ -1375,3 +1392,423 @@ class Config:
         }
       ]
     }
+
+
+class PlayItemResponse(BaseModel):
+  directory: Optional[str]  # Directory that the browser is in
+  status: Status
```
rtertiaer, line 1399: jw - why do we return status here? it's huge

rtertiaer (changes requested):
real excellent stuff! i have one final ask for you :sunglasses: 

can you write a couple of tests that demonstrate/ensure the REST endpoints cannot browse and play outside their defined media dir?

amplipi/streams.py
rtertiaer: note for the future: in the latest rendition of this code, two people can browse at the same time :sunglasses:

amplipi/models.py
rtertiaer: context: it's because that was the prior interface for this type of response, and we're retaining backwards compatibility

rtertiaer (approved):
thanks for all your hard work this summer!! I'll pick up the tests, in my infinite free time :joy:

PR #822, opened 2024-07-18
Add selectActiveSource to settings page, update documentation

SteveMicroNova:
Adding a little polish to #788 that I noticed would've been nice to have after it was merged

### Checklist

### DIFF
--- a/web/src/pages/Settings/Settings.jsx
+++ b/web/src/pages/Settings/Settings.jsx
@@ -23,6 +23,7 @@ import ListItem from "@/components/List/ListItem/ListItem";
 import List from "@/components/List/List";
 import { IsMobileApp, IsSaved, SaveURL, UnsaveURL } from "@/utils/MobileApp";
 import Badge from "@mui/material/Badge";
+import selectActiveSource from "@/utils/selectActiveSource";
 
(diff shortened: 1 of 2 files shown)


web/src/pages/Settings/Settings.jsx
SteveMicroNova, line 1: Added the function to the settings page so that the Player tab doesn't randomly disappear due to other user's interactions

web/src/pages/Settings/Settings.jsx
```diff
@@ -89,6 +90,8 @@ const Settings = ({ openPage }) => {
     const is_streamer = useStatusStore((s) => s.status.info.is_streamer);
     const [isSavedUrl, setIsSavedUrl] = React.useState(IsSaved());
 
+    selectActiveSource();
```
linknum23, line 93: Should we add a comment here explaining why this needs to be here?

linknum23:
```suggestion
    // Make sure player tab doesn't just disappear when navigating to the settings page
    selectActiveSource();
```

linknum23 (approved):

linknum23:
Untested but I think this makes sense.

Issue #741, opened 2024-06-07
System log persistence during important events


rtertiaer:
We do not save the system log right now, largely to save on disk io. We should have a mechanism to save the system log during important events though - when a user requests support, during system upgrades, etc.

rtertiaer:
It would also be excellent to select the actual log level, or at least turn on debug logging.

SteveMicroNova:
> It would also be excellent to select the actual log level, or at least turn on debug logging.

Is this a request to add a separate setting for the log level on the frontend, to set the default log level to debug, or to ensure that the log level is debug when the boolean is true?

rtertiaer:
This is a request to either add:
* a drop down to select the log level
* or a boolean to turn on DEBUG logging

I had to swap the log level in `lms_metadata.py` to [something less verbose than DEBUG already](https://github.com/micro-nova/AmpliPi/pull/802) because it was absolutely inundating and wiping out the system log - but it would be nice to be able to turn that back on when necessary.

PR #828, opened 2024-07-24
Log Persistence

SteveMicroNova:
Supports #741 by adding a new tab to the updater called "Admin Settings" that contains the password reset stuff as well as the option to toggle log persistence

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   * Add error handling on browser page for instances where the selected stream isn't browsable
   * Add scrollbars to tall modals
   * Change how events are handled with Modals to reduce accidental closures
+  * Rename Updater to Admin Panel
+  * Add "Admin Settings" tab to the Updater/Admin Panel
+    * Add toggleable option to persist system logs to Admin Settings
+    * Move "Set password" page to be accessible via Admin Settings tab
 * System
   * Make update process properly report errors
 
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -226,7 +226,7 @@ def shutdown():
   """
   # preemptively save the state (just in case the shutdown procedure doesn't invoke a save)
   get_ctrl().save()
-  # start the shutdown process and returning immediately (hopeully before the shutdown process begins)
+  # start the shutdown process and returning immediately (hopefully before the shutdown process begins)
   Popen('sleep 1 && sudo systemctl poweroff', shell=True)
 
 
--- a/amplipi/ctrl.py
+++ b/amplipi/ctrl.py
@@ -222,7 +222,7 @@ class Api:
     except Exception as exc:
       logger.exception("Error setting is_streamer flag: {exc}")
 
-    # determine if we're in LMS mode, based on a file
+    # determine if we're in LMS mode, based on a file existing
     lms_mode_path = Path(defaults.USER_CONFIG_DIR, 'lms_mode')
     if lms_mode_path.exists():
       logger.info("lms mode")
--- a/amplipi/updater/asgi.py
+++ b/amplipi/updater/asgi.py
@@ -36,6 +36,8 @@ import pathlib
 import shutil
 import asyncio
 
+import configparser
+
 # web framework
 import requests
 from fastapi import FastAPI, Request, File, UploadFile, Depends, APIRouter, Response
@@ -48,7 +50,8 @@ import uvicorn
 # pylint: disable=no-name-in-module
 from pydantic import BaseModel
 
-from ..auth import CookieOrParamAPIKey, router as auth_router, set_password_hash, unset_password_hash, NotAuthenticatedException, not_authenticated_exception_handler, create_access_key
+from ..auth import CookieOrParamAPIKey, router as auth_router, set_password_hash, unset_password_hash,\
+  NotAuthenticatedException, not_authenticated_exception_handler, create_access_key
 
 app = FastAPI()
 router = APIRouter(dependencies=[Depends(CookieOrParamAPIKey)])
@@ -60,6 +63,72 @@ logger.addHandler(sh)
 app.add_exception_handler(NotAuthenticatedException, not_authenticated_exception_handler)
 
 sse_messages: queue.Queue = queue.Queue()
+
+
+def validate_logging_ini():
+  """Fallback in case the ini file or any individual header  doesn't exist, set to default settings. Only really comes up during tests."""
+  tmp = '/tmp/logging.ini.tmp'
+  ini = '/var/log/logging.ini'
+  conf = configparser.ConfigParser(strict=False, allow_no_value=True)
+
+  with open(tmp, "+w", encoding="utf-8") as file:
+    if os.path.exists(ini):
+      conf.read(ini)
+    else:
+      conf.read(file)
+
+    if not conf.has_section("logging"):
+      conf.add_section("logging")
+
+    if not conf.has_option("logging", "auto_off_delay"):
+      conf.set("logging", "auto_off_delay", "14")
+    auto_off_delay = conf.get("logging", "auto_off_delay", fallback="14")
+    if not auto_off_delay.isdigit() and bool(re.fullmatch(r'\d*\.\d+', auto_off_delay)):
+      # regex to check decimal state, this would lead to "123.45" and ".45" being true but not "123."
+      # Exclude anything that isdigit() as to not overwrite valid user settings
+      rounded = round(float(auto_off_delay)) if round(float(auto_off_delay)) > 0 else 1  # Avoid instances where it could be zero as to not set the "do not deactivate" setting
+      conf.set("logging", "auto_off_delay", str(rounded))
+    elif not auto_off_delay.isdigit():  # Cannot be merged with the first check in an OR case as valid regex catches would be intercepted by that
+      conf.set("logging", "auto_off_delay", "14")
+    conf.write(file)
+
+  subprocess.run(['sudo', 'mv', tmp, ini], check=True)
+
+
+def validate_journald_conf():
+  """Fallback in case the config file or any individual header doesn't exist, set to default settings"""
+  tmp = '/tmp/journald.conf.tmp'
+  conf = '/etc/systemd/journald.conf'
+  confparse = configparser.ConfigParser(strict=False, allow_no_value=True)
+
+  with open(tmp, "+w", encoding="utf-8") as file:
+    if os.path.exists(conf):
+      confparse.read(conf)
+    else:
+      confparse.read(file)
+
+    if not confparse.has_section("Journal"):
+      confparse.add_section("Journal")
+
+    if not confparse.has_option("Journal", "Storage") or confparse.get("Journal", "Storage") not in ("volatile", "persistent"):
+      # While volatile is the system's default, having that value either not exist or be invalid points to there being a system error that caused it to get that way
+      # Set to persistent just in case
+      confparse.set("Journal", "Storage", "persistent")
+
+    # Set everything else to default while preserving user settings
+    if not confparse.has_option("Journal", "SyncIntervalSec"):
+      confparse.set('Journal', 'SyncIntervalSec', '30s')
+    if not confparse.has_option("Journal", "SystemMaxUse"):
+      confparse.set('Journal', 'SystemMaxUse', '64M')
+    if not confparse.has_option("Journal", "RuntimeMaxUse"):
+      confparse.set('Journal', 'RuntimeMaxUse', '64M')
+    if not confparse.has_option("Journal", "ForwardToConsole"):
+      confparse.set('Journal', 'ForwardToConsole', 'no')
+    if not confparse.has_option("Journal", "ForwardToWall"):
+      confparse.set('Journal', 'ForwardToWall', 'no')
+
+    confparse.write(file)
+  subprocess.run(['sudo', 'mv', tmp, conf], check=True)
 
 
 class ReleaseInfo(BaseModel):
@@ -93,6 +162,81 @@ except FileNotFoundError:
   pass
 except Exception as e:
   logger.exception(f'Error loading identity file: {e}')
+
+
+class Persist_Logs(BaseModel):
+  """Basemodel that consists of a bool and int, used to change different config files around the system via POST /settings/persist_logs"""
+  persist_logs: bool
+  auto_off_delay: int
+
+
+@router.get("/settings/persist_logs")
+def get_log_persist_state():
+  """
+  Checks /etc/systemd/journald.conf to find if the current storage setting is persistent and returns a bool
+  Note that returning false doesn't necessarily mean that logs are set to volatile, and could just mean that the config file is missing the line being read
+  """
+  validate_journald_conf()
+  journalconf = configparser.ConfigParser(strict=False, allow_no_value=True)
+  journalconf.read('/etc/systemd/journald.conf')
+
+  validate_logging_ini()
+  logconf = configparser.ConfigParser(strict=False, allow_no_value=True)
+  logconf.read('/var/log/logging.ini')
+
+  # Fallback set is the default value of the Storage variable under the Journal header of the conf file
+  # Used when the variable cannot be read but the file itself can (implying that the variable is missing, and should be set to a default)
+  ret = Persist_Logs(persist_logs=journalconf.get("Journal", "Storage", fallback="volatile") == "persistent", auto_off_delay=logconf.get("logging", "auto_off_delay", fallback="14"),)
+  return ret
+
+
+@router.post("/settings/persist_logs")
+def toggle_persist_logs(data: Persist_Logs):
+  """Toggles the option within journald to save logs to memory or storage, and sets the length of time before that setting is reset to volatile"""
+  try:
+    # Just in case
+    validate_logging_ini()
+    validate_journald_conf()
+
+    state = get_log_persist_state()
+
+    if state.persist_logs != data.persist_logs:
+      journalconf = '/etc/systemd/journald.conf'
+      journaltmp = '/tmp/journald.conf.tmp'
+      journal = configparser.ConfigParser(strict=False, allow_no_value=True)
+      journal.read(journalconf)
+
+      if not journal.has_section("Journal"):
+        journal.add_section('Journal')
+
+      # goal_value is true if you wish to turn persistent logging on and false if you wish to turn it off
+      journal.set('Journal', 'Storage', 'persistent' if data.persist_logs else 'volatile')
+
+      with open(journaltmp, 'w', encoding="utf-8") as conf_file:
+        journal.write(conf_file)
+
+      subprocess.run(['sudo', 'mv', journaltmp, journalconf], check=True)
+      subprocess.run(['sudo', 'systemctl', 'restart', 'systemd-journald'], check=True)
+      logger.info(f"persist_logs set to {data.persist_logs}")
+    else:
+      logger.info("persist_logs unchanged")
+
+    if state.auto_off_delay != data.auto_off_delay:
+      logconf = '/var/log/logging.ini'
+      logtmp = '/tmp/logging.ini.tmp'
+      log = configparser.ConfigParser(strict=False, allow_no_value=True)
+      log.read(logconf)
+      log.set('logging', 'auto_off_delay', f"{data.auto_off_delay}")  # Accept auto_off_delay as an int for type checking, parse to str for configParser validity
+      with open(logtmp, 'w', encoding='utf-8') as file:
+        log.write(file)
+      subprocess.run(['sudo', 'mv', logtmp, logconf], check=True)
+      logger.info(f"auto_off_delay set to {data.auto_off_delay}")
+
+    else:
+      logger.info("auto_off_delay unchanged")
+  except Exception as exc:
+    logger.exception(str(exc))
+    return 500
 
 
 @router.get('/update')
--- a/amplipi/updater/static/index.html
+++ b/amplipi/updater/static/index.html
@@ -11,6 +11,7 @@
     <!-- Font Awesome CSS -->
     <link rel='stylesheet' href='https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.min.css'
       onerror="this.onerror=null;this.href='static/css/libs/font-awesome/5.15.1/all.min.css';">
+      <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
     <!-- Bootstrap core CSS -->
     <link rel='stylesheet' href='https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.5.0/css/bootstrap.min.css'
       onerror="this.onerror=null;this.href='static/css/libs/twitter-bootstrap/4.5.0/bootstrap.min.css';">
@@ -28,22 +29,22 @@
       </p>
       <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1>
       <p class="lead mb-4">
-        Update your AmpliPi device
+        Update & configure your AmpliPi device
       </p>
       <div class="card">
         <div class="card-header">
           <ul class="nav nav-tabs card-header-tabs" id="myTab" role="tablist">
             <li class="nav-item">
-              <a class="nav-link active" id="latest-update-tab" data-toggle="tab" href="#latest-update" role="tab" aria-controls="latest-update" aria-selected="true">Latest</a>
+              <a class="nav-link active" id="latest-update-tab" data-toggle="tab" href="#latest-update" role="tab" aria-controls="latest-update" aria-selected="true">Latest Release</a>
             </li>
             <li class="nav-item">
               <a class="nav-link" id="older-update-tab" data-toggle="tab" href="#older-update" role="tab" aria-controls="older-update" aria-selected="false">Older Releases</a>
             </li>
             <li class="nav-item">
-              <a class="nav-link" id="manual-update-tab" data-toggle="tab" href="#manual-update" role="tab" aria-controls="manual-update" aria-selected="false">Custom</a>
+              <a class="nav-link" id="manual-update-tab" data-toggle="tab" href="#manual-update" role="tab" aria-controls="manual-update" aria-selected="false">Custom Update</a>
             </li>
             <li class="nav-item">
-              <a class="nav-link" id="set-password-tab" data-toggle="tab" href="#set-password" role="tab" aria-controls="set-password" aria-selected="false">Set Password</a>
+              <a class="nav-link" id="admin-settings-tab" data-toggle="tab" href="#admin-settings" role="tab" aria-controls="admin-settings" aria-selected="false">Admin Settings</a>
             </li>
             <li class="nav-item">
               <a class="nav-link" id="support-tunnel-tab" data-toggle="tab" href="#support-tunnel" role="tab" aria-controls="support-tunnel" aria-selected="false">Support Tunnel</a>
@@ -52,6 +53,7 @@
         </div>
         <div class="card-body text-center">
           <div class="tab-content" id="myTabContent">
+
             <div class="tab-pane fade show active" id="latest-update" role="tabpanel" aria-labelledby="latest-update-tab">
               <div id='latest-update-name'>...</div>
               <div id='latest-update-desc' class="text-left"></div>
@@ -59,6 +61,7 @@
                 onclick="ui_start_software_update($('#latest-update').attr('data-url'), $('#latest-update').attr('data-version'));">
                 Start Update</a>
             </div>
+
             <div class="tab-pane fade" id="older-update" role="tabpanel" aria-labelledby="older-update-tab">
               <select class="custom-select" id="older-update-sel" onchange="ui_select_release(this);">
                 <!-- Releases populated from GH -->
@@ -69,24 +72,14 @@
                  onclick="ui_start_software_update($('#older-update-sel').val(), $('#older-update-sel').find(':selected').data('version'));">
                 Start Update</a>
             </div>
+
             <div class="tab-pane fade" id="manual-update" role="tabpanel" aria-labelledby="manual-update-tab">
               <div class="custom-file">
                 <input id="update-file-selector" type="file" accept=".gz" onchange="$('#submit-custom-update').removeClass('disabled');">
               </div>
               <a id="submit-custom-update" role="button" class="btn btn-lg btn-primary text-center mt-3 disabled" onclick="ui_upload_software_update();">Start Update</a>
             </div>
-            <div class="tab-pane fade text-left" id="set-password" role="tabpanel" aria-labelledby="set-password-tab">
-              <div class="set-password-dialog">
-                This form allows you to set a password for the web interface. If the password fields are empty, it will unset the password protection. This will only set a password for the web interface, not the system password. This action will also log out any existing sessions.
-              </div>
-              <div class="set-password form-group">
-                <label for="password" class="form-label">Password:</label>
-                <input id="password" class="form-control" type="password">
-                <label for="confirm-password" class="form-label">Confirm Password:</label>
-                <input id="confirm-password" class="form-control" type="password">
-              </div>
-              <a id="password" role="button" class="btn btn-lg btn-primary text-center mt-3" onClick="set_password();">Update Password</a>
-            </div>
+
             <div class="tab-pane fade text-left" id="support-tunnel" role="tabpanel" aria-labelledby="support-tunnel-tab">
               <div id="support-tunnel-dialog">
                 This dialog allows you to request a support tunnel. Once requested, you must give the tunnel ID and preshared key to support, who can be reached via <a href="mailto:support@micro-nova.com">support@micro-nova.com</a>.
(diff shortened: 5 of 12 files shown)


rtertiaer:
I'm of split mind between putting this in the main amplipi setup, and in the updater (which we should probably rename "admin" or similar.)

SteveMicroNova:
> I'm of split mind between putting this in the main amplipi setup, and in the updater (which we should probably rename "admin" or similar.)

@linknum23 Thoughts?

linknum23:
Sorry I didn't get a chance to look at this before my early weekend adventures.

Currently the only thing that modifies the logging is the updater so modifying it in amplipi as well feels awkward. However that would mean that this menu has to live in the updater/admin interface somewhere. The right choice is not obvious to me right now. That probably means we need a better interface for stuff like this.

scripts/configure.py
```diff
@@ -99,6 +99,7 @@
             'sudo systemctl restart rsyslog.service',
 
             'echo "reconfiguring journald to only log to RAM"',
+            'sudo chmod 666 /etc/systemd/journald.conf',
```
rtertiaer, line 102:
```suggestion
```

I feel pretty strongly we shouldn't modify permissions on this file, and instead use privilege escalation if we want to poke it (ie, "go through the front door" - use `sudo` or similar in a `subprocess`.)

SteveMicroNova:
> Currently the only thing that modifies the logging is the updater

Where at? I'm looking at the updater and I just see the latest update, a way to revert to previous updates, a way to upload custom updates, a way to set a password for the web interface, and the support tunnel. Nothing that has to do with logs as far as I can see

SteveMicroNova:
Marking this as ready to review as both people that have chimed in are uncertain as to if this belongs in the updater or not
If we solidly say "this belongs in the updater", it will be reverted to draft while I work on that

amplipi/app.py
```diff
@@ -229,6 +230,23 @@ def shutdown():
...
+      200: {}
+    }
+  )
+def log_persistance(ctrl: Api = Depends(get_ctrl)):
+  """Toggles log persistance by directly changing the journald.conf and restarting the service"""
+  try:
+    ctrl.toggle_persist_logs()
+  except Exception as e:
+    logger.error("persist_logs toggle failed!")
+    raise e
```
SteveMicroNova, line 247: I found reasons to call this function from within the Api object itself and so moved it there, this imports the Api object as ctrl and calls the function

amplipi/ctrl.py
```diff
@@ -229,6 +232,23 @@ def reinit(self, settings: models.AppSettings = models.AppSettings(), change_not
...
+      storage_value = confparse.get("Journal", "Storage")
+      if storage_value == "volatile":
+        self.persist_logs = False
+      elif storage_value == "persistent":
+        self.persist_logs = True
+      else:
+        self.toggle_persist_logs(False)
+    except Exception as exc:
+      logger.exception("Error reading persist_logs flag, setting to False (volatile logs). Exception: {exc}")
+      self.toggle_persist_logs(False)
```
SteveMicroNova, line 250: Attempts to read the config file to see what the current mode is. If it cannot be read, default to not persisting logs

amplipi/ctrl.py
```diff
@@ -418,6 +439,61 @@ def reinit(self, settings: models.AppSettings = models.AppSettings(), change_not
...
+      config.read(f'{conf_path}/journald.conf')
+
+      if 'Journal' not in config:
+        config.add_section('Journal')
+
+      goal_value: bool
+      if value is not None:
+        goal_value = value
+      else:
+        goal_value = not self.persist_logs
```
SteveMicroNova, line 460: Added the option to explicitly manually toggle the persistence rather than always flicking the switch to the inverse of the present setting

SteveMicroNova:
Requesting review from Ryan as he reviews all my things

Requesting review from Lincoln as he's already had input and I would like to continue to ask his opinion, but I do not need his review per se

Requesting review from Klayton to potentially have a tiebreaker for the "should the be in the updater or not?" issue

rtertiaer:
> Where at?
at a minimum [here](https://github.com/micro-nova/AmpliPi/blob/4536e181fd4c3fb18455928405f98e0ac36fdae6/scripts/configure.py#L102), called from the "updater side" of things.

rtertiaer:
my vote would be to put this in the updater, given the above.

klay2000:
I don't feel like this goes in the updater because it's not exactly related to updates, however I also don't feel like the regular Amplipi web app should be able to change something like this.  Given that the updater also has the support tunnel stuff I suppose there is some precedent for something like this to go there. Ultimately I think as was said before we need a better place for this kind of thing, but for now the updater is good enough.

SteveMicroNova:
https://github.com/user-attachments/assets/5784bc4e-b71b-4807-a947-2f6f587e699f

Log persistence has been moved to the updater page (now called admin) and plugged in properly
See video for changes and proof of function, or go to lamplipi now to see for yourself. Next steps are make sure you can actually access persisted logs from the logs page, and find where those are stored in the first place

SteveMicroNova:
> Next steps are make sure you can actually access persisted logs from the logs page, and find where those are stored in the first place

I've found the file they persist to and verified that the logs page can find them, the only thing is persistent logs _always_ persist, and will be on top of the pile even when you turn off persistent storage. We should garbage clean this a bit, the only question is on what basis.

My immediate ideas:
1. simply wipe them out the second log persistence is turned off
2. turn on a midnight cronjob to clear the persist logs directory of all files at midnight if persist logs is deactivated

linknum23:
Should we change the Settings name from "Admin" to "Admin / Updates" to maintain consistency with the old interface?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
   * Move system state out of running software directory
   * Fix updates on systems with an API password set
   * Fix the display service on systems with an API password set
+  * Add toggleable option to persist logs
```
SteveMicroNova, line 40: This might need to move to a different section now that it's no longer in the webapp

amplipi/updater/asgi.py
```diff
@@ -95,6 +98,80 @@ class ReleaseInfo(BaseModel):
...
+  except Exception as exc:
+    logger.error("persist_logs toggle failed!")
+    raise exc
+
+
+@router.delete("/settings/persist_logs")
+def delete_logs():
+  """Deletes everything in the persistent logs directory if logs are not set to persist"""
+  if not get_log_persist_state():
+    subprocess.run(['sudo', 'rm', '/var/logs/journal/*'], check=True)
```
SteveMicroNova, line 172: Nice and simple deletion endpoint, currently only called by a curl command inside of a bash file that's called by a cronjob at midnight

amplipi/updater/static/index.html
```diff
@@ -69,12 +71,14 @@ <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1
...
                 Start Update</a>
             </div>
+
             <div class="tab-pane fade" id="manual-update" role="tabpanel" aria-labelledby="manual-update-tab">
               <div class="custom-file">
                 <input id="update-file-selector" type="file" accept=".gz" onchange="$('#submit-custom-update').removeClass('disabled');">
               </div>
               <a id="submit-custom-update" role="button" class="btn btn-lg btn-primary text-center mt-3 disabled" onclick="ui_upload_software_update();">Start Update</a>
             </div>
+
```
SteveMicroNova, line 81: Added spacing between divs to make them human readable, my eyes simply couldn't differentiate the sections at a glance otherwise

amplipi/updater/static/index.html
```diff
@@ -105,6 +110,40 @@ <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1
...
+                  const checkbox = document.getElementById("persist_logs_checkbox");
+                  const response = await fetch("/settings/persist_logs", {
+                    method: "POST",
+                    headers: { 'Content-Type': 'application/json' },
+                    body: JSON.stringify({ "value": !checkbox.checked }) // Explicitly set to the opposite of the present setting to ensure that two users hitting the checkbox don't toggle it twice
+                  });
+                }
+
+                window.onload = getPersist;
+              </script>
```
SteveMicroNova, line 133: potentially belongs in the js folder, currently it lives here because it was easiest to develop stuff inline

scripts/configure.py
```diff
@@ -103,6 +103,12 @@
             'sudo systemctl enable systemd-journald.service',
             'sudo systemctl restart systemd-journald.service',
 
+            # Create directories and cronjobs for log persistence option
+            'sudo mkdir -p /var/log/journal',
+            'sudo systemd-tmpfiles --create --prefix /var/log/journal',
+            'sudo cp scripts/delete_persisted_logs.sh /usr/local/bin',
+            'sudo cp config/delete_persisted_logs_crontab /etc/cron.d/delete_persisted_logs',
```
SteveMicroNova, line 110: Separated this chunk into its own section for the final edition of this code rather than having them mixed in under the other sectioned chunks within this logging-handling script

CHANGELOG.md
SteveMicroNova: Especially because this is now getting into a release a version or two later 🥴

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 ## Upcoming Release
 * Web App
   * Fix internet radio search on IOS app
+* System
+  * Rename Updater to Updater/Admin Control Panel
+  * Add "Admin Settings" tab to the Updater/Admin Control Panel
+    * Add toggleable option to persist system logs to Admin Settings
+    * Move "Set password" page to be accessible via Admin Settings tab
```
SteveMicroNova, line 10: Unusual formatting as a few of these bullets fall under the umbrella of another, I can untab the latter two entries no problem if we dislike this

SteveMicroNova:
Revoking reviews requested and marking as draft, I've found more bugs in this PR due to working on a PR based on this one (I'm building out the admin settings tab with more options)

amplipi/updater/asgi.py
```diff
@@ -95,6 +98,80 @@ class ReleaseInfo(BaseModel):
...
+  Note that returning false doesn't necessarily mean that logs are set to volatile, and could just mean that the config file is missing the line being read
+  """
+  conf_path = '/etc/systemd'
+  config = configparser.ConfigParser(strict=False, allow_no_value=True)
+  config.read(f'{conf_path}/journald.conf')
+  return config.get("Journal", "Storage", fallback="") == "persistent"
+
+
+@router.post("/settings/persist_logs")
+def toggle_persist_logs(value: Optional[bool] = None):
```
rtertiaer, line 114: `value` seems to be unused

config/delete_persisted_logs_crontab
```diff
@@ -0,0 +1,3 @@
+# Recurring persist_log garbage collection function handled by cron
+# Installed by AmpliPi
+0 0 * * * /usr/local/bin/delete_persisted_logs.sh
```
rtertiaer, line 3: I would prefer we don't implement this. We already set `SystemMaxUse` to 64mb. journalctl does not need log rotation, it will remove the oldest entries first when it gets to 64mb.

amplipi/updater/asgi.py
rtertiaer: considering I don't want to remove logs via cronjob, is this necessary? I would argue not. It also feels like a somewhat privileged action - an arbitrary user on the network should not be able to hide evidence of malfeasance by hitting an unauthenticated API endpoint.

amplipi/updater/asgi.py
```diff
@@ -95,6 +98,80 @@ class ReleaseInfo(BaseModel):
...
+      logger.info("persist_logs activated!")
+
+    else:  # Reset config to default as seen in configure.py
+      config.set('Journal', 'Storage', 'volatile')
+      config.set('Journal', 'RuntimeMaxUse', '64M')
+      config.set('Journal', 'ForwardToConsole', 'no')
+      config.set('Journal', 'ForwardToWall', 'no')
+
+      config.remove_option('Journal', 'SyncIntervalSec')
+      config.remove_option('Journal', 'SystemMaxUse')
```
rtertiaer, line 154: this strategy is somewhat brittle; what happens when an update is installed?

config/delete_persisted_logs_crontab
SteveMicroNova: The problem being solved here is that once you stop persisting new logs, the logs you've already persisted never get wiped and will still be at the top of amplipi.local:19531/entrie.
I perceive that to be an issue and so made this cronjob that fires off a bash script, the bash script uses curl to call the `DELETE /settings/logs` endpoint, this endpoint conditionally clears out persisted logs if storage is set to volatile (aka if persistence is off)

amplipi/updater/asgi.py
SteveMicroNova: Haven't had any updates that contain this code to test with, I imagine it just resets back to volatile as is defined in `configure.py`

amplipi/updater/asgi.py
SteveMicroNova: It's only really used by the delete endpoint, I've also tried to use it with the frontend to catch an edgecase but that proved unnecessary and error prone

amplipi/updater/static/index.html
```diff
@@ -105,6 +110,36 @@ <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1
...
+
+              <div id="admin-settings-dialog">
+                This tab allows you to change settings relating to the core of AmpliPi, things that do not belong on the standard config page
+              </div>
+              <p class="">
+                <a class="btn btn-primary btn-sm text-center" id="set-password-button" data-toggle="tab" href="#set-password" role="tab" aria-controls="set-password" aria-selected="false">Set Password</a>
+              </p>
+              <h6>WARNING: leaving log persistence on for extended periods can damage your system and void your warranty</h6>
```
rtertiaer, line 137: is this true?

amplipi/updater/static/index.html
```diff
@@ -105,6 +110,36 @@ <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1
...
+                async function setPersist() {
+                  const checkbox = document.getElementById("persist_logs_checkbox");
+                  const response = await fetch("/settings/persist_logs", {method: "POST"});
+                }
+
+                window.onload = getPersist;
+              </script>
+
+              <div id="admin-settings-dialog">
+                This tab allows you to change settings relating to the core of AmpliPi, things that do not belong on the standard config page
```
rtertiaer, line 132:
```suggestion
                This tab allows you to change settings relating to the core of AmpliPi.
```

config/delete_persisted_logs_crontab
rtertiaer: 1) if the max filesize of logs is only ever 64mb, is that actually necessary? The reason why logs aren't saved to disk right now is to prevent excessive writes, not because the size on disk is excessive. I'm unconvinced of its necessity, and tend to think instead this does actions beyond what's "written on the can", which could be problematic.
2) if you wanted to clean the logs when toggling this, why would you choose a cronjob to do that rather than the code that's actually doing the toggling? if you're worried about the call taking too long, fork it to the background
3) the entries logs being at the top is definitely something for us to be aware of when we are getting & viewing logs. however, if someone has chosen to persist logs, would you not expect them to ... persist? they chose that for a reason, probably because they were having a problem.

imo we should probably persist the logs users have asked us to persist, or make it abundantly clear that toggling "persist logs" off will wipe any persisted logs.

amplipi/updater/asgi.py
rtertiaer: you're correct; I would consider that a bug with this implementation

amplipi/updater/static/index.html
SteveMicroNova: I've been lead to believe that persisting logs will lead to an incredible amount of reads and writes and wear out the drive, if a customer chooses to knowingly do that to themselves I can't imagine we'd cover that; I haven't passed this line by anyone yet however

That said, I don't actually track how long persist_logs has been active and so a worn storage device doesn't necessarily point towards any culprit

amplipi/updater/asgi.py
SteveMicroNova: We initialize Amplipi's with an image right?
If so, would an acceptable solution be simply ripping that chunk of `configure.py` out?

I struggle to think of a method where I could check if persistent logs were set on purpose vs being set because that's the system default prior to the first `configure.py` run

scripts/configure.py
```diff
@@ -98,11 +98,15 @@
...
-            'echo "reconfiguring journald to only log to RAM"',
-            r'echo -e "[Journal]\nStorage=volatile\nRuntimeMaxUse=64M\nForwardToConsole=no\nForwardToWall=no\n" | sudo tee /etc/systemd/journald.conf',
+            'echo "If this is first deploy, reconfiguring journald to only log to RAM"',
+            r'[ ! -d /var/log/journal ] && echo -e "[Journal]\nStorage=volatile\nRuntimeMaxUse=64M\nForwardToConsole=no\nForwardToWall=no\n" | sudo tee /etc/systemd/journald.conf',
             'sudo systemctl enable systemd-journald.service',
             'sudo systemctl restart systemd-journald.service',
 
+            # Create directories and cronjobs for log persistence option
+            'sudo mkdir -p /var/log/journal',
+            'sudo systemd-tmpfiles --create --prefix /var/log/journal',
```
SteveMicroNova, line 108: If a directory does not exist, set logs to volatile
This directory is then made immediately afterwards, meaning that the first portion only works on first deploy and will not wipe out the user's `persist_logs` settings

amplipi/updater/asgi.py
SteveMicroNova: Alternative: https://github.com/micro-nova/AmpliPi/pull/828/files#r1707793661

config/delete_persisted_logs_crontab
SteveMicroNova: Path of least resistance says I leave it as it is in the current version by not deleting logs, but I'll still answer questions for the context:

1. Maybe, only if you care about not needing to search for up to date logs when you looks at the logs page
2. Cronjob was just in case the logs were accidentally toggled back, or if someone didn't realize that turning that off would un-persist logs before they got the chance to collect them
3. Fair enough

amplipi/updater/static/index.html
rtertiaer: warranties are legally binding, and thus we should be explicit about what exactly voids the warranty. if turning this setting on voids it, great, but specifying an indeterminate amount of time is probably not ideal.

ime, running a raspberry pi for years with vanilla journalctl hasn't worn out my eMMC. It's a good idea to limit writes, but I hesitate to say we should void warranties about this. the default case is good for limiting writes, and if people want to turn it on they should be able to.

@linknum23 do you have thoughts on this one particular line?

amplipi/updater/asgi.py
rtertiaer: your "alternative" link is broken.

we create the image using `configure.py`, so directly ripping that chunk out will not work. i would consider storing this state somewhere in the config dir, and using that as inputs to both configure.py and `get_log_perist_state` to set the correct values, with an uninitialized default of `volatile`

SteveMicroNova: That link was meant to direct to this line in configure.py that only creates the file if it doesn't already exist:
`r'[ ! -d /var/log/journal ] && echo -e "[Journal]\nStorage=volatile\nRuntimeMaxUse=64M\nForwardToConsole=no\nForwardToWall=no\n" | sudo tee /etc/systemd/journald.conf',`

amplipi/updater/static/index.html
SteveMicroNova: > ime, running a raspberry pi for years with vanilla journalctl hasn't worn out my eMMC

This gives me the sense that the line is unnecessary as the warranty will be expired by the time it matters. I can probably forgo the warning at that rate as anyone touching this setting ought to understand what that entails and want logs to stick around; I can't imagine a user that doesn't know what logs are would go persisting them.

I'll await Lincoln's two cents before removing anything, though

linknum23: We need to find a simpler way to do this without the scary warning. Maybe only enabling logging for 2 weeks?

SteveMicroNova: Instead of changing the setting automatically, would it suffice to have the frontend occasionally pop a notification that's like "hey, look out, this setting is still on!" And have it be colored yellow instead of the green/red the other notifications are

Perhaps not due to the number of people that only use a custom API caller to interact with the device, but I feel Ryan's instinct (or we least what I've interpreted his instinct to be) that we should leave this high level control in the hands of the device owners rather than trying to know better than them is for the best. Inform where necessary, but leave them to do their thing without any bothersome automations getting in the way of them doing what they want to do with their own system(s).

linknum23:
I like the certainty of something that will turn off after some time but I
also get the need/want for full control. This really should only be a
temporary solution to help diagnose users' problems.

I wonder if there is a realistic timeframe that would work for most users?

For more permanent logging on the current hardware I would suggest hooking
this up to an external server or an attached USB key.

SteveMicroNova:
> I like the certainty of something that will turn off after some time but I also get the need/want for full control. This really should only be a temporary solution to help diagnose users' problems. I wonder if there is a realistic timeframe that would work for most users?

I've elected to shoot for a middleground where there's a "persist logs auto-deactivate delay" that can be set right alongside the persist logs status, defaults to 14 days from activation with the option to set to 0 to turn off the auto-deactivation

I've worked on that all day and I'm not yet ready to commit it as it's nonfunctioning atm, but I'm announcing that to be my intended solution and it seems pretty good so far imo

SteveMicroNova:
> I've elected to shoot for a middleground where there's a "persist logs auto-deactivate delay" that can be set right alongside the persist logs status

This took some working and reworking, but I think it's done
I initially had them be a separate string of endpoints, but then I noticed that the `persist_logs` bool and the `auto_off_delay` int were too closely related to keep it that way, rerouted all the logic to make them work within the same pydantic object for every endpoint and made a script and cronjob to run the script that makes the `auto_off_delay` actually matter. Will be self-reviewing for commentary shortly

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,47 @@
...
+    body = {
+      "persist_logs": True,
+      "auto_off_delay": delay,
+    }
+    logging.info(f"Persist logs will be automatically turned off in {delay} day(s)")
+  else:
+    body = {
+      "persist_logs": False,
+      "auto_off_delay": 14,  # Reset to default for next time
+    }
```
SteveMicroNova, line 38: If `auto_off_delay - 1` would equal 0, instead reset it to default and turn off persistence

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,47 @@
...
+    logging.info("Persist logs has been turned off automatically")
+
+  json_data = json.dumps(body)
+  subprocess.run([
+      'curl', '-X', 'POST',
+      '--header', 'Content-Type: application/json',
+      '--data', json_data,
+      'http://localhost:5001/settings/persist_logs',
+  ], check=True)
```
SteveMicroNova, line 47: Everything in through the front door, no editing the ini files directly from this script

SteveMicroNova:
image

amplipi/updater/asgi.py
```diff
@@ -95,6 +113,90 @@ class ReleaseInfo(BaseModel):
...
+  auto_off_delay: int
+
+
+@router.get("/settings/persist_logs")
+def get_log_persist_state():
+  """
+  Checks /etc/systemd/journald.conf to find if the current storage setting is persistent and returns a bool
+  Note that returning false doesn't necessarily mean that logs are set to volatile, and could just mean that the config file is missing the line being read
+  """
+  create_logging_ini()
```
rtertiaer, line 128: does this `GET` not set the `auto_off_delay` to 14 every time?

amplipi/updater/static/index.html
```diff
@@ -105,6 +110,71 @@ <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1
...
+              <!-- Log Persistence -->
+              <div id="set-password" style="display: flex; flex-direction: column; gap: 10px;">
+                <div class="set-password form-group">
+                  <h6>
+                    Log Persistence
+                  </h6>
+                  <div class="set-password-dialog">
+                    WARNING: leaving log persistence on for extended periods can damage your system.
+                    <br/>
+                    As such, we suggest also using the "Auto off" function, that will deactivate persist logs after the input number of days.
```
rtertiaer, line 159:
```suggestion
                    As such, we suggest also using the "Auto off" function. This will deactivate persistant logs after the specified number of days.
```

scripts/configure.py
```diff
@@ -120,11 +120,17 @@
...
+            'echo "If first deploy, reconfiguring journald to only log to RAM"',
+            r'[ ! -d /var/log/journal ] && echo -e "[Journal]\nStorage=volatile\nRuntimeMaxUse=64M\nForwardToConsole=no\nForwardToWall=no\n" | sudo tee /etc/systemd/journald.conf',
             'sudo systemctl enable systemd-journald.service',
             'sudo systemctl restart systemd-journald.service',
 
+            'echo Create directories and cronjobs for log persistence options',
+            'sudo mkdir -p /var/log/journal',
+            'sudo systemd-tmpfiles --create --prefix /var/log/journal',
+            'sudo cp config/deactivate_persist_logs_crontab /etc/cron.d/deactivate_persist_logs',
+            'sudo cp scripts/increment_auto_off.py /usr/local/bin/increment_auto_off.py',
```
rtertiaer, line 132: could you use the `copy` functionality with the `sudo` param set?

scripts/increment_auto_off.py
rtertiaer: it looks like you get the status of the persistance timing from the config file, but set anything new via the web interface. it feels like you should pick one interface and stick with it, imo

rtertiaer: also, i'd request that you don't shell out to curl to set this. use `requests`, or if you don't want something outside of the standard lib use `urllib.request`.

amplipi/updater/asgi.py
SteveMicroNova: No, that's just a fallback value. A fallback value is only used if the section or header being read can't be found, in this case it will yield 14 and then any future `POST` requests will create both header and section so it will be properly read in the future

Basically it protects against edge cases where the file was deleted somehow, or the deploy script had an error.

amplipi/updater/asgi.py
```diff
@@ -62,6 +64,22 @@
...
 
+def create_logging_ini():
+  """Fallback in case the ini doesn't exist, set to default settings. Only really comes up during github tests."""
+  ini = '/var/log/logging.ini'
+  tmp = '/tmp/logging.ini.tmp'
+
+  if not os.path.exists(ini):
+    conf = configparser.ConfigParser(strict=False, allow_no_value=True)
+    with open(tmp, "+w", encoding="utf-8") as file:
+      conf.read(tmp)
```
rtertiaer, line 75: do you want to read `tmp` or `file` in this call?

linknum23:
> image

Could this be stylized a bit maybe with some added spacing and more significant headers? I can't tell where one config stops and the next one starts.

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+# A program that ticks a counter down until deactivating log_persistence
+
+import requests
```
rtertiaer, line 4: another thing to note here. you're running with the system python, outside of any virtualenv. it looks like our default image comes with the Debian package `python3-requests`, but I don't see it in `configure.py`, which means this would otherwise break when we start using automated images. Could you pull this package in?

SteveMicroNova:
image

linknum23:
Here's a couple of things that could make the log persistence config more cohesive
- Move the warning underneath the checkbox and make it grey?
- Change the "Log Persistence" check boxes label to "Enable" since we have the header above

SteveMicroNova:
> Here's a couple of things that could make the log persistence config more cohesive
> 
> * Move the warning underneath the checkbox and make it grey?
> * Change the "Log Persistence" check boxes label to "Enable" since we have the header above

image
image

linknum23:
I think this looks a lot better!

scripts/configure.py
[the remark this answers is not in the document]
rtertiaer: fixed in 90017f205074136fbd13da678e6bae7732463d4b

rtertiaer (changes requested):
couple final notes:

- [x] if we've renamed the updater, those first 3 tabs probably deserve more explicit verbiage ('Custom' doesn't make a whole lot of sense with specifying 'Release' or something.)
- [x] imo, this doesn't close out #741 unless this logging is turned on before/during important events
- [x] can you align these individual things, and standardize the vertical spacing between each contextual section? 
image

SteveMicroNova:
> imo, this doesn't close out [System log persistence during important events #741](https://github.com/micro-nova/AmpliPi/issues/741) unless this logging is turned on before/during important events

https://github.com/micro-nova/AmpliPi/issues/741#issuecomment-2243223153
Was this not your preferred solution after we talked about it?
I suppose it doesn't solve the issue to the letter, but to the spirit of the issue it adds the correct functionality

rtertiaer (changes requested):
couple more things:

- [x] * clicking "save persistence" gives the user no feedback that their request was handled by the UI; it feels unfinished or broken
- [x] * navigating around the individual menus in the updater is a bit broken now - try clicking "Admin Settings" -> "Set Password" -> "Support Tunnel" -> "Admin Settings" -> "Set Password"
- [x] * I've set persistence, and I see `/etc/systemd/journald/conf` reflect `storage=persistent`. When I run another update afterwards, it resets to `storage=volatile`, with no regard for my setting.

rtertiaer:
this ticket supports #741, in that it allows us to save logs persistently. However, it does not save logs during important events like upgrading or opening a support tunnel, which is explicitly what the ticket calls out in the title and description. If that is simply infeasible for whatever reason, that's a different story, but without that justification it doesn't close that ticket imo.

scripts/configure.py
```diff
@@ -113,18 +113,35 @@
...
 
-            'echo "reconfiguring journald to only log to RAM"',
-            r'echo -e "[Journal]\nStorage=volatile\nRuntimeMaxUse=64M\nForwardToConsole=no\nForwardToWall=no\n" | sudo tee /etc/systemd/journald.conf',
+            'echo "If first deploy, reconfiguring journald to only log to RAM"',
+            r'[ ! -d /var/log/journal ] && echo -e "[Journal]\nStorage=volatile\nRuntimeMaxUse=64M\nForwardToConsole=no\nForwardToWall=no\n" | sudo tee /etc/systemd/journald.conf',
             'sudo systemctl enable systemd-journald.service',
             'sudo systemctl restart systemd-journald.service',
 
+            'echo Handle dependencies for log persistence options',
+            'sudo mkdir -p /var/log/journal',
```
SteveMicroNova, line 141: > I've set persistence, and I see /etc/systemd/journald/conf reflect storage=persistent. When I run another update afterwards, it resets to storage=volatile, with no regard for my setting.

Not true, this line of `configure.py` checks if the `/var/log/journal` directory exists, and if not it sets the `journald.conf` file to our default
`/var/log/journal` does not exist by default, but is created a few lines down so this functionally works as an "Only on first run of `configure.py`, set `journald.conf` to these parameters"

SteveMicroNova:
https://github.com/user-attachments/assets/e4bc840b-e7a7-4b85-a4a2-f321d169be50

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,41 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+sh = logging.StreamHandler(sys.stdout)
+logger.addHandler(sh)
+
+
+response = requests.get('http://localhost:5001/settings/persist_logs', timeout=10)
+state = response.json()
+
+if state["persist_logs"] and state["auto_off_delay"] > 0:
```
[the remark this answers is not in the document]
rtertiaer: or test for these keys existence

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,52 @@
...
+        data=json_data,
+        timeout=10,
+      )
+
+      if response.ok:
+        loop = False
+  except Exception as exc:
+    logger.exception(f"Auto off increment has failed with the following exception, trying again in 5 seconds: {exc}")
+  finally:
+    time.sleep(5)  # Loop until there is a response.ok at the end
```
SteveMicroNova, line 52: The only mechanism by which this can fail is network issues, so I think that a simple retry will suffice
There is no way to have a key issue as the file is verified as being there at the start of every endpoint, and if it isn't there or doesn't have all the proper headers then the file is created or the headers added.

amplipi/updater/asgi.py
```diff
@@ -95,6 +112,93 @@ class ReleaseInfo(BaseModel):
...
+      with open(logtmp, 'w', encoding='utf-8') as file:
+        log.write(file)
+      subprocess.run(['sudo', 'mv', logtmp, logconf], check=True)
+      logger.info(f"auto_off_delay set to {data.auto_off_delay}")
+
+    else:
+      logger.info("auto_off_delay unchanged")
+  except Exception as exc:
+    logger.exception(str(exc))
+    return 500
```
SteveMicroNova, line 199: I don't think `return 500` is a sensible way to handle things, but it is what we do everywhere else and one of the notes was "do this like everything else" and so I've simply followed the template of the endpoints beneath this

SteveMicroNova (commented):
My comments turned into a review at some point during writing so here's an obligatory box of text with the comment tag so I don't have to rewrite what I've said

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
```
SteveMicroNova, line 1: In the last few commits I've entirely reworked this file

Calling attention to it because the whole thing is green either way

amplipi/updater/static/index.html
```diff
@@ -105,8 +98,133 @@ <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1
...
+                This tab allows you to change settings relating to the core of AmpliPi
+              </div>
+
+              <!-- Password -->
+              <div id="set-password column-item">
+                <h6 class="hover-label">
+                  Set Password
+                  <span class="hover-tooltip">
+                    This form lets you set a web interface password, which will make it so that you must input the same password to access the system's controls. Leaving the fields empty will remove password protection. Note that this only affects the web interface, not the system password, and will log out any active sessions.
+                  </span>
```
SteveMicroNova, line 169: Both tooltips probably need some rewording

I moved them to be attached to the headers rather than to the input labels with the last commit to make them mobile compatible, previously you would've had to potentially inadvertently check the "enable" checkbox of persist logs to see what the tooltip has to say on mobile due to their lack of hoverover. This has had the effect of changing the context of the tooltips as well

web/src/pages/Settings/Settings.jsx
```diff
@@ -151,7 +151,7 @@ const Settings = ({ openPage }) => {
                                 <UpdateIcon fontSize="inherit" />
                             </Badge>
                         </div>
-                    Updates
+                    Admin Panel
```
SteveMicroNova, line 154: Screenshot 2024-10-04 121440
Screenshot 2024-10-04 122056
Had to make a compromise to name this something accurate that also fits on the thinnest screens

rtertiaer: love this; it's official :sunglasses:

linknum23:
Looks like we are going to need to update some update docs in several places here and on the interwebz

SteveMicroNova:
> Looks like we are going to need to update some update docs in several places here and on the interwebz

Yeah, that was my concern with that change. There's a lot of places that we can't change either such as other people's posts on discourse and old email chains, though luckily those instances feel like they've got fairly limited blast radius
I'll make sure to update them in my #923 branch if this is merged before that (which I expect)

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,50 @@
...
+          logging.info(f"Persist logs will be automatically turned off in {delay} day(s)")
+        else:
+          logging.info("Persist logs has been turned off automatically")
+    else:
+      loop = False
+  except Exception as exc:
+    logger.exception(f"increment_auto_off.py has failed with the following exception, trying again in 5 seconds: {exc}")
+  finally:
+    if loop:
+      time.sleep(5)  # Loop until there is a response.ok at the end
```
rtertiaer, line 50: in this case, an endless loop is a bad idea. there is no indication that the error in question will suddenly resolve itself by virtue of trying again, since we catch all exceptions instead of ones that might actually work after retrying, such as anything `requests` raises. you also missed an entire logical branch where this could "fall through" and use some stale value.

even if it _is_ something that trying again could resolve, if it's not noticed in many days or months you could end up running hundreds of these if not caught. they should have an expiration date, at a bare minimum. 

looping to hit a webserver can exacerbate the issue; if the web server is not able to respond in a timely fashion, you could overload it with requests. that's especially problematic because of the above - you could end up with 300 scripts looping and hitting the web server over and over, and everything breaks down. that's especially sketchy because the server we're Denial-of-Service-ing has multiple ways to fix this when it happens (the support tunnel and a firmware upgrade), though a user is more liable to just restart their box before either of those. still, it's extremely bad in larger web app setups - your own webapp can DDOS your own webserver, if you ship it to your 10000 daily users (or at least make it balloon in scale and become very costly.)

I would personally not wrap any of this in a loop and [use a retry decorator with much better options](https://tenacity.readthedocs.io/en/latest/), especially for the exponential backoff and jitter options.

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,50 @@
...
+loop = True
+
+while loop:
+  try:
+    response = requests.get('http://localhost:5001/settings/persist_logs', timeout=10)
+    state = response.json()
+
+    if state["persist_logs"]:
+      future_persist_state = state["auto_off_delay"] != 1
+      delay = state["auto_off_delay"] - 1
```
rtertiaer, line 25: you _need_ to check for these keys existence. please add this file to our `mypy` checking and resolve all issues it reports.

rtertiaer (changes requested):

scripts/increment_auto_off.py
SteveMicroNova: image

I feel like I've added them to our workflows properly, though picking through the completed workflows I don't see the file getting checked either way. Here's something that approximates a badge of approval from the checkers, though.

scripts/increment_auto_off.py
SteveMicroNova: This is even more true now

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,116 @@
...
+  try:
+    # If the api is inaccessible, assume it is down for whatever reason and extrapolate that there is no race condition to reading it directly
+    journalconf = configparser.ConfigParser(strict=False, allow_no_value=True)
+    journalconf.read('/etc/systemd/journald.conf')
+
+    logconf = configparser.ConfigParser(strict=False, allow_no_value=True)
+    logconf.read('/var/log/logging.ini')
+
+    persist = journalconf.get("Journal", "Storage", fallback="volatile")
+    state_persist = True if persist not in ("volatile", "persistent") else persist == "persistent"  # If value is somehow invalid, assume it's True just in case
```
SteveMicroNova, line 45: I assume that persistence is on in case you somehow corrupt your journald.conf file during a catastrophic issue, that way it remains on for troubleshooting purposes

I do not see a reality where you have that file set to an invalid option and meant for it to be off. If that is the case though, simply wait 14 days by default and you'll be fine.

scripts/increment_auto_off.py
```diff
@@ -0,0 +1,126 @@
...
+  api_read_failure = True
+except Exception as exc:
+  logger.exception(f"increment_auto_off.py has failed with the following exception:\n{exc}")
+  api_read_failure = True
+
+
+if api_read_failure:
+  try:
+    # If the api is inaccessible, assume it is down for whatever reason and extrapolate that there is no race condition to reading it directly
+    # This section is a copy of the /settings/persist_logs GET in asgi.py
```
rtertiaer, line 39: if this is a copy of any section of the code, you should probably create a function out of it and call it from both places

rtertiaer (approved):

scripts/increment_auto_off.py
SteveMicroNova: This is difficult due to them being two entirely separate programs unfortunately

I tried doing something like adding a symbolic link between the amplipi directory and `/usr/local/bin` where this script ends up so I could do a relative import and that almost worked except for the fact that this script is outside of the venv and so doesn't have the dependencies to import anything without errors

Maybe someday I'll come back to this and make it even better, for now it's been two hours since you approved this PR so I'm gonna stop spending time on making it more than it is

Issue #836, opened 2024-07-26
Bluetooth does not work on 0.4.1


rtertiaer:

rtertiaer:
After investigating quite a bit, everything appears actuated correctly, and I took a step back. I flashed a fresh 0.3.1 and this is actually broken there too - this very well may be our testing dongle. Going to order a new one.

rtertiaer:
Conversely, this dongle works fine on my laptop. To note, it loads a different firmware, which has a ton of other discussion around it elsewhere: https://bugs.launchpad.net/ubuntu/+source/linux-firmware/+bug/1968604

We even package and ship our own firwmare for this device, though reinstalling the default fw doesn't seem to change anything.

rtertiaer:
I'm increasingly leaning towards this being a wacky combination of this chipset, firmware, and our OS. I'm going to bring in a personal dongle from home and test with that to get another data point.

rtertiaer:
It appears this ticket is actually because of some strange firmware business with the single USB transceiver we've been passing around at work - my crappy transceiver from home worked (almost) out of the box. See #842 for some shoring up against that "almost" qualifier.

PR #842, opened 2024-07-30
Some tweaks to bluetooth for robustness

rtertiaer:
It appears #836 is actually because of some strange firmware business with the single USB transceiver we've been passing around at work - my crappy transceiver from home worked (almost) out of the box. There are two tweaks I made to this though, for the first-run case I saw with my dongle from home - I suspect `bluetoothctl` timed out once in `mac_to_device_name`, but have no data to confirm that. This adds better logging and handling around that case.

I'm not sure this justifies a CHANGELOG entry. what do you think?

### Checklist


jonahshader (approved):
Changes make sense. I'd say this doesn't warrant a changelog entry.

PR #843, opened 2024-07-30
Replace Display Status Text with Number of Zones Playing/Paused/Stopped/Muted

mjustian:
### Checklist


rtertiaer (changes requested):
the flow control in `get_emoji_status` is extremely deep and brittle as a result - did we actually cover every case? If i have a new status or case, how do I add it (and how do I ensure I added it for every existing branch?)

at the cost of some computational complexity, it is possibly more robust to  do something like this:
```
enum STATE { ... } 

def get_zone_status(zone) -> STATE

def get_source_status(source) -> STATE

def total_status_counts(List[STATE]) -> Dict[str] // i'm sure there's a better representation but as a straightforward data manipulation example

def display(Dict[str]) // or you can cut out total_status_counts and handle a List[STATE]s directly

if streamer:
  states = total_status_counts([getSourceStatus(s) for s in sources])
else:
  states = total_status_counts([getZoneStatus(z) for z in zones])

display(states)
```

this can help decouple the data types, the data collection, and the display functionality, and provide various entrypoints for changes that don't require precise modifications every 10 lines for deeply nested cases.

I don't pretend to be an expert here though lol.

rtertiaer (approved):

PR #844, opened 2024-07-30
Revert b728da867 to prevent frequent re-renders

### LINKED ISSUE #839: Disabling LMS mode causes constant refreshes

rtertiaer:
This PR reverts b728da867 to fix #839 . Introducing `selectActiveSource` has made the Settings page stateful; this means it triggers a re-render every 1s, when we poll the upstream API for current state. I've tried removing all the state from the LMS Mode bits, but that's not what's causing the re-render anyways and did nothing.

A couple bigger/longer term fixes could include:
* rearchitecting the API to send only changes over websocket; this would rerender only on those changes, and depending on how decomposed those channels are we could easily prevent rerendering on changes by only updating, say, player objects instead of everything.
* creating an API endpoint to simply query lms_mode directly upon config page load. That feels like a hack, but would definitely work in the 99th percentile case.

I'm pretty sure there's a way to spend some time and make this work as intended, but this went deeper than expected and I'm not convinced the juice is worth the squeeze. Can you describe what case the introduction of `selectActiveSource` actually handles? If it's just two people using the same AmpliPi at once and someone deletes a player/source, that feels like a case not worth adding this additional complexity (given its side effects), tho it's likely I'm just not understanding the root cause here.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Streams
   * Internet radio can now play playlists from the `pls`, `m3u8`, and `m3u` formats.
   * Fixed bug where internet radio process would not completely stop when the stream was stopped.
+* Web App
+  * Fix a bug that caused re-renders on settings modals
 
 ## 0.4.1
 * Web App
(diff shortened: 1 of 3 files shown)


klay2000 (approved):
Yay react!!!
⚛️⛔

SteveMicroNova (approved):
You're correct that it only covers an instance of multiple users in one device and generally trying to never not have a source selected if possible, and you're also right that having people press one extra button in an edge case scenario is better than us refactoring the whole frontend

PR #845, opened 2024-07-30
Internet radio playlist

### LINKED ISSUE #801: runvlc/vlc streams sometimes don't stop running
Just encountered a rogue internet radio station playing on my AmpliPro; the station was playing, processes were not reaped, yet the system state didn't know about this. Restarting the stream actually started a new instance of everything. I don't have logs for how this occurred because of excessive debug logging. This seems suspiciously close to #400 in presentation. I'm starting this ticket in case it happens again.

There are some fixes that come to mind immediately:
* better track the PIDs of forked processes and map them to running streams
* examine the deactivate/et al functionality for VLC streams in particular and ensure we're cleaning up properly
* reduce logging so troubleshooting is easier

### LINKED ISSUE #827: Internet radio stream will not play .pls files
SomaFM changed the format of their stations and they no longer work with the internet radio stream, I'm assuming this is because they are now playlists and not just mp3s.

klay2000:
Fix internet radio on playlists
Fix runvlc.py not dying on internet radio shutdown
add test for internet radio

### Checklist

closes #801 
closes #827

### DIFF
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -38,7 +38,8 @@ import socket
 import hashlib  # md5 for string -> MAC generation
 import datetime
 
-from pandora.clientbuilder import SettingsDictBuilder  # pandora client from pydora
+from pandora.clientbuilder import SettingsDictBuilder
+import requests  # pandora client from pydora
 
 from amplipi import models
 from amplipi import utils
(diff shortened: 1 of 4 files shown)


amplipi/streams.py
```diff
@@ -1165,6 +1166,23 @@ def connect(self, src):
     src_config_folder = f"{utils.get_folder('config')}/srcs/{src}"
     os.system(f'mkdir -p {src_config_folder}')
 
+    # HACK check if url is a playlist and if it is get the first url and play it
```
rtertiaer, line 1169: I don't love it, but it wasn't made to be loved. i like not pulling in a new dep, and also your tests give me some confidence in this.

streams/runvlc.py
```diff
@@ -32,8 +32,16 @@
...
 import vlc
 import argparse
+import signal
 from typing import List, Optional, Any, IO
 
+
+def signal_handler(sig, _):
+  """Handle sigterm signal."""
+  log(f"Caught signal {sig}, exiting.")
+  sys.exit(0)
```
rtertiaer, line 42: does this actually kill the underlying VLC proc? [i don't see a `__del__()` method on the Instance class](https://www.olivieraubert.net/vlc/python-ctypes/doc/vlc.Instance-class.html) but perhaps this is being handled in another, ooky spooky way by the VLC lib (or I don't understand it well enough.)

tests/test_rest.py
```diff
@@ -1095,6 +1097,38 @@ def test_post_stream_cmd_live(clientnm, cmd):
   assert not jrv['info']['mock_streams']
   assert rv.status_code == HTTPStatus.OK
 
+# test internet radio
+
+
+def test_create_internetradio_live(clientnm):
+  """ for 10 most popular internet radio stations plus one fake, try creating an internet radio stream, wait 3 seconds, then check if it errored"""
```
rtertiaer, line 1104: :clap:

rtertiaer (approved):

streams/runvlc.py
rtertiaer: ah, this pokes `libvlc`, it doesn't create an underlying vlc proc. that works for me

PR #846, opened 2024-07-30
Let runvlc try to initially open a stream forever; retry streams forever

### LINKED ISSUE #832: Internet radio stations sometimes fail to load on first boot
It seems like the VLC player is trying to load teh URL before hte internet is online; the process itself doesn't fail out, but hangs instead.

rtertiaer:
This PR fixes #832 by permitting runvlc to loop forever when a stream first starts. Our prior logic bailed quickly; this instead calls a wrapped `restart_vlc()` when it fails and lets `restart_vlc` implement a backoff for persistently failed streams.

This PR also creates a new RestartVlcException and handles the restarting case "better" when a stream breaks when already started; we again attempt to restart forever. I'm running this on a started-but-disconnected amplipi overnight and seeing if it works when I plug it back in in the morning.

The initial bug occurred in our testing setup when the internet connection was shaky on first boot; I can imagine user cases where these are very helpful fixes, like with less-than-stable internet on a boat ;)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Streams
   * Internet radio can now play playlists from the `pls`, `m3u8`, and `m3u` formats.
   * Fixed bug where internet radio process would not completely stop when the stream was stopped.
+  * Fixed bug where internet radio process would permenantly fail during transient network errors
 * Web App
   * Fix a bug that caused re-renders on settings modals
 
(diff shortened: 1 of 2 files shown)


rtertiaer:
It turned the stream back on after being disconnected overnight :sunglasses:

klay2000 (approved):
Add those couple things we talked about (setting the stream info in a couple places) and it should be good!

PR #847, opened 2024-07-31
Fix ch0 playing mono on fm radio.

klay2000:
Fix ch0 playing mono on fm radio.

### Checklist

closes

### DIFF
--- a/config/asound.conf
+++ b/config/asound.conf
@@ -7,43 +7,40 @@ pcm.!default {
     slave.pcm       "ch0"
 }
 
-# Create source 0
-# Use type plug to autonegotiate connection bitrates/mono/stereo.
+# Create source 0 with volume control and channel conversion
 pcm.ch0 {
-    type                     plug
+    type            plug
+    slave.pcm       "ch0_softvol"
+    slave.channels  2
+}
+
+# Lower the output level to avoid clipping on the negative side of the preamp's
+# U2/U4/U6/U8 op-amps. The audio signal is centered at 2.5 V and the
+# op-amps cleanly pass down to ~1.4V, so max (2.5 - 1.4) / sqrt(2) = 0.78 Vrms.
+# Nominal DAC output is ~1.76 Vrms. So 0.78 / 1.76 = 0.44
+# Note: this is no longer true in Preamp Rev4 and greater, but
+# we have to leave this here for backwards compatibility.
+
+pcm.ch0_softvol {
+    type            softvol
+    slave.pcm       "ch0_dmix"
+    control.name    "Ch0 Volume"
+    control.card    0
+    max_dB          -7.1  # Limiting volume to 44% of max; 20 * log10(0.44) = ~-7.1 dB
+    resolution      256
+}
+
+pcm.ch0_dmix {
+    type            dmix
+    ipc_key         2867
+    ipc_perm        0666
     slave {
-        pcm {
-            type             dmix
-            ipc_key          2867
-            ipc_perm         0666
-            slave {
-                pcm          "hw:sndrpihifiberry,0"
-                period_time  0
-                period_size  1024
-                buffer_size  8192
-            }
-        }
+        pcm         "hw:sndrpihifiberry,0"
+        period_time 0
+        period_size 1024
+        buffer_size 8192
+        channels    2
     }
-
-    # Lower the output level to avoid clipping on the negative side of the preamp's
-    # U2/U4/U6/U8 op-amps. The audio signal is centered at 2.5 V and the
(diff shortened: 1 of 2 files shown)


config/asound.conf
```diff
@@ -7,43 +7,40 @@ pcm.!default {
...
-    # Note: this is no longer true in Preamp Rev4 and greater, but
-    # we have to leave this here for backwards compatibility.
-    ttable.0.0               0.44
-    ttable.1.1               0.44
+# Lower the output level to avoid clipping on the negative side of the preamp's
+# U2/U4/U6/U8 op-amps. The audio signal is centered at 2.5 V and the
+# op-amps cleanly pass down to ~1.4V, so max (2.5 - 1.4) / sqrt(2) = 0.78 Vrms.
+# Nominal DAC output is ~1.76 Vrms. So 0.78 / 1.76 = 0.44
+# Note: this is no longer true in Preamp Rev4 and greater, but
+# we have to leave this here for backwards
```
linknum23, line 22: compatibility?

We may also want to show the conversion from 0.44 to -3.57 dB

linknum23 (commented):
I would be curious if just adding channels 2 at each pcm level would fix this problem
```
pcm.ch0 {
    type            plug
    slave {
        pcm {
            type             dmix
            ipc_key          2867
            ipc_perm      0666
            channels             2
            slave {
                pcm          "hw:sndrpihifiberry,0"
                period_time       0
                period_size  1024
                buffer_size   8192
                channels             2
            }
        }
    }
  }
```

linknum23:
As far as I can tell this is the most significant change.

klay2000:
> I would be curious if just adding channels 2 at each pcm level would fix this problem
> 
> ```
> pcm.ch0 {
>     type            plug
>     slave {
>         pcm {
>             type             dmix
>             ipc_key          2867
>             ipc_perm         0666
>             channels            2
>             slave {
>                 pcm          "hw:sndrpihifiberry,0"
>                 period_time     0
>                 period_size  1024
>                 buffer_size  8192
>                 channels        2
>             }
>         }
>     }
>   }
> ```
I think I tried that but I just tried it again to be sure and channels does not appear to be a valid field in a PCM.

linknum23:
As discussed in person. This passes a general sniff test for me. Pushing it off to Michael for a thorough inspection since he last modified this portion.

config/asound.conf
```diff
@@ -7,43 +7,40 @@ pcm.!default {
...
-    # Amp gain = 32.8 dBV
-    # Max amp input = 23.8 dBV - 32.8 dBV = -9 dBV = 0.35 Vrms
-    # So to avoid clipping at 0 dB volume, set ttable to:
-    #   0.35 / 1.76 = 0.2
+pcm.ch0_softvol {
+    type            softvol
+    slave.pcm       "ch0_dmix"
+    control.name    "Ch0 Volume"
+    control.card    0
+    max_dB          -3.57  # Limiting volume to 44% of max; 10 * log10(0.44) = -3.57 dB
```
Lohrer, line 29: This used to be -6.8 dB, is this -3.57 correct?

Lohrer:
It looks like the major difference here is reverting back to having a softvol in front of the dmix, see the change that switched out the softvol for a dmix here https://github.com/micro-nova/AmpliPi/commit/935aa20c1d6c0ca8e9504fa06207f707db2e0cae. I would like to know if this new config can handle multiple things trying to play to it, which is why we used the dmix in the first place.

config/asound.conf
Lohrer: Oh I see, it should be 20*log10() instead of 10*log10() since we are measuring a change in voltage. So 20*log10(0.78/1.76) ~= -7.1 dBV, looks like I used the slightly less precise `0.78/1.7` before to get the -6.8 dB. Either is probably close enough for the average hardware.

klay2000:
this appears to allow multiple instances of vlc to play to ch0

config/asound.conf
```diff
@@ -7,43 +7,40 @@ pcm.!default {
...
-    # Max amp input = 23.8 dBV - 32.8 dBV = -9 dBV = 0.35 Vrms
-    # So to avoid clipping at 0 dB volume, set ttable to:
-    #   0.35 / 1.76 = 0.2
+pcm.ch0_softvol {
+    type            softvol
+    slave.pcm       "ch0_dmix"
+    control.name    "Ch0 Volume"
+    control.card    0
+    max_dB          -7.1  # Limiting volume to 44% of max; 20 * log10(0.44) = ~-7.1 dB
+    resolution      256
```
Lohrer, line 30: Looking at the ALSA source this appears to be a hardcoded value, so I'm not sure there's a point to this configuration value.

Lohrer (approved):
I looks reasonable to me

linknum23 (approved):

PR #849, opened 2024-07-31
Fix internet radio logo field validation bug

### LINKED ISSUE #848: Internet Radio Streams Disappeared and Can't Be Added After Update
I installed version 0.4.0 and my internet radio streams are no longer visible.
When trying to add the same streams back, I receive an error "create stream failed: expected string or bytes-like object"
I did not look what version I was previously on, but these same streams worked fine before this update.

An example URL of one of the streams: https://us9.maindigitalstream.com/ssl/knza

rtertiaer:
This PR fixes #848 . The `logo` field is marked as `Optional[str]` in the function that calls `validate_sttream`, and a `None` gets passed in when we do not specify one. This broke the regex processing; I've gated that with a truthy test first.

I also took the opportunity to revisit a couple of stream's `validate_stream()` methods to explicitly ensure required values are set. We don't bump into this because our frontend also does field validation for required fields, but helps the API/automation case. This has the potential side effect of breaking existing streams in the world that don't implement required fields, but I think they'd just be broken anyhow.

I check my code with `ruff`, and it spotted a couple of actual errors too and errata; I've fixed those here as well.

I've tested every stream I modified through the frontend, and all still instantiate & function as expected.

### Checklist

### DIFF
--- a/amplipi/streams.py
+++ b/amplipi/streams.py
@@ -23,7 +23,6 @@ a consistent interface.
 
 import os
 import traceback
-from re import sub
 import sys
 import subprocess
 import time
@@ -31,7 +30,6 @@ from typing import Union, Optional, List, ClassVar
 import threading
 import re
 import logging
-import ast
 import json
 import signal
 import socket
@@ -1171,7 +1169,7 @@ class InternetRadio(BaseStream):
     # parse each playlist type and get the urls from them
     PLAYLIST_TYPES = ['pls', 'm3u', 'm3u8']
     if self.url.split('.')[-1] in PLAYLIST_TYPES:
-      logger.info(f'Playlist detected, attempting to get playlist...')
+      logger.info('Playlist detected, attempting to get playlist...')
       try:
         req = requests.get(self.url)
         urls = re.compile(r'(http.*?)[\r\n]').findall(req.text)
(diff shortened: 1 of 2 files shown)


amplipi/streams.py
```diff
@@ -23,15 +23,13 @@
...
 import traceback
-from re import sub
 import sys
 import subprocess
 import time
 from typing import Union, Optional, List, ClassVar
 import threading
 import re
 import logging
-import ast
```
rtertiaer, line 34: these are unused.

amplipi/streams.py
```diff
@@ -1249,10 +1254,13 @@ def send_cmd(self, cmd):
...
 
   def validate_stream(self, **kwargs):
+    if 'url' not in kwargs or not kwargs['url']:
+      raise InvalidStreamField("url", "url is required")
     URL_LIKE = r'^https?://[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)$'
     if 'url' in kwargs and not re.fullmatch(URL_LIKE, kwargs['url']):
       raise InvalidStreamField("url", "invalid url")
-    if 'logo' in kwargs and not re.fullmatch(URL_LIKE, kwargs['logo']):
+    # Logo is Optional[str]
+    if 'logo' in kwargs and kwargs['logo'] and not re.fullmatch(URL_LIKE, kwargs['logo']):
```
rtertiaer, line 1263: this is the actual fix for #848

rtertiaer:
Okay, I've removed a bunch of the stream validation things I added; apparently we use the API in that particular way (makes sense in the context of PATCHes.)

klay2000 (approved):
this was actually bothering me because I thought we didn't require the logo before, but I just figured I was misremembering

Issue #599, opened 2024-01-03
Update updater to flash full images


rtertiaer:

rtertiaer:
what i envisioned, at a high level:

-     the amplipi requests a list of available updates for it from an upstream api, sending in its current hw/fw/sw revs
-     the upstream api returns a list of versions available for that wombo-combo and where to grab a more detailed manifest for a particular version
-     the amplipi grabs that manifest. it contains checksums, gpg signatures for the image itself, but importantly it also has a list of pre- & post-install commands
-     during upgrade, pre-commands are run "first" (perhaps concurrently to flashing?), then dding the image, then either chroot & run post-install & OR reboot & run post-install.
-     maybe send an 'upgrade successful' response back to the API, so we can spot things before they become issues

SteveMicroNova:
This is almost entirely done on #1106, minus the e-ink portion.

PR #851, opened 2024-08-01
Take regular backups of the config to ~pi/backups

rtertiaer:
This PR is a simple, barebones point-in-time backup of a user's `.config` directory. We appear to write back the config at most every 5s during changes, through the `save_on_success` wrapper in `ctrl.py`, so simply creating tarballs like this feels acceptable. We could implement a frontend to allow users to restore prior configs more-automated-ly, but that is a bigger task and I'm unconvinced I want to do that now, when a simple support case or user-initiated SSH session can fix this; as an interim solution, I've documented in our docs how to backup and restore both manual and cronjob-generated backups.

I've opted to stash these in ~pi/backups because /home is where we intend to store state that moves across versions of OS we deploy post- #599 . A test backup on my appliance consumed ~1.5kb.

I've tested the script and it works; I'm going to wait overnight to make sure the cronjob kicks off on my appliance as well.

### Checklist

* [ ] Have you tested your changes and ensured they work? - just waiting on cronjob

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   * Delete generated tarball after finishing deploying using scripts/deploy
 * System
   * Fixed bug where the logs would be filled with errors when startign Spotify.
+  * Take regular configuration backups to the backend
 
 ## 0.4.1
 * Web App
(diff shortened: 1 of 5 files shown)


klay2000 (approved):
:floppy_disk:

PR #854, opened 2024-08-02
break up streams.py

klay2000:
Breaking up streams.py into separate files, throwing this here before I finish the rest of the issue to opinions and merge it before it becomes too much of a nightmare.


rtertiaer:
my opinion is YES OMG PLS YES

linknum23 (commented):
Looking good. Still need to test this...

linknum23:
Can you try getting this to work with Airplay?  Every stream in our config worked besides that one.

PR #864, opened 2024-08-05
fix internet radio search on IOS app

### LINKED ISSUE #856: Internet Radio search is broken on iOS app
iOS: 17.5.1

klay2000:
fixes internet radio search on ios

### Checklist

closes #856

### DIFF
--- a/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
+++ b/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
@@ -117,7 +117,7 @@ const InternetRadioSearch = ({ onChange }) => {
             return;
         }
 
-        fetch(`http://${host}/json/stations/search`, {
+        fetch(`https://${host}/json/stations/search`, {
             method: "POST",
             headers: { "Content-Type": "application/json" },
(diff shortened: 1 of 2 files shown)


rtertiaer (approved):
I don't love it but it is what it is. For whomever finds this PR later, the first server is pinned because of upstream's SSL cert configuration - there are no subject alternative names that specify `all.api.radio-browser.info`. I'm less concerned about this than would first appear, because we're still load balancing the expensive requests across the servers listed in this request.

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -130,7 +130,7 @@ const InternetRadioSearch = ({ onChange }) => {
 
     return (
         React.useEffect(() => {
-            fetch("http://all.api.radio-browser.info/json/servers").then((res) =>
+            fetch("https://at1.api.radio-browser.info/json/servers").then((res) =>
```
linknum23, line 133: There should be a comment here

linknum23: what happens when at1 is down?

PR #869, opened 2024-08-08
prevent avahi from serving support_tunnel addresses

### LINKED ISSUE #826: a running support tunnel causes avahi to sometimes serve up the wireguard address instead of the LAN address

rtertiaer:
This PR stops avahi from serving addresses that are used with the support_tunnel. It does this by adding and removing entries from `avahi-daemon.conf`'s `deny-interfaces`. I would have liked to accomplish this another way - it's more than a little messy to modify and reload service configs on the fly like this - but avahi does not support any `conf.d`-style includes.

I've tested this a bunch. If you also wanted to test this, be sure to run a `support_tunnel` on your AmpliPi with https://github.com/micro-nova/support_tunnel/pull/26 merged. 

It also adds a little helper in `configure.py` for files being copied that do not yet have a parent directory.

This fixes #826 

### Checklist

### DIFF
--- a/scripts/support_tunnel_post_down.sh
+++ b/scripts/support_tunnel_post_down.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# This script runs after a support tunnel comes down. Right now, 
+# it only prevents Avahi from advertising using the wireguard interface
+# $1 is the wireguard interface to remove
+
+# The avahi-daemon.conf follows an ini-style convention; Python's
+# configparser makes it easy to safely add & remove contents in a particular
+# section, and "shelling out" gives us an opportunity to use `sudo`.
+echo "import configparser
+import sys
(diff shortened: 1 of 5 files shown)


scripts/support_tunnel_post_down.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+# This script runs after a sdownport tunnel comes down. Right now,
```
klay2000, line 2: after a support tunnel comes down?

klay2000 (approved):
:+1:

scripts/support_tunnel_post_down.sh
rtertiaer: lmfao, this was when I copied the `post-up` script and just ran `s/up/down/` :woozy_face:

PR #880, opened 2024-08-13
Make internet radio URL validation more correct & robust

rtertiaer:
This PR does less obtuse and more thorough checking of internet radio URLs by farming the validation logic out to a library, whose sole job is to validate inputs.

### Checklist

### DIFF
--- a/amplipi/streams/internet_radio.py
+++ b/amplipi/streams/internet_radio.py
@@ -1,5 +1,6 @@
 from .base_streams import BaseStream, InvalidStreamField, logger
 from typing import ClassVar, Optional
+from urllib.parse import urlparse
 from amplipi import models, utils
 import subprocess
 import time
@@ -8,6 +9,7 @@ import re
 import requests
 import json
 import sys
+import validators
 
 
 class InternetRadio(BaseStream):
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -38,5 +38,6 @@ types-ujson==5.8.0.1
 types-urllib3==1.26.25.14
 upnpy==1.1.8
 uvicorn==0.20.0
+validators==0.20
```
rtertiaer, line 41: latest version compatible with Python 3.7

amplipi/streams/internet_radio.py
```diff
@@ -137,9 +139,15 @@ def send_cmd(self, cmd):
...
+      if not validators.url(kwargs['url']):
+        raise InvalidStreamField("url", "invalid url")
+      if urlparse(kwargs['url']).scheme not in ['http', 'https']:
+        raise InvalidStreamField("url", "unsupported protocol/scheme in url")
+
     # Logo is Optional[str]
-    if 'logo' in kwargs and kwargs['logo'] and not re.fullmatch(URL_LIKE, kwargs['logo']):
-      raise InvalidStreamField("logo", "invalid logo url")
+    if 'logo' in kwargs and kwargs['logo']:
+      if not validators.url(kwargs['logo']):
```
rtertiaer, line 150: this interface changes sometime after 0.20 - `url()` will instead raise an exception. We'll at least catch it when we upgrade ezpz thanks to testing, but just to note cuz this being way different while staying within the major version (even tho it's 0) confused me for way longer than it should have.

klay2000 (approved):
we should use this elsewhere as well

rtertiaer:
> we should use this elsewhere as well
It turns out we don't do any regex validation of URLs in any other `validate_stream`, which was surprising to me but makes sense for the classes where it was implemented - Airplay, pandora, spotify.

PR #888, opened 2024-08-15
Do not do validation on stream configs loaded from disk

### LINKED ISSUE #875: Broken streams should not disappear from the configuration
Even when a stream fails validation, it should not completely disappear from the configuration. This has occurred as a result of our new validation scheme, and has inadvertently shown both logical errors (#849) as well as unanticipated but valid stream configurations (#874.) 

@linknum23 proposed casting all streams that fail validation into an `InvalidStream` type, which will permit future updates to use configs that are broken in the current release and match our "save running state as the config" model quickly. A dead letter queue is also an option.

ref: https://amplipi.discourse.group/t/introducing-our-0-4-series/318/10?u=amplipi

rtertiaer:
This PR fixes #875 by disabling validation for stream configs loaded from disk. Editing an "invalid" stream will still produce an error indicating how it is failing validation, which potentially allows us to fix streams going forward without wiping invalid configs from disk; more or less, validation happens in every other case.

~I get that changing the function signatures to capture and pass around **kwargs is a little less than rigorious, but [there are tools to type this dict in modern Python](https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs) to add some rigor, and it's definitely flexy. Lemme know if that feels too messy though - I can simply only input the `validate` param in the `build_streams` cases where it's used too, np.~ edit: I made this more straightforward and touch less LOC, feel decent bout it. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
   * Stream validation for URLs has been made more robust
   * Minor bugfixes with Airplay 1 stream handling
   * Reset API password, support tunnels during factory reset
+Streams
+  * Fixed a bug where streams would disappear during upgrades by disabling validation when loading from disk
 
 ## 0.4.2
 * Streams
--- a/amplipi/ctrl.py
+++ b/amplipi/ctrl.py
@@ -338,7 +338,7 @@ class Api:
       assert stream.id is not None
       if stream.id:
         try:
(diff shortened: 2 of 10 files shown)


rtertiaer:
I'll write some tests for this functionality when the strategy here is given a green light.

klay2000 (approved):
seems fine

PR #891, opened 2024-08-16
StreamModal Errors

### LINKED ISSUE #887: Errors in the stream edit/creation dialog aren't evident
When testing some error-catching functionality, I intentionally triggered an error. It did not appear to do anything:
image

But after scrolling the viewport (for which there is no scrollbar, so it's not clear there is more to the dialog), you can finally see the error:
image

Make the error very evident. The message should be prominently displayed. Extra credit: it may also make sense to shake the entire modal after encountering an error; users intuitively understand that to mean "retry me".

SteveMicroNova:
Move `Alert` component somewhere more visible on StreamModal
Closes #887 

https://github.com/user-attachments/assets/a5746c9a-58f0-41af-a570-39376e985ef9

### Checklist

### DIFF
--- a/web/src/components/StatusBars/StatusBar.jsx
+++ b/web/src/components/StatusBars/StatusBar.jsx
@@ -1,0 +1,39 @@
+import React from "react";
+import PropTypes from "prop-types";
+import "./StatusBars.scss";
+import Snackbar from "@mui/material/Snackbar";
+
+import Alert from "@mui/material/Alert"
+
+export default function StatusBar(props) {
+    const {
+        open,
+        status,
+        text,
+        onClose,
+    } = props;
+
+    return(
+        <Snackbar
+            autoHideDuration={3000}
+            anchorOrigin={{vertical: "bottom", horizontal: "left"}}
+            open={open}
+            onClose={onClose}
+        >
+            <Alert // Cannot be AlertBar due to React rendering throwing errors despite it being the same thing
+                onClose={onClose}
+                severity={status ? "success" : "error"}
+                variant="filled"
+                style={{width: "100%"}}
+            >
+                {text}
+            </Alert>
+        </Snackbar>
+    )
+};
+StatusBar.propTypes = {
+    open: PropTypes.bool.isRequired,
+    status: PropTypes.bool.isRequired,
+    text: PropTypes.string.isRequired,
+    onClose: PropTypes.func.isRequired,
+};
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   * Migrate more components to mui, a Material UI implementation
   * Add marquee functionality the header of all modals, if the header text doesn't fit it will now scroll instead of pushing all the content down
   * Only allow one FM or Bluetooth stream to run at once
+  * Change how errors are displayed on Stream create/edit modal to be more obvious
 * System
   * Fixed a bug where support tunnel addresses would be served when querying for amplipi.local
   * Stream validation for URLs has been made more robust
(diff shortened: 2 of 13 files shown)


klay2000:
I think this might make it less visible, can you put the error message back on the modal and make it shake when an error occurs?

SteveMicroNova:
> can you put the error message back on the modal 

No, I moved it off the modal because there simply isn't any space for it unfortunately

> and make it shake when an error occurs?


https://github.com/user-attachments/assets/1fb460f5-c1a8-4829-9994-4f37de3d6401

rtertiaer:
>  there simply isn't any space for it unfortunately

why not put it between the scrolling viewport and the submit/cancel buttons? that's extremely prominent because it's right next to the user's cursor, and you can definitely shrink the scrolling viewport if you run out of vertical space

web/src/components/StatusBars/StatusBar.jsx
SteveMicroNova, line 1: This component is no longer used anywhere at all now that the `Alert` component is directly in the modal instead of being this version

I'd like to leave it in since it is a useful component that can easily be used elsewhere, but if it's requested to be removed as to avoid an instant vestigial folder I will remove it.

SteveMicroNova:
tl;dr it looked like more trouble than it was worth at first glance, that was not ultimately the case

I tried that initially to bad results due to where the `footer` param of the `ModalCard` component was placed
Originally, simply adding it to the `footer` param squished the buttons to the right and put the alert on the left side of the buttons, and I didn't attempt to change that because of how interconnected `ModalCard` is throughout the whole app
Since you asked, I dug a little deeper and found that the `footer` param isn't used in any of the 7 uses of `ModalCard` in the entire app, and so moving it was trivial in the end.

https://github.com/user-attachments/assets/f708a514-5ed0-4009-99ad-79d0d5c7fd36

klay2000:
add the shake back and should be good!

SteveMicroNova:
https://github.com/user-attachments/assets/cb10f52b-9581-438e-ac26-3b99e8588b97

SteveMicroNova:
@klay2000 It seems that the "every click reanimates the alert" thing is indeed a >10 minute endeavor (unless we want to add a useState that literally only exists to trigger rerenders, and either another useRef to handle the adding and removing of the animation or a useEffect to take note of the state to do the same)

tldr it is very easy if you don't give a shit about state management, time consuming to engineer otherwise

rtertiaer:
>  "every click reanimates the alert"

I think this is important. The goal of this ticket is to make it abundantly clear what's broken; this only shaking on the first click doesn't help if there is more than one error or you don't catch that the error showed up (unlikely, but possible.) It can help reiterate that the request still errored - say for example when the OG field was corrected, but another field also has issues. It also just feels broken if it only triggers on the first instance.

SteveMicroNova:
Can be seen on lamplipi as well:

https://github.com/user-attachments/assets/988b2430-755e-459c-992a-04ecdb0b5daa

rtertiaer (approved):
incrementing a counter to achieve this is definitely a smell, but also it seems pretty robust, safe, extremely contained.

SteveMicroNova:
> incrementing a counter to achieve this is definitely a smell, but also it seems pretty robust, safe, extremely contained.

Yeah, unfortunately [this is the way to do this sort of thing in React](https://www.geeksforgeeks.org/re-rendering-components-in-reactjs/#approach-1-simple-rerender), and [the standard Javascript way](https://stackoverflow.com/questions/33347992/reuse-css-animation-in-reversed-direction-by-resetting-the-state#:~:text=No%2C%20there%20is%20no%20way,delay)%20for%20it%20to%20restart.) is incompatible with React

Will def try to avoid that as much as possible in the future, luckily this here isn't too painful as you said

PR #893, opened 2024-08-16
Fix lms port edit

### LINKED ISSUE #878: LMS port removal fails in existing stream
ref: https://amplipi.discourse.group/t/amplipro-streamer-first-impressions/320?u=amplipi

rtertiaer:
This PR fixes up some of the rough edges around stream editing; this was brought up via editing LMS to remove the port from the config. 

There are two commits here. The first handles API errors in the frontend much more robustly, following the InvalidField conventions but for Pydantic generated errors. This is somewhat less than ideal, but seems robust enough for right now.

The second is legitimately an interface change and might deserve some further testing. This commit omits all "" members from an object we will PATCH to the API. This allows Pydantic to actually handle the `Optional` case in a relatively Pythonic way. I've linked a reference to a Pydantic issue. Lemme know your gut read on this and if it's at all sketchy I can rip it out or extensively test.

This PR fixes #878 .

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * Change "<- Back to App" button on the updater page to redirect to the settings page rather than the homepage
   * Update links on "About" settings tab to open in a new tab
   * Add scroll functionality to more modals, hide the scrollbars
+  * Handle errors better in the stream editing dialog
 * System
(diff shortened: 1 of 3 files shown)


klay2000 (approved):

rtertiaer:
alright, i've tested this with every optional field in the frontend and it seems to behave correctly. shipping~

PR #896, opened 2024-08-19
Fix Pandora account validation for new streams

### LINKED ISSUE #895: Cannot save new pandora streams
Because we now validate the user/pass with the pandora client, we attempt to use a client that gets instantiated in the constructor. however, that client has not yet been instantiated when this initial stream validation occurs. This client probably needs to be moved into a static method or similar, or some stuff shifted around.

This came up during testing for #893

rtertiaer:
This PR fixes #895 by moving forward the pandora client instantiation before the parent constructor calls `validate()`. I could also move this client outside of the class, or make it a static method; lemme know if you'd prefer those options.

### Checklist

* [ ] If applicable, have you updated the CHANGELOG? - I'm not updating the changelog because the bug this fixes is not a part of any release and only present on HEAD.


klay2000 (approved):
lgtm

Issue #900, opened 2024-08-20
Make UI report when running in an unsupported browser


rtertiaer:
Determine our minimum emcascript version necessary (we use `structuredClone`, used to use top level await, [check here](https://caniuse.com/)) and report to the user that their browser is unsupported and their experience is not guaranteed.

SteveMicroNova:
This is on [the 0.5.0 board](https://github.com/orgs/micro-nova/projects/3/views/7) but not [the filtered 0.5.0 board](https://github.com/orgs/micro-nova/projects/3/views/9)
Seems strange seeing as `user-experience` is part of the filtration model

Issue #857, opened 2024-08-02
FM Radio streams should not be able to run more than the number of adaptors present


rtertiaer:
It is possible to start multiple FM radio streams with only one adapter, but the process dies hard in the backend because the adaptor cannot be used by multiple streams at once.

rtertiaer:
I'm going to hardcode our support for FM radio to anticipate only one dongle at a time, and make that clear in the frontend.

rtertiaer:
I've disallowed this in our UI; having our backend handle this is either something for the distant future, or obviated by our redesign of streams et al. I'm going to move this out of the 0.5 milestone for now, and we can close or whatever when the future makes that path clearer.

PR #901, opened 2024-08-20
Disallow running multiple FM or Bluetooth streams at once in UI

rtertiaer:
This PR prevents running multiple FM stations or Bluetooth streams at the same time in the UI. (This is a bit of a misnomer for Bluetooth, as there is only ever one stream definition turned on or off anyways, but I handle that anyways for the future.) There's also some helper bits in the backend stream defs for these - `proc` & `bt_proc` are not present/set in the constructor, so `_is_running()` can exception if called before `connect()` is. (I intended to solve this in the backend, but leaving this to implement in client code is not the end of the world for now.) That means this is probably only half of #857 , but that second half can get punted much further into the future, especially considering we have aspirations of rewriting the stream scaffolding.

One thing to call out is [the use of `structuredClone`](https://caniuse.com/?search=structuredClone), just cuz it's so new. I added #900 regardless if `structuredClone` is in or out, because that will help folks have a good user experience.

It looks like this:
Screenshot from 2024-08-20 17-28-18

I've tested in Firefox, still gotta try other platforms.

### Checklist

### DIFF
--- a/web/src/components/StreamsModal/StreamsModal.jsx
+++ b/web/src/components/StreamsModal/StreamsModal.jsx
@@ -6,12 +6,12 @@ import { getIcon } from "@/utils/getIcon";
 import { setRcaSourceId } from "../ZonesModal/ZonesModal";
 import { moveSourceContents, setSourceStream } from "@/utils/APIHelper";
 import { setRcaStatus } from "../ZonesModal/ZonesModal";
-import List from "@/components/List/List";
-import ListItem from "../List/ListItem/ListItem";
+import List from "@mui/material/List";
+import ListItemButton from "@mui/material/ListItemButton";
+import ListItemText from "@mui/material/ListItemText";
+import Divider from "@mui/material/Divider";
 
 import PropTypes from "prop-types";
-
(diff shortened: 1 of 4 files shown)


web/src/components/StreamsModal/StreamsModal.jsx
```diff
@@ -6,13 +6,12 @@ import { getIcon } from "@/utils/getIcon";
 import { setRcaSourceId } from "../ZonesModal/ZonesModal";
 import { moveSourceContents, setSourceStream } from "@/utils/APIHelper";
 import { setRcaStatus } from "../ZonesModal/ZonesModal";
-import List from "@/components/List/List";
-import ListItem from "../List/ListItem/ListItem";
+import List from "@mui/material/List";
+import ListItemButton from "@mui/material/ListItemButton";
+import ListItemText from "@mui/material/ListItemText";
```
rtertiaer, line 11: replacing our own components with Material's has allowed a bit of consolidation on highlighting, styling, accessibility - but in this case, it particularly gives us a `disabled` param on `ListItemButton` that does what we would want out of the box.

klay2000 (approved):

PR #903, opened 2024-08-21
Move and backup primary config during upgrades

### LINKED ISSUE #902: Config move to new location isn't happening
In #804 , I attempted to move the primary config, `house.json`, into `~/.config/amplipi`. This doesn't appear to have taken. A fix for this both involves robustly copying over any running state into this new place during upgrade, and actually specifying the new path correctly.

rtertiaer:
This PR actually accomplishes #804 - the prior fix was incomplete. I'm filing this now but intend to do a bunch of testing tomorrow to make sure this does exactly what it says on the tin. This fixes #902 .

This PR also takes a backup of the config during _every_ upgrade. This would have saved us a lot of grief, and the script to make said backup is already in our repo - just need to trigger it during upgrades. 

### Checklist

* [x] Have you tested your changes and ensured they work? Yes, but I intend to poke at this more too.

### DIFF
--- a/docs/manual/TROUBLESHOOTING.md
+++ b/docs/manual/TROUBLESHOOTING.md
@@ -43,9 +43,12 @@ Manual backups are taken by navigating to ⚙ -> Config -> Download Config. This
 
 ### Automated system backups and restores
 
-Automated backups are also taken nightly at 2AM and stored for 30 days. These are raw system backups, accessible only by an advanced user via the backend. These backups help either support, or a technical user, to restore prior system configuration from a particular point in time. If the below instructions do not make sense to you, you should feel free to email [support@micro-nova.com](mailto:support@micro-nova.com) and we can help restore a backup from your appliance.
+Automated backups are also taken nightly at 2AM and stored for 90 days, and also whenever you upgrade to a newer version of software. These are raw system backups, accessible only by an advanced user via the backend. These backups help either support, or a technical user, to restore prior system configuration from a particular point in time. If the below instructions do not make sense to you, you should feel free to email [support@micro-nova.com](mailto:support@micro-nova.com) and we can help restore a backup from your appliance.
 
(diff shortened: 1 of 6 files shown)


rtertiaer:
Alright, this is ready for review. I've implemented some feedback in a second commit:
* takes a backup both before and after fw upgrade
* documents this a bit better

I've tested against our automated image; I'm going to test against an upgrade directly from our 0.3.1 image too. Note that this is only being tested with `./scripts/deploy`, but I'll make sure it works when I test the pre-release through the updater too.

rtertiaer:
I've tested the "upgrade from 0.3.1" case too - this all behaves as expected.

docs/manual/TROUBLESHOOTING.md
```diff
@@ -43,9 +43,9 @@ Manual backups are taken by navigating to ⚙ -> Config -> Download Config. This
...
 
-These backups are dated tarballs stored at `~pi/backups`. This tarball contains the entire `~pi/.config/amplipi` directory. To restore this backup, stop the AmpliPi service (`systemctl stop --user amplipi` as the `pi` user), unpack the tarball, overwrite the `~pi/.config/amplipi` directory with the contents of the backup, and start AmpliPi again (`systemctl start --user amplipi`).`
+These backups are dated tarballs stored at `~pi/backups`. This tarball contains the entire `~pi/.config/amplipi` directory. To restore this backup, stop the AmpliPi service (`systemctl stop --user amplipi` as the `pi` user), unpack the tarball (something like `tar --force-local xvzf $FILE`), overwrite the `~pi/.config/amplipi` directory with the contents of the backup, and start AmpliPi again (`systemctl start --user amplipi`).`
```
linknum23, line 48:
```suggestion
These backups are dated tarballs stored at `/home/pi/backups`. This tarball contains the entire `/home/pi/.config/amplipi` directory. To restore this backup:
1. Stop the AmpliPi service (`systemctl stop --user amplipi` as the `pi` user).
1. Unpack a backup tarball and overwrite the contents of `.config/amplipi` (something like `tar --force-local -xvzf backups/config_2024-08-22T12:42:31-04:00_pre-fw-upgrade.tgz -C /`).  Here we use `config_2024-08-22T12:42:31-04:00_pre-fw-upgrade.tgz` as an example backup file. 
1. Start AmpliPi again (`systemctl start --user amplipi`).`
```

linknum23: I verified the commands above work to restore an old config.

PR #904, opened 2024-08-22
Remove `Modal` component from Settings' `Page` component

### LINKED ISSUE #898: Stream selection header in settings truncated in small iOS
IMG_0043

The device is an iPhone 12 Mini running iOS 17.5.1.

SteveMicroNova:
Closes #898 by removing `Modal` component that was oversized for the screen

Update PR template to be slightly more concise, yet specific in a way that would've caught this issue's survival before merging #767, which aimed to fix it originally.

### Checklist


rtertiaer:
My hesitancy in approving this is that you have not reimplemented the `event.stopPropoagation()` bit or the CSS selector stuff. It is unclear offhand if those things are used.

SteveMicroNova:
> My hesitancy in approving this is that you have not reimplemented the `event.stopPropoagation()` bit or the CSS selector stuff. It is unclear offhand if those things are used.

I don't believe they are, but I've readded it just in case

web/src/pages/Settings/Settings.jsx
```diff
@@ -90,7 +76,13 @@ const Settings = ({ openPage }) => {
...
 
     if (openPage != "") {
-        return <Page openPage={openPage} />;
+        return (
+            <div className="settings-page-container">
+                <StopProp>
+                    <CorePage openPage={openPage} />
+                </StopProp>
+            </div>
+        );
```
rtertiaer, line 85: `Modal` accomplishes a couple of things:
* `event.stopPropogate()` during `onClose()`
* wraps all children in a `StopProp`
* applies a decent chunk of CSS from `Modal.scss`
* tags with a `modal_container` CSS class, which is used in the above class but could also be added to  elsewhere

can you validate that in all the `CorePage` cases none of this is used or necessary? that's my hesitancy to rip this out - it may fix the spacing issue, but I'm unconvinced we can test the surface area changes ripping it out entails in short order, before trying to press a release.

rtertiaer:
I think https://github.com/micro-nova/AmpliPi/pull/908 is actually the fix for this.

PR #924, opened 2024-08-30
bump version

SteveMicroNova:


SteveMicroNova:
"What happened to `0.4.3`?" the tags were fucked because the `0.4.3` prerelease wasn't tagged `-beta.0` and had bugs in it, Lincoln and I decided it was a lesser sin to skip a version than to replace the content of a tag with how our current update notifications work

linknum23 (approved):

PR #927, opened 2024-09-03
Add error catching to browse page fetch requests

SteveMicroNova:
### What does this fix?
There was a React error on the browse page when you either didn't have a stream selected, or selected a non-browsable stream and still got there (either by you browsing a stream another user closed, or had a bookmark for, or however else you'd choose to access that url when the direct link is unavailable).
React errors are serious problems as they require the mobile app to restart to be able to continue operating the device, this PR aims to give a device error instead by obviating the backend error into the frontend without React being bothered along the way

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.4.5
 * Web App
   * Ensure that abnormally-shaped album art is still horizontally centered
+  * Add error handling on browser page for instances where the selected stream isn't browsable
 
 ## 0.4.4
 * Web App
--- a/web/src/components/StatusBars/ResponseBar.jsx
+++ b/web/src/components/StatusBars/ResponseBar.jsx
@@ -18,7 +18,7 @@ export default function ResponseBar(props) {
         if(response != null){
             try {
                 const res = await response;
-                if (res.ok) {
+                if (res.ok && successText !== null) {
                     setSuccess(true);
                     text.current = successText;
                 } else {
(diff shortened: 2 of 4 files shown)


web/src/components/StatusBars/ResponseBar.jsx
```diff
@@ -48,7 +48,7 @@ export default function ResponseBar(props) {
     );
 }
 ResponseBar.propTypes = {
-    successText: PropTypes.string.isRequired,
+    successText: PropTypes.string,
     response: PropTypes.instanceOf(Promise).isRequired,
 };
```
SteveMicroNova, line 53: Making this optional is useful for cases where you want to show errors without caring about successes

SteveMicroNova:
Before and after video:
https://github.com/user-attachments/assets/0c19b26e-7346-44d8-aade-4860f848a49d

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## 0.4.5
+* Web App
+  * Add error handling on browser page for instances where the selected stream isn't browsable
+
+## 0.4.4
+* Web App
+  * Ensure that media device audio streams stop outputting audio when disconnected
+  * Limit max height of group creation dropdown
+  * Improve how marquee components look when not scrolling
```
SteveMicroNova, line 11: due to the ex-post-facto nature of 0.4.4, [the changelog](https://github.com/micro-nova/AmpliPi/releases/tag/0.4.4) isn't in our current git history on main

klay2000 (approved):
As discussed offline this is not a perfect solution as if you click too fast and end up in the situation described in the issue you get an error when everything is happening as intended, but we haven't managed to click that fast in several attempts so I think this is fine.

PR #935, opened 2024-09-23
make update process more rigorous

rtertiaer:
This PR covers edge cases in the upgrade process that can potentially prevent proper updates. This PR deserves a self-review to make it clear what each case helps against. I'm filing this a draft right now to self review & get it in front of people, and will begin immediately checking the below checkboxes.

### Checklist


amplipi/updater/asgi.py
```diff
@@ -282,7 +282,7 @@ def progress_sse(tasks):
           _sse_info(output)
         else:
           logger.warning(f'error: {output}')
-          _sse_error(output)
+          _sse_failed(output)
```
rtertiaer, line 285: this is the only change I am not sure about. Should a failed task fail the update completely?

amplipi/updater/asgi.py
```diff
@@ -329,7 +329,8 @@ def request_support():
       '/opt/support_tunnel/venv/bin/python3 -m invoke request'.split(),
       capture_output=True,
       cwd='/opt/support_tunnel',
-      timeout=120
+      timeout=120,
+      check=True
```
rtertiaer, line 333: putting `check=True` here handles this case just a smidge better; it's not strictly necessary.

scripts/backup_config.sh
```diff
@@ -12,3 +12,4 @@ tar --one-file-system -czf "${backup_filename}" "${HOME}/.config/amplipi/"
 # remove all backups older than 90 days
 echo "removing old backups..."
 find "${backup_dir}" -type f -ctime +90 -delete
+echo "completed ${0}"
```
rtertiaer, line 15: the very last command in a shell script's exit status is considered to be the script's exit status.. and `find` has some funny conditions around returning a non-zero exitstatus with `-delete` :melting_face: 

if we fail to delete backups? don't fail upgrades or consider this a failed process.

amplipi/updater/asgi.py
linknum23: I think this just makes it show failed. I'm unsure if it does more than that?

scripts/configure.py
```diff
@@ -1176,6 +1176,8 @@ def failed():
     pre_backup = _create_backup(env, "_pre-fw-upgrade")
     progress(pre_backup)
     tasks += pre_backup
+  if failed():
+    return False
```
rtertiaer, line 1180: considering we've had a spate of failed upgrades recently, and that i've changed the upgrade script most recently here with backups, this was where I started looking. It's of interest, because this pre-backup happens before we install requirements w/pip, but would not necessary fail until later - obviating the actual source of the issue.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+## Upcoming release
+* System
+  * Make update process properly report errors
+
+## 0.4.4
+* Web App
+  * Ensure that media device audio streams stop outputting audio when disconnected
+  * Limit max height of group creation dropdown
+  * Improve how marquee components look when not scrolling
+
```
rtertiaer, line 12: It appears we never updated the changelog for the released material in 0.4.4

linknum23: Yup. We did a couple of things our of order in the 0.4.4 release.

linknum23 (approved):
Looks good!

amplipi/updater/asgi.py
rtertiaer: That's correct, but the question is do we encounter errors during updates that should not qualify the update as failed? (Will this make the update seem sketchier than it actually is and prompt people to retry often?) I haven't seen any cases where that's true, but our customer boxes are potentially still in wildly disparate states.

linknum23: One place an error might make sense to be tolerated is in stream installation/update.

rtertiaer: I'm reverting this change; it only serves to surface issues with the configure.py script, but possibly in cases where it isn't "broken", and gives it the appearance of being broken anyways.

CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
 # AmpliPi Software Releases
 
-## 0.4.5
```
rtertiaer, line 3: 0.4.5 is not tagged yet.

PR #953, opened 2024-09-27
add disk image manipulation scripts

rtertiaer:
I've been gently refining these little disk image utilities for nearly a year now, and it feels time to make them a part of the actual repository here. These scripts allow one to quickly find an amplipi and flash to it; mount an image to a test dir; run applications within the image's chroot using `binfmt-support` & `qemu-user-static`, and unmount the same directory structure. 

### Checklist

### DIFF
--- a/scripts/image_testing/chroot.sh
+++ b/scripts/image_testing/chroot.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+
+if [ $(whoami) != "root" ] ; then
+  echo "Please run as root."
+  exit
+fi
+
+# perhaps this is the least commonly used util. perhaps not. 
+if [ "$(dpkg --get-selections qemu-user-static)" ] ; then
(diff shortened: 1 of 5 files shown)


scripts/image_testing/chroot.sh
```diff
@@ -0,0 +1,15 @@
...
+  echo "Please run as root."
+  exit
+fi
+
+# perhaps this is the least commonly used util. perhaps not. 
+if [ "$(dpkg --get-selections qemu-user-static)" ] ; then
+  apt install binfmt-support qemu-user-static
+fi
+
+chroot root /bin/bash
```
rtertiaer, line 14: a case could be made here to set up bindmounts or instances of namespaced filesystems (/sys, /proc, /dev, etc), but there's a nonzero likelihood that a setup like this could damage the user's system; thus, with the loss of some functionality, we explicitly do not do this.

PR #960, opened 2024-10-01
Limit looping in Browse.jsx

SteveMicroNova:
I merged #927 yesterday but made a last minute change that made it not work actually, it worked when I tested it because I tested the wrong version without realizing it (my bad)
This solves a newly created react error on the browse page by making the only function that is capable of having errors not run if there is actively an error message that is open on the page

### Checklist


klay2000 (approved):
Seems fine

PR #970, opened 2024-10-09
ZoneSettingsFeedback

### LINKED ISSUE #899: Applying zone settings should give user feedback
When I click "apply", the request is sent - but there is no indication to the user that this has happened. It may make some sense to move individual zone editing into a modal, like everywhere else, instead of the dropdowns it currently uses.

SteveMicroNova:
Closes #899
Implements StatusBar notifications into zone settings list items to give feedback when settings are saved

https://github.com/user-attachments/assets/127ee5d0-1c43-4b5b-ac9b-94e3dddf4b69

### Checklist

### DIFF
--- a/web/src/pages/Settings/Zones/Zones.jsx
+++ b/web/src/pages/Settings/Zones/Zones.jsx
@@ -14,6 +14,7 @@ import Divider from "@mui/material/Divider";
 import TextField from "@mui/material/TextField";
 import Switch from "@mui/material/Switch";
 import FormControlLabel from "@mui/material/FormControlLabel";
+import StatusBar from "@/components/StatusBars/StatusBar";
 
 const ZoneListItem = ({ zone }) => {
     const [open, setOpen] = React.useState(false);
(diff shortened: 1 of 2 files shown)


web/src/pages/Settings/Zones/Zones.jsx
```diff
@@ -108,19 +123,26 @@ const ZoneListItem = ({ zone }) => {
...
 };
 
+
 const Zones = ({ onClose }) => {
     const zones = useStatusStore((state) => state.status.zones);
 
     const listItems = zones.map((zone) => {
         return(
-            <>
+            < >
```
klay2000, line 145: ?

klay2000 (changes requested):
fix those couple accidental whitespace additions and then it's good

web/src/pages/Settings/Zones/Zones.jsx
SteveMicroNova: I was having trouble along the way in this PR and my console was full of warnings and errors relating to how these empty tags don't have keys
I briefly turned them into diva and when that didn't help me I reverted the change, but clearly not entirely

PR #974, opened 2024-10-14
Link to LMS Control Panel from settings

SteveMicroNova:
image
Change LMS toggle switch so that it doesn't download system config on LMS deactivation, include link to LMS server when LMS Mode is active

### Checklist

### DIFF
--- a/web/src/pages/Settings/Config/ConfigPanels/LMSMode.jsx
+++ b/web/src/pages/Settings/Config/ConfigPanels/LMSMode.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import "../Config.scss";
+import Button from "@mui/material/Button/Button";
+import Grid from "@mui/material/Grid/Grid";
 import ConfigPanel from './ConfigTemplates/ConfigPanel.jsx';
 import { useStatusStore } from "@/App.jsx";
 import Switch from '@mui/material/Switch/Switch';
@@ -19,6 +21,16 @@ export default function LMSMode() {
         setResponse(resp);
         setLoading(false);
     };
+
+    function LMSControlLink() {
+        if(lmsMode){
+            return (
+                <Button target="_blank" href={`http://${window.location.hostname}:9000`} className="config-spacing">
+                    LMS Control Panel
+                </Button>
+            )
+        }
+    }
 
     function LMSModal(){
         if(lmsMode){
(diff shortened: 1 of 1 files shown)


web/src/pages/Settings/Config/ConfigPanels/LMSMode.jsx
```diff
@@ -20,12 +21,22 @@ export default function LMSMode() {
...
+    }
+
     function LMSModal(){
         if(lmsMode){
             return(
                 <ConfigModal
                     body={"This will reset AmpliPi to factory settings, you will have to either manually reconfigure it or reupload the config that was downloaded when LMS mode was initially toggled."}
-                    onApply={() => {ConfigDownload(); LMSModeHandler();}}
+                    onApply={() => {LMSModeHandler();}}
                     open={modalOpen}
```
rtertiaer, line 40: not a fan of this change: https://github.com/micro-nova/AmpliPi/issues/976#issuecomment-2411833840 & https://github.com/micro-nova/AmpliPi/pull/805

rtertiaer (changes requested):
is there any way that button could look a little less haphazardly placed? I don't have any suggestions here - I know we're largely not designers - but I'd love to ensure we at least don't make things look worse. some things that I notice:
* it's a button, but doesn't do a button-like thing - it does a link-like thing
* there are two other types of buttons on the page; this is themed like neither of them
* there's zero margin/padding on this button between it and the divider - feels cramped
* it's not horizontally aligned to anything but the prior content, which feels unfinished

SteveMicroNova:
> * it's a button, but doesn't do a button-like thing - it does a link-like thing

I'm happy to make it a link again, though we'll probably have to hash that out a little further given Klayton's dislike of my previous link strategy

> * there are two other types of buttons on the page; this is themed like neither of them

Fixed

> * there's zero margin/padding on this button between it and the divider - feels cramped

This is the case for all of the buttons, but obviated by the `contained` styling; corrected in recent commit

> * it's not horizontally aligned to anything but the prior content, which feels unfinished

I've got no good way of fixing this as there isn't really a gridmodel in place for multi-input sections but I've taken a note from the "Upload Config" section and just made there be some spacing between the inputs


image

klay2000 (approved):
Lincoln, Steve, and I have decided the button-link thing should be fine as long as it opens in a new tab.

PR #978, opened 2024-10-14
PresetAlignment

### LINKED ISSUE #973: Select Preset modal item text is left-aligned
image

SteveMicroNova:
Add toggleable option for column formatting to ListItem, enact max-width + text wrapping on PresetsModal
Closes #973

Before:
image

After:
image

### Checklist

### DIFF
--- a/web/src/components/List/ListItem/ListItem.scss
+++ b/web/src/components/List/ListItem/ListItem.scss
@@ -1,17 +1,19 @@
 @use "src/general.scss";
 
 .list-item-container {
-  display: grid;
-  grid-template-columns: 0fr 2fr 1fr;
-  grid-template-rows: 1fr;
   align-items: center;
   padding: 0.5rem;
   border-bottom: 1px solid general.$secondary;
   @include general.button-hover;
 }
-
 .list-item-container:last-child {
   border-bottom: none;
+}
+
+.columns{
+  display: grid;
+  grid-template-columns: 0fr 2fr 1fr;
+  grid-template-rows: 1fr;
 }
 
 .list-item-child {
--- a/web/src/components/PresetsModal/PresetsModal.scss
+++ b/web/src/components/PresetsModal/PresetsModal.scss
@@ -8,6 +8,9 @@
 
 .presets-modal-card {
   max-height: 75vh;
+  width: 90vw;
+  max-width: 30rem;
+  text-overflow: wrap;
   overflow: hidden;
   display: flex;
   flex-direction: column;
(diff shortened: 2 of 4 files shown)


web/src/components/List/ListItem/ListItem.scss
```diff
@@ -1,19 +1,21 @@
 @use "src/general.scss";
 
 .list-item-container {
-  display: grid;
-  grid-template-columns: 0fr 2fr 1fr;
-  grid-template-rows: 1fr;
```
SteveMicroNova, line 6: This was the root of the issue, the parent component having grid templating works in 99% of cases because 99% of cases also have image icons in the first column, in the case of the presets modal it does not and so it tries to cram the whole content into the first column

I've made it a toggleable argument option that is true by default to not have to change those 99% of cases

web/src/components/PresetsModal/PresetsModal.scss
```diff
@@ -8,6 +8,9 @@
 
 .presets-modal-card {
   max-height: 75vh;
+  width: 90vw;
+  max-width: 30rem;
+  text-overflow: wrap;
```
SteveMicroNova, line 13: Gave it the same text wrapping and width limits as the streams and zones modals so that long names don't do this:
image

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
     * Move "Set password" page to be accessible via Admin Settings tab
   * Add success/fail messages to zones settings changes
   * Add state polling to some user workflows (stream starting, stopping; using presets) to remove the illusion of lag
+  * Sand off more rough edges too tiny to specifically list
```
rtertiaer, line 15: I think I'd prefer either explicit fixes listed here, a simple "UI fixes", or if they're too small omitting it from the CHANGELOG entirely. In this case, I like the second.

rtertiaer (commented):
seems fine; what platforms explicitly did you test against?

CHANGELOG.md
SteveMicroNova: Fair, I've just felt weird omitting like 7 PRs in a row from the changelog

SteveMicroNova:
> seems fine; what platforms explicitly did you test against?

Desktop:
- Chrome
- Fireox
- Edge

Mobile:
- Safari

klay2000:
This looks better, but what this dialog really needs is an overhaul so it's consistent with the others
The main two things are dividers and some sort of icon (not exactly sure how to deal with that yet)

SteveMicroNova:
> This looks better, but what this dialog really needs is an overhaul so it's consistent with the others The main two things are dividers and some sort of icon (not exactly sure how to deal with that yet)

image

It has dividers, and I don't think that having icons is a natural thing for this particular modal; I can't conceive of a way to add them without them feeling obviously shoehorned whether they be user-set or just numbered or whatever.

rtertiaer (approved):

PR #979, opened 2024-10-14
Fix streamer preset creation

SteveMicroNova:
Before:
image

After:
image

### Checklist

### DIFF
--- a/web/src/pages/Settings/Presets/CreatePresetModal/CreatePresetModal.jsx
+++ b/web/src/pages/Settings/Presets/CreatePresetModal/CreatePresetModal.jsx
@@ -14,6 +14,8 @@ import FormControlLabel from "@mui/material/FormControlLabel";
 import Box from "@mui/material/Box";
 import TextField from "@mui/material/TextField";
 
+let is_streamer = false;
+
 const setCheckedRecur = (dict, checked) => {
     dict.checked = checked;
     dict.content.forEach((it) => setCheckedRecur(it, checked));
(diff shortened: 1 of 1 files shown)


web/src/pages/Settings/Presets/CreatePresetModal/CreatePresetModal.jsx
```diff
@@ -108,7 +110,7 @@ const buildTreeDict = (status, showInactive = false) => {
 
     const statusClone = JSON.parse(JSON.stringify(status));
     const sourceDicts = statusClone.sources
-        .filter((source) => showInactive || getSourceInputType(source) !== "none")
+        .filter((source) => showInactive || getSourceInputType(source) !== "none" && getSourceInputType(source) !== "unknown")
```
SteveMicroNova, line 113: This closes #980 
Getting to the bottom of why state isn't being set properly on stream close is a different issue (#981)

web/src/pages/Settings/Presets/CreatePresetModal/CreatePresetModal.jsx
```diff
@@ -123,17 +125,34 @@ const StructuredDictAsTree = ({ dict, depth = 0, path = [] }) => {
...
+            entries.push(
+                <StructuredDictAsTree
+                    key={i}
+                    dict={d}
+                    depth={depth + 1}
+                    path={[...path, i]}
+                />
+            );
+            i += 1;
+        }
```
rtertiaer, line 155:
```suggestion
    let entries = []; 
    let i = 0;
    for (const d of dict.content) {
        let e = <StructuredDictAsTree
            key={i}
            dict={d}
            depth={depth + 1}
            path={[...path, i]} 
        >   
        if(!is_streamer || (is_streamer && depth === 0)) {
            entries.push(e)
        }   
        i += 1;
    }
```

try something like this instead (it's untested)

SteveMicroNova: It is now tested, works like a charm :+1:

rtertiaer (approved):

Issue #907, opened 2024-08-23
Stream selection modal font size too small


SteveMicroNova:
I don't currently know if it was intentional as it __does__ make long stream names fit better, but the font size of the stream selection modal is currently `1rem` instead of `1.5rem`. See attached pics:

Before:
![image](https://github.com/user-attachments/assets/8c899e0f-b6ee-49b2-9fb4-dca17c9f4bac)

After:
![image](https://github.com/user-attachments/assets/3105be8a-5323-4b57-ac04-bccc3b262427)

SteveMicroNova:
Seems it's from a mui package updating, the current state of `main` is uploaded to inhouse and doesn't have it, yet the current state of main on my computer does

rtertiaer:
This is a result of #901 converting the items in this list to an MUI `List`, which applies the typography/theming stuff from `App.jsx`. It's not just the font size - it's also the font itself, which should be Raleway but that typography style specifies as Open Sans

PR #982, opened 2024-10-15
ListFontSizeFix

SteveMicroNova:
During #901, we added MUI theming to our frontend; this automatically styles all MUI components. In doing this, #907 was caused, this fixes that by overriding the default font size applied to many MUI things and allowing us to once again set font size on a per-case basis using CSS

Before:
image

After:
image

### Checklist


klay2000 (approved):
🎉

PR #983, opened 2024-10-15
SpotifyNameFieldDescription

### LINKED ISSUE #948: Spotify demands at least 2 chars in name, name field description is innaccurate
image
The text underneath the name textbox says "This can be anything you like!" but that isn't strictly true; Spotify demands at least two non-special chars (so "1" is invalid, "11" is valid, "1-" is invalid, "1-1" is valid again due to two of the chars being non-special)

I would've whipped up a quick fix for this by doing something like this:
```
const NAME_DESC = ( 
  streamType == "spotify" ? "This name requires a minimum of two non-special characters - it will be used to select this stream from the 
source selection dropdown" 
  :
  "This name can be anything - it will be used to select this stream from the source selection dropdown"
);
```
In StreamModal.jsx, but that file doesn't seem to be aware of the stream type being created

SteveMicroNova:
Add an argument to make StreamModal aware of the stream type being created, add toggle to name field description if the stream type is Spotify

Closes #948

### Checklist

### DIFF
--- a/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
+++ b/web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
@@ -14,8 +14,6 @@ import Avatar from '@mui/material/Avatar';
 import AlertBar from "@/components/StatusBars/AlertBar";
 
 
-const NAME_DESC =
-  "This name can be anything - it will be used to select this stream from the source selection dropdown";
 const DISABLED_DESC = "Don't show this stream in the input dropdown";
 const RESTART_DESC = "Sometimes the stream gets into a bad state and neds to be restarted. If that happened to this stream, click this to restart the stream.";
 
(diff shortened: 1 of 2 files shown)


web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
```diff
@@ -218,7 +216,11 @@ const StreamModal = ({ stream, onClose, apply, del }) => {
         "name": "name",
         "title": "Name",
         "type": "text",
-        "desc": NAME_DESC,
+        "desc": type == "spotify" ?
+            "This name requires a minimum of two non-special characters, and cannot have certain characters such as underscores or slashes - it will be used to select this stream from the source selection dropdown"
+            :
+            "This name can be anything - it will be used to select this stream from the source selection dropdown"
+        ,
```
klay2000, line 223: Adding a special case like this seems wrong, I think the Spotify stream should be changed to re-format an invalid name to something valid.

klay2000 (changes requested):

web/src/pages/Settings/Streams/StreamModal/StreamModal.jsx
SteveMicroNova: I think this is going to need a special case either way, given that "This name can be anything! Oops, not that though." is a bad experience
We should totally still have backend name validation (as suggested in the new issue #986), this PR is just for the accuracy of the description field and I think that's all we strictly need right now to obviate the issue that users could come across

rtertiaer: I agree with @klay2000

SteveMicroNova:
This is too much of a hack and the issue is to low priority to continue in short order. Closing this PR and will perhaps open a new one eventually.

PR #990, opened 2024-10-31
UsageSurvey

SteveMicroNova:
Creating an opt-in device usage survey that will gather statistics about what streams are used and how often, as well as a few hardware statistics that can be used by our dev team to better prioritize bugs and improve features, and by our marketing team to find what's a hit with customers

Currently in a **very** early state, I don't even know what file these basemodels belong in yet or if they should even be part of AmpliPi proper or be more akin to the usb drive mounting and be more of a subprocess off to the side

### Checklist

### DIFF
--- a/amplipi/models.py
+++ b/amplipi/models.py
@@ -21,10 +21,16 @@ Encourages reuse of datastructures across AmpliPi
 
 # type handling, fastapi leverages type checking for performance and easy docs
 from functools import lru_cache
-from typing import List, Dict, Optional, Union, Set
+from typing import List, Dict, Optional, Union, Set, Annotated
 from types import SimpleNamespace
 from enum import Enum
 from pathlib import Path
+
+import os
+import json
+
+from shutil import disk_usage
+from psutil import virtual_memory
 
 # pylint: disable=no-name-in-module
 from pydantic import BaseSettings, BaseModel, Field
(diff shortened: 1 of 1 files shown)


amplipi/models.py
```diff
@@ -110,6 +116,114 @@ class fields_w_default(SimpleNamespace):
...
+
+  # The highest number of concurrent streams running within the survey cycle, constrained to be a number between 0 and 4
+  concurrent_polls: int = 0  # Used to calculate average
+  average_concurrent_streams: Annotated[int, Field(ge=0, le=4)] = 0
+  peak_concurrent_streams: Annotated[int, Field(ge=0, le=4)] = 0
+
+  # The length in seconds of longest-connected instance within the survey cycle, maxing out after 24 hours. Is not additive with multiple concurrent streams.
+  runtime_polls: int = 0  # Used to calculate average
+  average_runtime: Annotated[int, Field(ge=0, le=86400)] = 0
+  peak_runtime: Annotated[int, Field(ge=0, le=86400)] = 0
```
SteveMicroNova, line 147: For runtime, I've got two ideas for measuring:

1. Measure it by recording the time when a stream starts and recording again when it stops and splitting the difference
2. Have a regularly repeated polling function that just records when a given stream and gives a rough number divisible by the 5-10 minute intervals that that polling function is ran, as specificity isn't important and this doesn't require as many stored variables but does require more than 2 polling instances per stream

linknum23 (commented):
Do these changes belong in models.py? Models.py has historically defined the pieces of the API interface.

PR #991, opened 2024-11-07
Upgrade to go-librespot spotify backend

linknum23:
### Checklist


scripts/configure.py
```diff
@@ -994,6 +1016,13 @@ def print_progress(tasks):
       tasks += print_progress(_stop_service('amplipi-updater-test'))
       tasks += print_progress(_remove_service('amplipi-updater-test'))
 
+  # bring up amplipi-tasks
+  tasks += print_progress(_create_service('amplipi-tasks', _tasks_service(env['base_dir']), env))
+  tasks += print_progress(_enable_service('amplipi-tasks'))
+  if not env['is_ci']:
+    tasks += print_progress(_restart_service('redis-server'))  # TODO: is this needed?
```
rtertiaer, line 1023: > is this necessary?

Kinda a complicated question. Strictly necessary? nope. but what if the form & function of the way we use redis changes? restarting is a simple way to wipe everything out, unless you have the on disk persistence enabled (I expect that it is disabled, but it is probably worth checking.) This will also act as a bit of a forcing function for us to never use this as a persistent data store, which I have seen folks do and seems unwise without further justification. With a little thought, I'm in favor of keeping a restart here.

rtertiaer (approved):
:sunglasses:

scripts/configure.py
rtertiaer: though I guess a restart of the appliance as a whole probably accomplishes this too :thinking:

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "amplipi"
-version = "0.4.4"
+version = "0.4.4+96ce6a6-StreamerPresetVolume-dirty"
```
linknum23, line 3: oops

amplipi/ctrl.py
```diff
@@ -561,7 +561,7 @@ def sync_stream_info(self) -> None:
     for sid, stream_inst in self.streams.items():
       # TODO: this functionality should be in the unimplemented streams base class
       # convert the stream instance info to stream data (serialize its current configuration)
-      st_type = type(stream_inst).__name__.lower()
+      st_type = type(stream_inst).stream_type
```
rtertiaer, line 564: :pray:

amplipi/streams/__init__.py
```diff
@@ -31,7 +31,7 @@
 from .internet_radio import InternetRadio
 from .rca import RCA
 from .airplay import AirPlay
-from .spotify import Spotify
+from .spotify_connect import SpotifyConnect
```
rtertiaer, line 34: I thought we had discussed just continuing to use our existing Spotify naming convention, and not calling this SpotifyConnect?

amplipi/streams/spotify_connect.py
```diff
@@ -0,0 +1,198 @@
...
+      source.artist = ", ".join(metadata['track']['artist_names'])
+      if metadata['stopped']:
+        source.state = "stopped"
+        source.supported_cmds = ['play']
+      elif metadata['paused']:
+        source.state = "paused"
+        source.supported_cmds = self.supported_cmds
+      else:
+        source.state = "playing"  # or "unknown"
+        source.supported_cmds = self.supported_cmds
```
rtertiaer, line 159: should the supported_cmds stuff be above line 147, so that even if we fail to munge the rest of the data we can still control the stream?

rtertiaer (approved):
one note, nothing I'd block deployment over.

additionally though, I would *love* to begin using [git lfs](https://git-lfs.com/), if we're gonna continue stuffing binaries in here :woozy_face: ime golang binaries are pretty large by default due to static compilation, and this PR adds 30mb alone (!) to the repo. go-librespot is actively developed, and assuredly we'll want to ship an update at some point. the math is unsettling.

amplipi/streams/__init__.py
linknum23: The main thing I was trying to avoid was API breakage, this was fixed by using the stream_type=spotify. I left the class named spotify connect as a reminder of our intent

linknum23: If that seems too confusing it can be easily reverted.

linknum23:
@rtertiaer as suggested I updated this to use git-lfs so we don't add 35 MB to the repo. Can you try it out?

Adding the guardrails and docs for git-lfs will be a TODO for a different PR.

PR #1004, opened 2024-12-02
Persist Logs during updates

SteveMicroNova:
Adds a function prior to firing the update script that checks if persist_logs is set to true, and if not sets it to true for three days

### Checklist

### DIFF
--- a/amplipi/updater/asgi.py
+++ b/amplipi/updater/asgi.py
@@ -255,11 +255,27 @@ def save_upload_file(upload_file: UploadFile, destination: pathlib.Path) -> None
     upload_file.file.close()
 
 
+def persist_logs_during_update():
+  """Used during system updates to ensure persist logs is activated and has a minimum delay"""
+  persist_data = get_log_persist_state()
+  existing_persist = persist_data.persist_logs
+  existing_delay = persist_data.auto_off_delay
+  # If persist logs is already on and has a larger delay, keep that delay; otherwise ensure it has our sane minimum for support
+  if existing_persist and (existing_delay > 3 or existing_delay == 0):
+    data = Persist_Logs(persist_logs=True, auto_off_delay=existing_delay)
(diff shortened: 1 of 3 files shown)


amplipi/updater/static/index.html
```diff
@@ -44,7 +44,7 @@ <h1><span class="text-white">Ampli</span><span class="text-danger">Pi</span></h1
               <a class="nav-link" id="manual-update-tab" data-toggle="tab" href="#manual-update" role="tab" aria-controls="manual-update" aria-selected="false">Custom Update</a>
             </li>
             <li class="nav-item">
-              <a class="nav-link" id="admin-settings-tab" data-toggle="tab" href="#admin-settings" role="tab" aria-controls="admin-settings" aria-selected="false">Admin Settings</a>
+              <a class="nav-link" id="admin-settings-tab" data-toggle="tab" href="#admin-settings" role="tab" aria-controls="admin-settings" aria-selected="false" onClick="getPersist();">Admin Settings</a>
```
SteveMicroNova, line 47: Now that the state of that tab can be changed by updating, change the update method from `window.onload` to `tab.onclick`

amplipi/updater/asgi.py
```diff
@@ -280,11 +280,27 @@ def download(url, file_name):
...
+    data = Persist_Logs(persist_logs=True, auto_off_delay=3)
+    toggle_persist_logs(data=data)
+
+
 @router.post("/update/download")
 async def download_update(info: ReleaseInfo):
   """ Download the update """
   logger.info(f'downloading update from: {info.url}')
   try:
+    persist_logs_during_update()
```
linknum23, line 303: On second thought We probably need to do this in update/download and update/upload. Those are the two main entry points to start an update.

linknum23 (approved):
This does exactly what I expect. Nice work @SteveMicroNova

PR #1007, opened 2024-12-12
Stat Collector

SteveMicroNova:
Add an optional user statistics survey to help us decide what to spend our devtime on, and potentially help find bugs and improve support

Currently this is a draft that only contains the script that collects the stats, this will eventually contain a good means of invoking the script, a means of sending the data upstream to home base, and a portion of the UI that allows a user to activate the functionality in general

### Checklist

### DIFF
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -60,6 +60,7 @@ from sse_starlette.sse import EventSourceResponse
 import amplipi.utils as utils
 import amplipi.models as models
 import amplipi.defaults as defaults
+import amplipi.statcollector as statcollector
 from amplipi.ctrl import Api, ApiResponse, ApiCode  # we don't import ctrl here to avoid naming ambiguity with a ctrl variable
 from amplipi.auth import CookieOrParamAPIKey, router as auth_router, NotAuthenticatedException, not_authenticated_exception_handler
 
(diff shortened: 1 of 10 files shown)


amplipi/app.py
```diff
@@ -163,6 +164,22 @@ class params(SimpleNamespace):
...
+
+
+@api.post('/collected_stats')
+@api.post('/collected_stats/')
+def post_collected_stats():
+  """Send the contents of the statcollector output file to AmpliPi devs"""
+  survey = statcollector.UsageSurveySchema.load_from_disk()
+  survey.phone_home()
+
+
```
SteveMicroNova, line 182: These do not belong at the top of this file
I do not know where they belong, but it isn't the top of the top
They stay here in the meantime because it's easier to not need to scroll too much to edit imports if need be during devwork

amplipi/statcollector.py
```diff
@@ -0,0 +1,224 @@
...
+    if stream_count == 0 and len(self.active_streams) != 0:  # If a stream has just been shut off, record average based on current runtime
+      new_average = average(self.average_runtime, self.runtime_weight, self.current_runtime)
+      self.average_runtime = new_average
+      self.runtime_weight += 1
+
+    matches = find_matches(self.active_streams, stream_names)
+    if len(matches) != 0:  # If there is at least one playing match, calculate runtime
+      self.current_runtime = self.current_runtime + timediff.seconds
+
+      self.peak_runtime = max(self.peak_runtime, self.current_runtime) if self.current_runtime < 86400 else 86400
```
SteveMicroNova, line 67: There's no real reason to limit this to 24 hours, I just had the feeling that more than 24 hours would lead to outliers and this is a method of negating that. It might be worthwhile to remove the limit just so we can see what sort of lifespan that long-lived stream errors depend on

amplipi/app.py
linknum23: These stats are effectively metadata about the api. Putting them in the API feels weird. That said keeping them outside of the api may make accessing them a little more awkward.

SteveMicroNova:
This is incredibly stale and will be reopened under a new branch if we ever decide to return to it

PR #1009, opened 2024-12-18
Update Airplay icon

SteveMicroNova:
We noted that our airplay logo was out of date
It is now up to date using an official airplay logo

Sources:
[Apple](https://www.apple.com/airplay/#:~:text=Your%20audio.%20Your%C2%A0way.)
[Where I got the icon](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:airplay:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=airplay&icon.size=24&icon.color=%23e8eaed)

Before:
image

After:
image

### Checklist


gorski123:
I think I prefer the circular lines with the triangle, or at the very least the box in blue?
airplay-custom-color-icon-set@2x

web/src/utils/getIcon.jsx
```diff
@@ -2,7 +2,7 @@ import spotify from "@/../static/imgs/spotify.png";
 import dlna from "@/../static/imgs/dlna.png";
 import bluetooth from "@/../static/imgs/bluetooth.png";
 import fmradio from "@/../static/imgs/fmradio.png";
-import shairport from "@/../static/imgs/shairport.png";
+import shairport from "@/../static/imgs/shairport.svg";
```
linknum23, line 5: Any reason this is svg instead of png?

linknum23:
What is the difference in intent between the 2 icons? I assume one makes more sense for amplipi  than the other.

web/src/utils/getIcon.jsx
SteveMicroNova: We use the icons in both large and small formats, svgs are better at not having jagged edges when you use CSS to scale them like that

However, I also grabbed a PNG because without one I'd have to change every other stream icon to SVG as well, and that's a larger lift than this issue deserved at this moment

I've decided to skip having two different image files and just keep it to PNGs

SteveMicroNova:
> What is the difference in intent between the 2 icons? I assume one makes more sense for amplipi than the other.

The triangle with three rings is for audio, the triangle in a rectangle is for audio and video. Not everywhere uses it as that, for example Pandora does not use any video but still uses that icon; I also used that icon because it was what was available in my trusted icon libraries for airplay
I've now replaced it with the Airplay Audio icon that I recolored blue by Jason's request above

SteveMicroNova:
To my understanding, Lincoln is taking on the task of finding a better icon that has better line thickness and a proper color
This PR is likely blocked by that

klay2000 (approved):
According to the [Apple human interface guidelines for Airplay](https://developer.apple.com/design/human-interface-guidelines/airplay) these are the two icons. In quickly skimming through them I didn't see anything specifying that we should use either the radiating circles or rounded rectangle for audio however the consensus seems to be the radiating circles icon is for audio, so I believe it is most correct for us to use that.

As for the line weight, these are the icons apple provides, I doubt we're going to find anything with a finer line and if we do this is still in my opinion the best icon to use.

The color is a bit more complicated, Apple's guidelines basically say to use whatever you use for other services. We use the full color logo for Pandora but for everything else we use logos with at most 2 colors. I think we should use the blue as shown in the screenshots above but Apple seems to not specify what color to use, just that it's colored. Perhaps in the future we should make our iconography a little more concise by finding a 2 color version of the pandora logo and keeping the icons we have control over within a consistent color palate since at the moment some icons have very similar colors and I think this looks bad.

I definitely went a little crazy on this but overall I think we should just stick with the radiating circles icon in blue.

PR #1011, opened 2024-12-19
One-handed mobile control

SteveMicroNova:
Move song controls, preset and add buttons to bottom of the screen on mobile

Adjust player page volume dropdown to more visibly include the underlying components, and limit height to make them scrollable without needing to scroll the whole page

Add many CSS breakpoints to ensure elements continue to work on the smallest screens

Home Page:
Before:
image

After:
image

Player Page:
Before:
image
image

After:
image
image

### Checklist

### DIFF
--- a/web/src/App.scss
+++ b/web/src/App.scss
@@ -1,10 +1,16 @@
 @use "src/general";
 
+.background-gradient {
+  background: general.$bg-gradient;
+  position: fixed;
+  z-index: -1;
+  height: 100vh;
+  width: 100vw;
+}
+
 .app {
   // display: flex;
   // width: 100vw;
-  height: 100vh;  // Required to not have weird paneling with gradient backgrounds
-  background: general.$bg-gradient;
   // padding-top: 0.6rem;
   // padding-bottom: 0.6rem;
   // padding: 0.6rem;
@@ -23,4 +29,39 @@
 
 .app-body {
   padding-bottom: general.$navbar-height;
+}
+
+@media (general.$is-landscape){
+  // Themed scrollbar is overridden by 90% of mobile apps and browsers, but only partially overridden by one (andorid app)
+  // This still looks better on desktop, so gate it to only being on desktop without running the risk for the weird android app situation
+
+  .pill-scrollbar {
+    max-height: inherit;
+    overflow-y: auto;
+    scrollbar-width: thin;
+    scrollbar-color: rgba(0, 0, 0, 0.5) transparent;
+  }
+
+  .pill-scrollbar::-webkit-scrollbar {
+    width: 12px;
+    height: 12px;
+  }
+
+  .pill-scrollbar::-webkit-scrollbar-track {
+    background: transparent;
+  }
+
+  .pill-scrollbar::-webkit-scrollbar-thumb {
+    background-color: rgba(0, 0, 0, 0.5);
+    border-radius: 999px;
+    border: 3px solid rgba(255, 255, 255, 0.5);
+    @supports (-webkit-overflow-scrolling: touch) {
+      border: none;
+    }
+  }
+
+  .pill-scrollbar::-webkit-scrollbar-thumb:hover {
+    background-color: rgba(0, 0, 0, 0.7);
+  }
+
 }
--- a/web/src/components/CustomMarquee/CustomMarquee.jsx
+++ b/web/src/components/CustomMarquee/CustomMarquee.jsx
@@ -42,14 +42,13 @@ export default function CustomMarquee(props) {
         }
     }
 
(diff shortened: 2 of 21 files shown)


SteveMicroNova:
I was far too hasty on this one, on reflection I've noticed that this implementation on the player page literally only works if you don't have volume bars to play with
Revoking requested reviews and setting as draft

web/src/components/CustomMarquee/CustomMarquee.jsx
```diff
@@ -42,14 +42,11 @@ export default function CustomMarquee(props) {
...
 
-    let resizeTimout;
     function handleResize(){
         if(!resizeCooldown.current){
             resizeCooldown.current = true;
 
             assessMarquee()
-
-            resizeTimout = setTimeout(()=>{resizeCooldown.current = false;}, 1000) // set a cooldown for resize checks to avoid excessive renders
         }
```
SteveMicroNova, line 50: I removed unused code from our marquee component because I noticed it during the course of looking at the individual components that made up the player page stream title

SteveMicroNova:
Requesting Klayton and Lincoln as they've both had opinions along the way for this, requesting review from Jason as he requested this feature

SteveMicroNova:
To my end, I have noticed that our unsafe volume sliders (that are very easy to bump and immediately set to max) are something of a hazard when you now have to scroll various nested dropdowns

My current solution has been to limit the length of the volume bars so that there is ample thumb space for scrolling, as a long term goal I'd like to make the volume sliders are made safer from misinput. I've looked into this somewhat, my current thoughts are to either add a [debounce](https://developer.mozilla.org/en-US/docs/Glossary/Debounce) or make it so you have to specifically click and drag the slider knob to change volume on mobile, but I'm undecided

klay2000 (changes requested):
As discussed IRL:
make this only happen on desktop
try to add some drag rejection to the volume sliders
remove scrolling through zones in groups

web/src/App.scss
```diff
@@ -1,10 +1,17 @@
 @use "src/general";
 
+.background-gradient {
+  background: general.$bg-gradient;
+  position: fixed;
+  z-index: -1;
+  height: 100vh;
+  width: 100vw;
+}
```
SteveMicroNova, line 9: I had to remake how the gradient background worked, if you have enough zones when you use the dropdown on the player page you leave the gradient behind as you scroll at 100vh, and at 100% the gradient stretches weird

Now it stays the same no matter where you scroll

web/src/components/MediaControl/MediaControl.scss
```diff
@@ -30,11 +30,11 @@
     padding-right: 1.25rem;
   }
 
-  @media (min-width: 365px) and (max-width: 425px) {
+  @media (min-width: 365px) and (max-width: 435px) {
     font-size: 2.5rem;
   }
 
-  @media (min-width: 425px) {
+  @media (min-width: 435px) {
```
SteveMicroNova, line 37: I found a specific type of apple phone that needed to fit in this bucket, so I expanded the definition of this everywhere

SteveMicroNova: I ought to make this an scss var to avoid such wide-touching bucketing changes, but I'm uncertain I want to bloat this PR any more
If you want that change to make it into this one, say so; otherwise I will make another discreet PR shortly after this is merged.

web/src/components/StreamBar/StreamBar.scss
SteveMicroNova, line 1: Most every change in this file is to ensure our webapp still has the same form and function even on comically tiny screens like the iPhone 4

web/src/general.scss
```diff
@@ -23,6 +23,13 @@ $page-header-height: 3rem;
 // This number is the natural height of the navbar when an explicit height is not provided, and was not set to 56px arbitrarily
 $navbar-height: 56px;
 
+$small-mobile: 365px;
+$medium-mobile: 435px;
+
+// meant to be used directly in @media queries, like @media (general.$is-portrait){stuff}
+$is-portrait: "max-aspect-ratio: 1/1";
+$is-landscape: "min-aspect-ratio: 1/1";
```
SteveMicroNova, line 31: These are functionally equivalent to something like `if 100vh > 100vw` and 1if 100vw >= 100vh` respectively

SteveMicroNova: This is deserved to be reflected as a comment since it was concise and explained it well, so this is now a comment inside the code as well

klay2000 (approved):
As discussed before we should probably clean up some of the scaling, that uses px right now to make handling smaller screens easier, that's for another pr though.

PR #1014, opened 2025-01-07
Fix issue causing status bar to block app controls

SteveMicroNova:
I found that our current StatusBar components would completely block all app controls on mobile, bump the bottom of the StatusBar up above the controls so you can now leave the problem screen

Before:
image

After:
image

### Checklist

### DIFF
--- a/web/src/App.scss
+++ b/web/src/App.scss
@@ -20,3 +20,7 @@
   height: 100vh;
   background-color: green;
 }
+
+.app-body {
+  padding-bottom: general.$navbar-height;
+}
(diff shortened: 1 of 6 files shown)


web/src/components/StatusBars/StatusBar.jsx
```diff
@@ -15,6 +15,7 @@ export default function StatusBar(props) {
 
     return(
         <Snackbar
+            style={{bottom: "60px"}}
```
linknum23, line 18: Is it hard to reference the size of the control bar here?

web/src/App.scss
```diff
@@ -20,3 +20,7 @@
   height: 100vh;
   background-color: green;
 }
+
+.app-body {
+  padding-bottom: general.$navbar-height
+}
```
linknum23, line 26: Missing a newline here. This should be being fixed by vscodes editior config plugin if its installed.

Issue #415, opened 2022-12-15
Update to Python 3.11 to get lots of speed improvements


linknum23:
Maybe we could use pyenv or snap to make the switch.

SteveMicroNova:
Bumping that this is required for #695, we currently run 3.7.3 and the tidalapi package requires 3.9 or higher to function

Going to continue devving on that branch for sure, but this is a blocker to merging in tidal support (once I get to that step)

linknum23:
Just note that this is mostly blocked by changes to how we update.

SteveMicroNova:
Closing as this is superseded by #995

PR #1018, opened 2025-01-10
Pin ubuntu to keep testing against python 3.7

linknum23:


SteveMicroNova (approved):
I really need to get on #415 huh

SteveMicroNova:
Taking the liberty of merging this so I can also merge #1017 faster

PR #1022, opened 2025-01-27
Volume delta fix

### LINKED ISSUE #1021: Volume changes lag with multiple child zones
I've found that in CardVolumeSlider.jsx there is a logic flow that divides the individual zones into discreet api calls, so if you have every zone tuned in to the same source and you use the big volume slider on the home or player page it will go through zones 1-6 one by one. This makes it so that if you change the volume a lot quickly, it functionally queues up a bunch of individual zone volume updates and this can cause the actual auditory volume to lag a few seconds behind the user input and this compounds with every expander they have hooked up to the same source as well

Either batch those requests more efficiently or debounce them a little harder so you don't create this queue in the first place (or a secret third thing I've yet to think of)

SteveMicroNova:
Closes #1021

We had a system of figuring out the delta of every individual zone attached to a source and then forcing those through the source endpoints one by one, and that caused lag. This PR brings that behavior to the backend, handles it with the /zones endpoint so we can do this more effectively and keep the system from lagging due to having a large number of fetch requests if you move the volume bar up and down rapidly.

### Checklist


SteveMicroNova:
This is a draft at the present moment as there is garbage in this PR that I am going to use the github comparison view to clean out

amplipi/ctrl.py
```diff
@@ -847,13 +848,18 @@ def set_vol():
           """ Update the zone's volume. Could be triggered by a change in
               vol, vol_f, vol_min, or vol_max.
           """
-          # 'vol' (in dB) takes precedence over vol_f
+          # Order of operations: vol (db) > vol_delta > vol (float)
```
linknum23, line 851: Is this right? At first glance it looks like the opposite order/precedence where vol_f has the highest priority.

amplipi/ctrl.py
```diff
@@ -847,13 +848,18 @@ def set_vol():
...
           if update.vol_f is not None and update.vol is None:
             vol_db = utils.vol_float_to_db(vol_f, zone.vol_min, zone.vol_max)
             vol_f_new = vol_f
```
linknum23, line 854: oops it looks like this should be clamped right?

amplipi/models.py
```diff
@@ -71,6 +71,8 @@ class fields(SimpleNamespace):
   Volume = Field(ge=MIN_VOL_DB, le=MAX_VOL_DB, description='Output volume in dB')
   VolumeF = Field(ge=MIN_VOL_F, le=MAX_VOL_F,
                   description='Output volume as a floating-point scalar from 0.0 to 1.0 representing MIN_VOL_DB to MAX_VOL_DB')
+  VolumeDeltaF = Field(ge=(MAX_VOL_F * -1), le=MAX_VOL_F,
+                       description='Output volume as a floating-point scalar from -1.0 to 1.0 representing the distance between the current and goal volume')
```
linknum23, line 75: Nitpick but Maybe "adjustment to output volume"? This delta is not a volume by itself.

amplipi/ctrl.py
SteveMicroNova: Delta is not before vol_f due to me moving things around to make tests work, but vol_db is definitely at the top of the stack due both ifs before the else processing `if update.vol is None:`

amplipi/ctrl.py
SteveMicroNova: It wasn't when I got there, but it can be and so I'll update that in next commit

amplipi/ctrl.py
```diff
@@ -847,13 +848,19 @@ def set_vol():
           """ Update the zone's volume. Could be triggered by a change in
               vol, vol_f, vol_min, or vol_max.
           """
-          # 'vol' (in dB) takes precedence over vol_f
-          if update.vol_f is not None and update.vol is None:
-            vol_db = utils.vol_float_to_db(vol_f, zone.vol_min, zone.vol_max)
-            vol_f_new = vol_f
+          # Order of operations: vol (db) > vol_delta > vol (float)
```
linknum23, line 851:
```suggestion
          # Field precedence: vol (db) > vol_delta > vol (float)
          # NOTE: checks happen in reverse precedence to cover default case of unchanged volume
```

SteveMicroNova:
Add tests before moving on

tests/test_rest.py
```diff
@@ -1597,6 +1597,14 @@ def patch_zone(json: Dict, expect_failure: bool = False) -> Optional[Dict]:
   patch_zone({'vol_min': amplipi.models.MIN_VOL_DB, 'vol_max': amplipi.models.MIN_VOL_DB +
              amplipi.models.MIN_DB_RANGE / 2}, expect_failure=True)
 
+  # test that volume delta changes work as expected
+  z = patch_zone({'vol_delta_f': -1.0})
+  assert z['vol_f'] == amplipi.models.MIN_VOL_F
+  z = patch_zone({'vol_delta_f': 1.0})
+  assert z['vol_f'] == amplipi.models.MAX_VOL_F
+  z = patch_zone({'vol_delta_f': -0.5})
+  assert z['vol_f'] == (amplipi.models.MAX_VOL_F + amplipi.models.MIN_VOL_F) / 2
```
SteveMicroNova, line 1606: I'm not experienced with creating tests so it's possible that there's some paradigm or specific intent missing from my addition

I've made tests that pass when the deltas work as they do currently, and don't work if the delta endpoint stops working as it currently does, if there's an changes to be made I may also need to be onboarded into the practices and thought processes that you use when creating tests

SteveMicroNova: `set_zones` has this but `set_group` did not receive the proper plumbing to make this possible. Everything else mentioned has been added in the recent commit.

linknum23: I think you mean the opposite. `set_zones` only has a couple of simple tests and `set_group` has `test_set_group_vol` which looks a lot like the `test_set_zone_vol` testing. 

Both of these use set_zone under the hood, so we can probably get a lot of mileage just testing the basic case of set_group/zones with a volume delta in its zone update and validating that the volumes change as expected. This validates that the plumbing through to set_zone makes basic sense and will validate that this functionality still works in the future..

SteveMicroNova: I meant that the `patch_groups` endpoint does not have the ability to process volume delta changes. The `GroupUpdate` model doesn't have space for it, I don't really want it to either if we're not gonna use it especially because there's already something called `vol_delta` in that model

Currently the only thing that happens if you add the equivalent set of tests to the test_set_group_vol tests is exactly nothing, every assertion fails because nothing is actually ingested when you do something like `g = patch_group({"vol_delta_f": 1.0})`, the returned data will be unchanged from the current group state

linknum23: Oh my mistake, your right a group update doesn't make sense. In that case we just need to add a simple test to `set_zones`.

linknum23:
````python
def test_patch_zones_vol_delta(client):
  """ Try changing multiple zones volume using volume delta """
  zones = [z for z in base_config()['zones']]
  # set each zone to a random initial volume
  for z in zones:
    z['vol_f'] = random.uniform(0.0, 0.9)  # random initial volume
    rv = client.patch('/api/zones/{}'.format(z['id']), json={'vol_f': z['vol_f']})
    assert rv.status_code == HTTPStatus.OK
  # update each zones volume by 10%
  rv = client.patch('/api/zones', json={'zones': [z['id'] for z in zones], 'update': {'vol_delta_f': 0.1}})
  assert rv.status_code == HTTPStatus.OK
  jrv = rv.json()
  assert len(jrv['zones']) >= 6
  # check that each update worked as expected
  for z in jrv['zones']:
    if z['id'] in range(6):
      assert z['vol_f'] - (zones[z['id']]['vol_f'] + 0.1) < 0.0001
```

linknum23: Something like that would fit my expectations

SteveMicroNova: I've added this along with equivalents of all other tests I've added to the set_zone

linknum23 (approved):

PR #1023, opened 2025-01-29
Add default streams for first startup

SteveMicroNova:
Sets system default state to Groove Salad connected to all zones at 0 volume and muted everywhere, as well as Spotify and Airplay2 sources connected to no zones for first startup
Can be tested by going to Setttings -> Config -> Factory Reset

### Checklist

### DIFF
--- a/amplipi/defaults.py
+++ b/amplipi/defaults.py
@@ -19,10 +19,10 @@ USER_CONFIG_DIR = os.path.join(os.path.expanduser('~'), '.config', 'amplipi')
 DEFAULT_CONFIG = {  # This is the system state response that will come back from the amplipi box
   "version": 1,
   "sources": [  # this is an array of source objects, each has an id, name, type specifying whether source comes from a local (like RCA) or streaming input like pandora
-    {"id": 0, "name": "Input 1", "input": ""},
-    {"id": 1, "name": "Input 2", "input": ""},
-    {"id": 2, "name": "Input 3", "input": ""},
-    {"id": 3, "name": "Input 4", "input": ""},
+    {"id": 0, "name": "Output 1", "input": ""},
+    {"id": 1, "name": "Output 2", "input": ""},
+    {"id": 2, "name": "Output 3", "input": ""},
+    {"id": 3, "name": "Output 4", "input": ""},
   ],
   # NOTE: streams and groups seem like they should be stored as dictionaries with integer keys
   #       this does not make sense because JSON only allows string based keys
@@ -34,6 +34,10 @@ DEFAULT_CONFIG = {  # This is the system state response that will come back from
     {"id": RCAs[3], "name": "Input 4", "type": "rca", "index": 3, "browsable": False, "disabled": False},
     {"id": 1000, "name": "Groove Salad", "type": "internetradio", "url": "http://ice6.somafm.com/groovesalad-32-aac",
      "logo": "https://somafm.com/img3/groovesalad-400.jpg", "browsable": False, "disabled": False},
+    {"id": 1001, "name": "AmpliPro 1", "type": "spotify", "disabled": False, "browsable": False},
+    {"id": 1002, "name": "AmpliPro 2", "type": "spotify", "disabled": False, "browsable": False},
+    {"id": 1003, "name": "AmpliPro 1", "type": "airplay", "disabled": False, "ap2": False, "browsable": False},
+    {"id": 1004, "name": "AmpliPro 2", "type": "airplay", "disabled": False, "ap2": True, "browsable": False}
   ],
   "zones": [  # this is an array of zones, array length depends on # of boxes connected
     {"id": 0, "name": "Zone 1", "source_id": 0, "mute": True, "disabled": False,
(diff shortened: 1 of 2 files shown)


linknum23:
Wasn't the plan to also have 2 Spotify and 2 Airplay streams added out of the box. Additionally there would be one of each of those that was connected to a source but not connected to any zones.

linknum23:
I have this debate as to whether it makes sense to start with some small volume instead of being completely muted.

SteveMicroNova:
> Wasn't the plan to also have 2 Spotify and 2 Airplay streams added out of the box. Additionally there would be one of each of those that was connected to a source but not connected to any zones.

I feel that we've gone back and forth on this, so I did the smallest lift. I can happily go back to the airplay+spotify version of this intent

> I have this debate as to whether it makes sense to start with some small volume instead of being completely muted.

I'd argue completely muted, I don't want to be responsible for figuring out what is a good volume to have it audible on quiet speakers but not explode loud ones before they can dial in their DB gate settings.

amplipi/defaults.py
```diff
@@ -19,10 +19,10 @@
 DEFAULT_CONFIG = {  # This is the system state response that will come back from the amplipi box
   "version": 1,
   "sources": [  # this is an array of source objects, each has an id, name, type specifying whether source comes from a local (like RCA) or streaming input like pandora
-    {"id": 0, "name": "Input 1", "input": ""},
-    {"id": 1, "name": "Input 2", "input": ""},
-    {"id": 2, "name": "Input 3", "input": ""},
-    {"id": 3, "name": "Input 4", "input": ""},
```
SteveMicroNova, line 25: I don't think that these being called inputs is used anywhere, but they _are_ outputs so I made sure they were named that in line with the streamer section as well
Correct me if needed

linknum23:
We could potentially setup 3 sources in the cleanup script. One connecting groove salad to the first source with all 6 zones, with the 2 other sources connected to a spotify and an airplay stream respectively.

SteveMicroNova:
I've updated the cleanup script and ran it against my dev unit, this is the result:
image

linknum23:
Looks good. Any reason these streams shouldn't each be named AmpliPi 1 and AmpliPi 2? Or maybe AmpliPro 1 and AmpliPro 2? I would expect them to be called be the name of the device initially when trying to find them on Spotify or Airplay. Most people will have several spotify/airplay capable devices at home and this will make them easy to identify initially.
@gorski123  any thoughts are what name they should show up as?

SteveMicroNova:
> Looks good. Any reason these streams shouldn't each be named AmpliPi 1 and AmpliPi 2? Or maybe AmpliPro 1 and AmpliPro 2?

We add 4 new streams in this PR: SpotiPi 1, SpotiPi 2, AmpliPlay 1, and AmpliPlay 2. The ones that connect by default are SpotiPi 1 (because that's the first one) and AmpliPlay 2 (because there is one that is Airplay 1 and one that is Airplay 2, and having those numbers not match up seemed bothersome)
They're named what they are because I wanted to cleverly say "This is this stream time on your AmpliPi" with only one word to ensure the name would fit on the devices you're connecting from

linknum23:
Yeah I get that. I guess I just have a preference for the Spotify Streams to be name `AmpliPi 1`, `AmpliPi 2` and also have the Airplay streams be name `AmpliPi 1` and `AmpliPi 2`. I realize the names clash but I fee like that may give the user a less clunky experience. I realize this is an opinion which is why I was asking for input from Jason. I should probably as @klay2000 as well.

SteveMicroNova:
> Airplay streams show up in Spotify too on IOS right? If I'm remembering that correctly you should def not have any overlap.

image

No, Airplay does show up but you have to open it as a separate window so that should be safe

klay2000:
Oh ok, call it Amplipro then?

SteveMicroNova:
Jason has also said "AmpliPro 1 and 2 is fine for all of those" so that seems to be the way forward
I've made this change, so now that we have consensus this is due to be approved I'd think

amplipi/defaults.py
```diff
@@ -34,6 +34,10 @@
     {"id": RCAs[3], "name": "Input 4", "type": "rca", "index": 3, "browsable": False, "disabled": False},
     {"id": 1000, "name": "Groove Salad", "type": "internetradio", "url": "http://ice6.somafm.com/groovesalad-32-aac",
      "logo": "https://somafm.com/img3/groovesalad-400.jpg", "browsable": False, "disabled": False},
+    {"id": 1001, "name": "AmpliPi 1", "type": "spotify", "disabled": False, "browsable": False},
+    {"id": 1002, "name": "AmpliPi 2", "type": "spotify", "disabled": False, "browsable": False},
+    {"id": 1003, "name": "AmpliPi 1", "type": "airplay", "disabled": False, "ap2": False, "browsable": False},
+    {"id": 1004, "name": "AmpliPi 2", "type": "airplay", "disabled": False, "ap2": True, "browsable": False}
```
linknum23, line 40: These should probably be called AmpliPro 1/2 as well right?

SteveMicroNova: Good catch, solved in recent commit

linknum23 (approved):

PR #1015, opened 2025-01-09
Preset and add creation

SteveMicroNova:
### What does this change intend to accomplish?
Closes #707

Moved stream and preset creation workflows to the components folder, then spread them to relevant places (preset creation on presets settings page and presets modal on home page, stream creation in all places streams can be selected)

### Checklist

* [x] Have you tested your changes and ensured they work?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [x] If applicable, have you updated the documentation/manual?
* [x] If applicable, have you updated the CHANGELOG?
* [x] Does your submission pass linting & tests? You can test on localhost using `./scripts/test`
* [x] If this is a UI change, have you tested it across multiple browser platforms on both desktop and mobile?

PR #1026, opened 2025-02-06
Add shorthand button creation to ModalCard

SteveMicroNova:
This is a sister PR to #1015 and so will be merged into #1015 if this merges first, and into main if this merges second

This is not directly a part of #1015 because the amount of replumbing that this entailed seemed like any issues this causes would easily become a needle in a haystack

### Checklist

### DIFF
--- a/web/src/pages/Settings/Groups/GroupModal/GroupModal.jsx
+++ b/web/src/pages/Settings/Groups/GroupModal/GroupModal.jsx
@@ -26,14 +26,10 @@ const GroupModal = ({ group, zones, onClose, del, apply }) => {
         <ModalCard
             onClose={onClose}
             header="Edit Group"
-            onAccept={() => {
-                apply(groupName, groupZones);
-                onClose();
-            }}
-            onDelete={() => {
-                if (del) del();
-                onClose();
-            }}
+            buttons={[
+                [ "Confirm", () => { apply(groupName, groupZones); onClose(); } ],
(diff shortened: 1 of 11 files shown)


SteveMicroNova:
Only asking Klayton to review this time as he has opinions on how to React
I think that my solution is potentially gross, but is good for dynamically creating any number of buttons without needing to import new components on every page that uses the ModalCard

SteveMicroNova:
image

web/src/pages/Settings/Groups/GroupModal/GroupModal.jsx
```diff
@@ -26,14 +26,10 @@ const GroupModal = ({ group, zones, onClose, del, apply }) => {
...
-                onClose();
-            }}
-            onDelete={() => {
-                if (del) del();
-                onClose();
-            }}
+            buttons={[
+                [ "Confirm", () => { apply(groupName, groupZones); onClose(); } ],
+                [ del ? "Delete" : "Cancel", () => { if (del) del(); onClose(); } ]
+            ]}
```
SteveMicroNova, line 32: This right here is why there isn't a default close button on the ModalCard, sometimes it has to be delete instead

I do not want to add a check in ModalCard for whether or not there is a delete button before displaying such a default close button either, so the passing of onClose buttons will continue.

klay2000 (approved):
As discussed irl, its a little gross but since we're being more explicit with the button text we kind of have to set it everywhere. With the padding added below the button I think it looks pretty decent, approving.

PR #1036, opened 2025-03-06
Hide "Reset Stream" button when creating streams

SteveMicroNova:
Before:
image

After:
image

### Checklist

### DIFF
--- a/web/src/components/CreateStreamModal/StreamModal/StreamModal.jsx
+++ b/web/src/components/CreateStreamModal/StreamModal/StreamModal.jsx
@@ -376,14 +376,15 @@ const StreamModal = ({ stream, onClose, apply, del }) => {
                     }}
                 />
 
-                <ButtonField
+                {/* del is only used during edit, so use that to define when the modal should be in edit mode */}
+                {del && <ButtonField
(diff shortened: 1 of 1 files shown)


web/src/components/CreateStreamModal/StreamModal/StreamModal.jsx
```diff
@@ -376,14 +376,14 @@ const StreamModal = ({ stream, onClose, apply, del }) => {
...
-                <ButtonField
+                {del && <ButtonField
                     name="Restart"
                     text="Restart Stream"
                     desc={RESTART_DESC}
                     onClick={() => fetch("/api/streams/" + stream.id + "/restart",{
                         method: "POST"
                     })}
-                />
+                />}
```
SteveMicroNova, line 386: We are currently using `del` as a means of checking if the modal should be in create or edit mode, as only the edit version passes a function to delete the stream

See line 242 for the other instance of this flag being checked for this purpose

linknum23: This comment belongs in the code above

PR #1041, opened 2025-04-01
Add support for home assistant's STATE_OFF

SteveMicroNova:
Support `STATE_OFF` for use with home assistant media_player entities

### Checklist

### DIFF
--- a/amplipi/models.py
+++ b/amplipi/models.py
@@ -53,6 +53,17 @@ MAX_SOURCES = 4
 SOURCE_DISCONNECTED = -1
 """ Indicate no source connection, simulated in SW by muting zone for now """
 
+ZONE_OFF = -2
+"""
+Indicate that a zone is considered off for the purpose of external interfaces such as home assistant
+
+ON = source_id != ZONE_OFF
+OFF = source_id == ZONE_OFF
+
+To turn off: set to ZONE_OFF
+To turn on: set to SOURCE_DISCONNECTED or any valid source_id
+"""
+
 
 def pcnt2Vol(pcnt: float) -> int:
   """ Convert a percent to volume in dB """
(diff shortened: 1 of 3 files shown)


SteveMicroNova:
I forget if this needed new tests to be made for it since we only talked about this verbally a few days ago and I have since fed my brain into a taffy puller with the various things I've been touching lately

Let me know what changes this needs, if any, and I'll make a rapid release so it can be used with the upcoming hacs_amplipi update

amplipi/models.py
```diff
@@ -53,6 +53,9 @@
 SOURCE_DISCONNECTED = -1
 """ Indicate no source connection, simulated in SW by muting zone for now """
 
+ZONE_OFF = -2
+""" Indicate that a zone is set to STATE_OFF within home assistant """
```
linknum23, line 57:
```suggestion
""" Indicate that a zone is set to STATE_OFF used for tracking a power off state for external interfaces that treat a zone as a media player with an on/off button. Used for integrations such as home assistant. Use the following mapping to use as an on/off indicator:
- ON: source_id != ZONE_OFF
- OFF: source_id == ZONE_OFF
To turn off: set to ZONE_OFF. 
To turn back on: set to SOURCE_DISCONNECTED (to wait for the user to change an input) or set to any valid source id to start playing something. 
 """
```

linknum23: That probably needs a touch of rework but gets the general concept past.

SteveMicroNova: I've added a less verbose version in my recent commit, see here:

```
"""
Indicate that a zone is considered off for the purpose of external interfaces such as home assistant

ON = source_id != ZONE_OFF
OFF = source_id == ZONE_OFF

To turn off: set to ZONE_OFF
To turn on: set to SOURCE_DISCONNECTED or any valid source_id
"""
```

linknum23 (approved):
As discussed this looks good!

PR #1058, opened 2025-09-24
Other Releases

SteveMicroNova:
The Older Releases tab in the Admin Panel has had a bad name for a long time, given that you can get the current and upcoming (beta) versions from there as well; I've finally done something about it by simply doing a "find and replace in all files" operation


SteveMicroNova:
This might technically be a bad name due to it including the current release and the name "Other Releases" implies "Other than the Latest Release", this could just as well be called the All Releases tab

linknum23 (approved):

PR #1063, opened 2025-10-31
Add volume sync between Spotify and AmpliPi

### LINKED ISSUE #1054: Spotify-side volume changes are not realized
Previously when we were running [spotifyd](https://github.com/Spotifyd/spotifyd), Spotify volume controls functioned
spotifyd decayed over time so we swapped to [go-librespot](https://github.com/devgianlu/go-librespot), which does support volume changes but they sound notably worse than the same implementation in spotifyd so we made the decision to deactivate Spotify side volume support for now

This is while we await go-librespot's development so that they reach a point where we can have _volume propagation_, a feature that will allow us to merge the volume of Spotify and AmpliPi on those streams, making it so that they move as a single unit rather than requiring users to use both the Spotify and amplipi volume sliders to dial in their volume properly.

For devs, this issue is:
- Keep an eye on go-librespot's development
- add volume propagation functionality with go-librespot once available

### LINKED ISSUE #1067: Vsrc values drift on system reboot
When you use a stream that uses a `vsrc` value, it will go to one of the sources in `[v0, v1, v2, v3]`. If you leave one such stream open when rebooting the system, the list of `vsrc`s drifts to `[v1, v2, v3, v4]`, leaving `v0` inaccessible and creating a nonexistent `v4`

This isn't a bug that breaks anything _except_ for any streams that uses a subprocess to keep an eye on the amplipi-side state such as the upcoming #1063 that adds a listener to Spotify streams to watch the volume(s) of the connected zone(s) to synchronize the volume across both the Spotify client and AmpliPi, I suspect a similar method would be used for airplay as well

SteveMicroNova:
Make it so that Spotify side volume changes are reflected in AmpliPi and vice-versa

### Checklist

### DIFF
--- a/amplipi/streams/base_streams.py
+++ b/amplipi/streams/base_streams.py
@@ -5,6 +5,7 @@ from typing import Optional, List
 import logging
 from amplipi import models
 from amplipi import utils
+from amplipi import app
 
 logger = logging.getLogger(__name__)
 logger.level = logging.DEBUG
@@ -61,6 +62,24 @@ class BaseStream:
     self.browsable = isinstance(self, Browsable)
     if validate:
       self.validate_stream(name=name, mock=mock, **kwargs)
+
+  def get_zone_data(self):
+    if self.src is not None:
+      ctrl = app.get_ctrl()
+      state = ctrl.get_state()
+      return [zone for zone in state.zones if zone.source_id == self.src]
+
+  @property
+  def connected_zones(self) -> List[int]:
+    connected_zones = self.get_zone_data()
+    return [zone.id for zone in connected_zones]
+
+  @property
+  def volume(self) -> float:
+    connected_zones = self.get_zone_data()
+    if connected_zones:
+      return sum([zone.vol_f for zone in connected_zones]) / len(connected_zones)
+    return 0
 
   def __del__(self):
     self.disconnect()
@@ -242,7 +261,7 @@ class PersistentStream(BaseStream):
       raise Exception(f'Failed to deactivate {self.name}: {e}') from e
     finally:
       self.state = "disconnected"  # make this look like a normal stream for now
-      if 'vsrc' in self.__dir__() and self.vsrc:
+      if 'vsrc' in self.__dir__() and self.vsrc is not None:
         vsrc = self.vsrc
         self.vsrc = None
         vsources.free(vsrc)
--- a/amplipi/streams/spotify_connect.py
+++ b/amplipi/streams/spotify_connect.py
@@ -2,18 +2,26 @@
 
 import io
 import os
+import threading
 import re
 import sys
 import subprocess
 import time
 from typing import ClassVar, Optional
 import yaml
+import logging
+import json
 from amplipi import models, utils
-from .base_streams import PersistentStream, InvalidStreamField, logger
+from .base_streams import PersistentStream, InvalidStreamField
 from .. import tasks
 
 # Our subprocesses run behind the scenes, is there a more standard way to do this?
 # pylint: disable=consider-using-with
+
+logger = logging.getLogger(__name__)
+logger.level = logging.DEBUG
+sh = logging.StreamHandler(sys.stdout)
+logger.addHandler(sh)
 
 
 class SpotifyConnect(PersistentStream):
@@ -33,9 +41,30 @@ class SpotifyConnect(PersistentStream):
     self._log_file: Optional[io.TextIOBase] = None
     self._api_port: int
     self.proc2: Optional[subprocess.Popen] = None
+    self.volume_sync_process: Optional[subprocess.Popen] = None  # Runs the actual vol sync script
+    self.volume_watcher_process: Optional[threading.Thread] = None  # Populates the fifo that the vol sync script depends on
+    self.src_config_folder: Optional[str] = None
     self.meta_file: str = ''
-    self.max_volume: int = 100  # default configuration from 'volume_steps'
-    self.last_volume: float = 0
+    self._volume_fifo = None
+
+  def watch_vol(self):
+    """Creates and supplies a FIFO with volume data for volume sync"""
+    while True:
+      try:
+        if self.src is not None:
+          if self._volume_fifo is None and self.src_config_folder is not None:
+            fifo_path = f"{self.src_config_folder}/vol"
+            if not os.path.isfile(fifo_path):
+              os.mkfifo(fifo_path)
+            self._volume_fifo = os.open(fifo_path, os.O_WRONLY, os.O_NONBLOCK)
+          data = json.dumps({
+            'zones': self.connected_zones,
+            'volume': self.volume,
+          })
+          os.write(self._volume_fifo, bytearray(f"{data}\r\n", encoding="utf8"))
+      except Exception as e:
+        logger.error(f"{self.name} volume thread ran into exception: {e}")
+      time.sleep(0.1)
 
   def reconfig(self, **kwargs):
     self.validate_stream(**kwargs)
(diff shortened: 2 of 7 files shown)


streams/spot_connect_meta.py
```diff
@@ -155,76 +156,94 @@ def from_json(json_data: dict) -> "Event":
...
     elif e.event_type == "volume":
       e.data = VolumeChange(**json_data["data"])
+      volume = e.data.value
+      if volume != parent.last_volume:
+        url = f"http://localhost:{parent._api_port}"  # TODO: Implement spotify-side volume bar to reflect the source volume bar's position
+        # tasks.post.delay(url + 'volume', data={'volume': int(volume * 100)})
+        delta = float((volume - parent.last_volume) / 100)
+        vol_change = requests.patch("http://localhost/api/zones", json={"zones": [0, 1, 2, 3, 4, 5], "update": {"vol_delta_f": delta}}, timeout=5)  # TODO: set zone list correctly
+        if vol_change.ok:
+          parent.last_volume = volume  # update last_volume for future syncs
```
SteveMicroNova: A Spotify volume change is perceived as an event change just like play/pause changes, given that this file is where events are listened to it felt like the right place to add this functionality but on the other hand given that this is a metadata reader file it likely shouldn't be handling actual system state changes

amplipi/streams/spotify_connect.py
```diff
@@ -99,20 +100,30 @@ def _activate(self, vsrc: int):
     logger.info(f'{self.name}: starting metadata reader: {meta_args}')
     self.proc2 = subprocess.Popen(args=meta_args, stdout=self._log_file, stderr=self._log_file)
 
+    vol_sync = f"{utils.get_folder('streams')}/spotify_volume_handler.py"
+    vol_args = [sys.executable, vol_sync, str(self._api_port), str(self.vsrc), "--debug"]
```
SteveMicroNova: the `self.vsrc` is me attempting to pass the `source_id` down to the volume handler so it knows what `source_id` to look for amongst all zones when reading the amplipi-side volume

It appears that `vsrc` isn't always equal to the `source_id`, such as when a spotify stream is open when the amplipi restarts it will be the same as `source_id +1`, which makes the handler not work. This is not solveable by simply closing and reopening the stream either, as it still goes to that `source_id+1` `vsrc`

Is this a bug with how `vsrc` is handled within amplipi, or is it an issue with how I'm attempting to use it?

SteveMicroNova:
I'd like to take note that we stress tested this earlier with multiple people playing tug of war with the volume slider using a Spotify jam and the underlying Spotify stream failed before the script did, according to the logs

It's unclear how the underlying stream failed, and perhaps this is solved with the bump from go-librespot `0.5.0` to `0.5.2`, but it's hard to say until we put it through another office-wide stress test (I simply am not able to input enough changes myself even though I am able to create a multi account jam session)

linknum23:
One thing that we should do is test what the update speed is by logging all
of the requests. Realistically we may want to limit the response to 10Hz or
something just to avoid overwhelming the Spotify and amplipi APIs.

streams/volume_synchronizer.py
```diff
@@ -0,0 +1,134 @@
...
+        elif event == "amplipi_volume_changed":
+          self.vol_set_point = self.stream.set_vol(self.amplipi.volume, self.vol_set_point)
+      except queue.Empty:
+        continue
+      except (KeyboardInterrupt, SystemExit):
+        self.logger.exception("Exiting...")
+        break
+      except Exception as e:
+        self.logger.exception(f"Exception: {e}")
+        continue
```
SteveMicroNova: > One thing that we should do is test what the update speed is by logging all of the requests. Realistically we may want to limit the response to 10Hz or something just to avoid overwhelming the Spotify and amplipi APIs.


There is no such gating in this program but you're also not able to make changes any faster than what you can do in the spotify app by default, I've never been rate limited in any way during testing

amplipi/streams/spotify_connect.py
```diff
@@ -33,9 +41,30 @@ def __init__(self, name: str, disabled: bool = False, mock: bool = False, valida
...
+              os.mkfifo(fifo_path)
+            self._fifo = os.open(fifo_path, os.O_WRONLY, os.O_NONBLOCK)
+          data = json.dumps({
+            'zones': self.connected_zones,
+            'volume': self.volume,
+          })
+          os.write(self._fifo, bytearray(f"{data}\r\n", encoding="utf8"))
+      except Exception as e:
+        logger.error(f"{self.name} volume thread ran into exception: {e}")
+      time.sleep(1)
```
SteveMicroNova: https://github.com/micro-nova/AmpliPi/pull/1063/files#r2578774743

Though this here realistically gates the spotify update cadence to 10Hz / 1 per second

amplipi/streams/airplay.py
```diff
@@ -27,8 +27,8 @@ class AirPlay(PersistentStream):
 
   stream_type: ClassVar[str] = 'airplay'
 
-  def __init__(self, name: str, ap2: bool, disabled: bool = False, mock: bool = False, validate: bool = True):
-    super().__init__(self.stream_type, name, disabled=disabled, mock=mock, validate=validate)
+  def __init__(self, name: str, stream_id: int, ap2: bool, disabled: bool = False, mock: bool = False, validate: bool = True):
```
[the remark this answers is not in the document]
SteveMicroNova: Nope, will remove in next commit

web/src/pages/Home/Home.jsx
```diff
@@ -119,7 +119,7 @@ const Home = () => {
                     // on apply, we want to call
                     onApply={async (customSourceId) => {
                         const ret = await executeApplyAction(customSourceId);
-                        if(ret.ok){ret.json().then(s => setSystemState(s))};
```
linknum23: Whats going on here?

streams/spotify_volume_handler.py
```diff
@@ -0,0 +1,86 @@
...
+from spot_connect_meta import Event
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+sh = logging.StreamHandler(sys.stdout)
+logger.addHandler(sh)
+
+
+class SpotifyData(StreamData):
```
SteveMicroNova: Rename to watcher, rename other classes and variables to more intuitive names

streams/volume_synchronizer.py
```diff
@@ -0,0 +1,198 @@
...
+        def set_vol(self, new_vol: float, vol_set_point: float):
+          ...
+
+        {
+          build argsparser here
+        }
+
+        handler = VolumeSynchronizer(SomeStream(**kwargs), args.config_dir, args.debug)
+        handler.watcher_loop()
+  """
```
SteveMicroNova: Simplify the length of this a little bit

SteveMicroNova:
Note that the vol sync will not work on systems that are effected by #1067, which are all systems that reboot while playing a persistent stream

amplipi/streams/base_streams.py
```diff
@@ -242,7 +262,7 @@ def deactivate(self):
       raise Exception(f'Failed to deactivate {self.name}: {e}') from e
     finally:
       self.state = "disconnected"  # make this look like a normal stream for now
-      if 'vsrc' in self.__dir__() and self.vsrc:
+      if 'vsrc' in self.__dir__() and self.vsrc is not None:
```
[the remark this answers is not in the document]
linknum23: Just make sure this doesnt get squashed into the rest. It should be a separate commit.

SteveMicroNova:
With #1022 merged and #1092 open, I've moved bases from the former to the latter
I've retested to make sure it works and while it looks pretty good, I would like to put this on inhouse for a bit before merging

Issue #1079, opened 2026-02-04
Add an "Authoritative" vs "Relative" volume change setting to groups


SteveMicroNova:
Currently, groups commit authoritative volume changes, that is they forcibly change all underlying zones to be the same volume value. This is good for when a group is a series of zones that are actually one room as this makes rooms with one zone and rooms with multiple zones have the same overall control experience, but it can make less sense when you have an "Everywhere" style group

Add a toggle to set individual groups to have relativistic volume controls in the group settings, that way you can match how a source/stream handles volume if you so choose.

This could potentially remove the need of solving #1076

PR #1062, opened 2025-10-24
Volume Delta Expansion

SteveMicroNova:
### What does this change intend to accomplish?
This PR aims to preserve the relative distance between zone volume levels when they would otherwise exceed the min or max vol_f

[Demonstration](https://amplipi.discourse.group/t/new-amplipi-software-release-0-4-11/413/2?u=amplipi)

### Checklist

* [x] Have you tested your changes and ensured they work?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
* [ ] If applicable, have you updated the documentation/manual?
* [x] If applicable, have you updated the CHANGELOG?
* [x] Does your submission pass linting & tests? You can test on localhost using `./scripts/test`
* [x] Have you written new tests for your core features/changes, as applicable?
* [x] If this is a UI change, have you tested it across multiple browser platforms on both desktop and mobile?

PR #1081, opened 2026-02-11
Player zone add

### LINKED ISSUE #1076: Never hide the source/stream/parent volume bar in the player view
Currently the volume bar that shows on the player view runs through a check to see if there is only one group connected, and if yes it shows that group as the parent bar instead of the bar that has the same functionality as the bars on the home screen do.

This was a value add back when those bars did the same thing, now there's no good way to change volumes in a relativistic manner if you want the groups volumes to not be synced to the same point

To maintain the added value while also restoring relativistic vol change functionality, I'd suggest potentially just having the sole group be pre-expanded to provide the same minimal click/tap per operation rate that the volume bar check was added for

### LINKED ISSUE #1080: Add way of adding zones from the player view
You can do nearly every operation you could possibly want to do from the player page now EXCEPT add or remove zones and groups

My immediate idea for how to do this would be to just add an instance of the [  +  ] button from the home page to the dropdown of the main volume bar, though that isn't the required solution to this issue.

SteveMicroNova:
Add the ability to change zones and groups from the player page using new button

Closes #1080 and #1076, potentially removes need for #1079 thanks to the relativistic volume bar always being there now.

### Checklist

### Desktop:

* No Zones/Groups image
* One Zone image

### Mobile:
* No Zones/Groups image
* One Zone image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
   * Upgraded volume calculations to preserve relative positions when hitting the min or max setting via source volume bar
   * Update our spotify provider `go-librespot` to `0.7.3`
   * Upgrade from Logitech Media Server 8.5.2 to Lyrion Music Server 9.0.3
+* Web App
+  * Added [  +  ] button to player page to allow selection of zones and groups from player page
 
 # 0.4.11
 * System
--- a/web/src/components/RectangularButton/RectangularButton.jsx
+++ b/web/src/components/RectangularButton/RectangularButton.jsx
@@ -11,7 +11,7 @@ export default function RectangularButton({onClick, large, label}){
         >
             {label}
         </div>
-    )
+    );
 }
 RectangularButton.propTypes = {
     onClick: PropTypes.func.isRequired,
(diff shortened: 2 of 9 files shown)


SteveMicroNova:
My local main appears to have been poisoned with #1062 somehow, let me sort that out before looking at this
Only the last 5 commits are actually meant to be in this PR (starting from https://github.com/micro-nova/AmpliPi/pull/1081/changes/2d4199b0eb27eee084f4b0e2a6fe3aaaad35430e)

klay2000 (commented):
This looks fine, I will approve it once the bug with relative volume changes is fixed.

SteveMicroNova:
> This looks fine, I will approve it once the bug with relative volume changes is fixed.

To translate for any readers, this is until #1062 is cleaned up and merged

SteveMicroNova:
> > This looks fine, I will approve it once the bug with relative volume changes is fixed.
> 
> To translate for any readers, this is until #1062 is cleaned up and merged

This hurdle has been surpassed, rerequesting review on that basis @klay2000

PR #1083, opened 2026-02-24
Deactivate frontend caching via the backend

SteveMicroNova:
Deactivates caching of the frontend css and javascript files to avoid users being stuck with stale code, particularly fetch requests that no longer match backend endpoints

Will require users to clear cache one time post release but never again afterwards

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # Future Release
 
+* Web App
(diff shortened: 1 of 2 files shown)


linknum23:
How was this tested?

linknum23:
Can you validate this works on android/ios as well.

PR #1087, opened 2026-03-02
Bump go-librespot to version 0.7.1

SteveMicroNova:
A simple golibrespot version bump. I don't think this will solve the ongoing spotify issues if the issue is what we've thought it is thus far (https://amplipi.discourse.group/t/spotify-not-connecting/403/4?u=amplipi) but it never hurts to bump, either; especially when there's been 2 major bumps since our last upgrade

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # AmpliPi Software Releases
 
 # Future Release
-
+* System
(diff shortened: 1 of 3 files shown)


SteveMicroNova:
deployed to our in-office unit now and doesn't seem to have any issues. Continuously uncertain this solves any problems, but again not a bad thing to do regardless

linknum23 (approved):

PR #1088, opened 2026-03-10
Replace github sponsor link to be our online store

SteveMicroNova:
I noticed our sponsor link goes to our long-defunct kickstarter, I've replaced the link with a place people can actually give us money if they so choose.


linknum23 (approved):
Seems reasonable. Tested the link.

PR #1091, opened 2026-05-26
Error Surfacing

### LINKED ISSUE #539: Reset Preamp Micro on Communication Failure
If the preamp board's microcontroller fails to communicate, try resetting the firmware and check if communication works after a reset. If a failure still occurs after a reset, it might also be a good idea to show the failure on the web UI.

### LINKED ISSUE #643: Detect communication failures and attempt graceful HW reset
### Context
The AmpliPi application running on the Pi does not have direct control over the zone inputs and volumes. It uses an I2C interface to send commands to firmware running on the preamp board. The firmware running on the preamp board controls the zone volumes and inputs over a separate I2C bus. 

### Problem
A customer has a firmware communication issue where the AmpliPi preamp board stops responding to zone/volume control commands and needs to be reset. This happens to them daily.

### Solution
Add a counter for detecting this failure case X times in a short period. When the failure case is detected attempt to reset the firmware and configure it back to its most recent state

### LINKED ISSUE #1065: I2C File Descriptor Leak Causing "Too Many Open Files" Error
Summary
AmpliPi is leaking I2C file descriptors, causing the application to crash with OSError: [Errno 24] Too many open files after running for several hours.

Environment

AmpliPi Version: 0.4.8
Hardware: Raspberry Pi
Python Version: 3.7
OS: Raspbian/Raspberry Pi OS

Problem Description
The AmpliPi service crashes repeatedly with the following error:
textNov 07 20:43:02 amplipi authbind[851]: OSError: [Errno 24] Too many open files
Nov 07 20:43:02 amplipi authbind[851]: ERROR:asyncio:socket.accept() out of system resource

Root Cause Analysis
Investigation reveals that the application is leaking file descriptors to /dev/i2c-1. The process accumulates over 1000 open file handles to the I2C device, eventually hitting the system limit of 1024.
Diagnostic Output:
```
bash$ lsof -p 851 | grep CHR | awk '{print $NF}' | sort | uniq -c | sort -rn
   1014 /dev/i2c-1
      2 /dev/gpiomem
      1 /dev/null
bash$ cat /proc/851/limits | grep "open files"
Max open files            1024                 1048576              files
bash$ ls -l /proc/851/fd | wc -l
1025
```
The process has 1025 open file descriptors with 1014 of them being handles to /dev/i2c-1.

Expected vs Actual Behavior
Expected: The I2C bus should be opened once and reused, OR file handles should be properly closed after each I2C operation using context managers.
Actual: Each I2C operation opens a new file handle to /dev/i2c-1 that is never closed, causing file descriptor exhaustion over time.

Likely Code Issue
The problem is likely in code that accesses I2C devices (preamps, audio chips, etc.). The code probably looks something like this:
```
:x: Bad (current):
pythonbus = smbus.SMBus(1)  # Opens /dev/i2c-1
bus.write_byte_data(addr, reg, value)
```
# Never closes the file descriptor!
```
:white_check_mark: Good (should be):
python# Option 1: Use context manager
with smbus.SMBus(1) as bus:
    bus.write_byte_data(addr, reg, value)
```
# Automatically closes

```
# Option 2: Open once, reuse everywhere
class AudioController:
    def __init__(self):
        self.bus = smbus.SMBus(1)  # Open once
    
    def set_volume(self, addr, value):
        self.bus.write_byte_data(addr, reg, value)  # Reuse
    
    def __del__(self):
        self.bus.close()  # Clean up
```

```
Workaround (Temporary Fix)
Increase the file descriptor limit to prevent immediate crashes:
bashsystemctl --user edit amplipi.service --full
Add under the [Service] section:
iniLimitNOFILE=8192
Then reload and restart:
bashsystemctl --user daemon-reload
systemctl --user restart amplipi.service
```

:warning: Note: This is just a workaround that buys time. The underlying code leak still needs to be fixed.

Suggested Fix for Developers

Search the codebase for all smbus.SMBus() or similar I2C access calls
Choose one approach:

Use context managers (with statements) to ensure handles are closed
Open the I2C bus once at application startup and reuse the same handle
Explicitly call .close() on I2C handles when done

Consider using a singleton pattern or dependency injection to share a single I2C bus instance across the application

Additional Notes
This issue will affect any AmpliPi installation running continuously. The time to failure depends on how frequently I2C operations occur (volume changes, input switching, preamp adjustments, etc.).
Happy to provide additional diagnostic information or logs if needed!

SteveMicroNova:
Surface backend errors to users better

Also mitigate some minor issues that required surfacing

### Checklist

### DIFF
--- a/amplipi/app.py
+++ b/amplipi/app.py
@@ -657,6 +657,12 @@ def debug() -> models.DebugResponse:
     logging.exception("couldn't load debug file: {e}")
     return models.DebugResponse()
 
+
+@api.patch("/api/info/alerts/hide", response_model=List[models.Alert])
+def hide_alert(alert: models.Alert):
+  """Hide an Alert based on the Alert's message"""
+  return utils.hide_alert(message=alert.message)
+
 # include all routes above
 
 
--- a/amplipi/models.py
+++ b/amplipi/models.py
@@ -25,6 +25,7 @@ from typing import List, Dict, Optional, Union, Set
 from types import SimpleNamespace
 from enum import Enum
 from pathlib import Path
+import datetime
 
 # pylint: disable=no-name-in-module
 from pydantic import BaseSettings, BaseModel, Field
@@ -1013,6 +1014,43 @@ class FirmwareInfo(BaseModel):
   git_dirty: bool = Field(default=False, description="True if local changes were made. Used for development.")
 
 
+class AlertLevel(Enum):
+  """What color should the alert be as per the Mui style guide: https://mui.com/material-ui/react-alert/#severity"""
+  WARNING = "warning"
+  ERROR = "error"
+  INFO = "info"
+  SUCCESS = "success"
+
+
+class Alert(BaseModel):
+  message: str
+  severity: AlertLevel = AlertLevel.ERROR
+  """What color should the alert be as per the Mui style guide: https://mui.com/material-ui/react-alert/#severity"""
+  hidden: bool = False
+  """Has this Alert been hidden by the user?"""
+  timestamp: datetime.datetime = Field(
+    default_factory=lambda: datetime.datetime.now(datetime.timezone.utc)
+  )
+
+  @property
+  def expired(self) -> bool:  # Used to limit alerts to have only a single instance per week. If the state that caused the alert is still valid after a week, the same alert will be made.
+    return (datetime.datetime.now(datetime.timezone.utc) - self.timestamp) > datetime.timedelta(weeks=1)
+
+  class Config:
+    schema_extra = {
+      'examples': {
+        'Example Alert': {
+          'value': {
+            "message": "Writing data to the I2C bus has failed multiple times, please contact AmpliPi Support at mailto:support@micro-nova.com",
+            "severity": "error",
+            "hidden": False,
+            "timestamp": "2026-05-26T19:28:57.907099+00:00"
+          }
+        },
+      }
+    }
+
+
 class Info(BaseModel):
   """ AmpliPi System information """
   version: str = Field(description="software version")
@@ -1033,6 +1071,7 @@ class Info(BaseModel):
     default=[], description='The stream types available on this particular appliance')
   extra_fields: Optional[Dict] = Field(default=None, description='Optional fields for customization')
   connected_drives: List[str] = Field(default=[], description='A list of all external drives connected')
+  global_alerts: List[Alert] = Field(default=[], description='A list of alerts to be shown to all users via the frontend global alert bar')
 
   class Config:
     schema_extra = {
--- a/amplipi/rt.py
+++ b/amplipi/rt.py
@@ -82,6 +82,13 @@ _DEV_ADDRS = [0x08, 0x10, 0x18, 0x20, 0x28, 0x30]
 
 MAX_ZONES = 6 * len(_DEV_ADDRS)
 
+_I2C_ALERT_MSG = (
+  "Writing data to the I2C bus has failed and automatic recovery was unsuccessful. "
+  "Please go to Settings -> Config -> Hardware Reset.\n"
+  "If you see this message again in a short time period, "
+  "contact AmpliPi Support at support@micro-nova.com"
+)
+
 
 class FanCtrl(Enum):
   MAX6644 = 0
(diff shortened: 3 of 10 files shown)


amplipi/app.py
```diff
@@ -635,6 +635,11 @@ def play_media(media: models.PlayMedia, ctrl: Api = Depends(get_ctrl)) -> models
     raise HTTPException(404, f'source id not found')
   return code_response(ctrl, ctrl.play_media(media))
 
+
+@api.patch("/api/alert/hide")
```
SteveMicroNova: This pathing might need some work. It made more sense when I had multiple alert endpoints being added here but I found 2/3 of the endpoints didn't make sense as I kept devving on this

amplipi/models.py
```diff
@@ -1004,6 +1005,28 @@ class FirmwareInfo(BaseModel):
...
+  WARNING = "warning"
+  ERROR = "error"
+  INFO = "info"
+  SUCCESS = "success"
+
+
+class Alert(BaseModel):
+  message: str
+  severity: AlertLevel = AlertLevel.ERROR
+  """What color should the alert be as per the Mui style guide: https://mui.com/material-ui/react-alert/#severity"""
```
SteveMicroNova, line 1028: The docstring under the severity var would probably also make sense as a docstring for the AlertLevel

amplipi/utils.py
```diff
@@ -495,3 +495,64 @@ def clear_custom_configs():
...
+      item for item in alerts
+      if item.message == alert.message and not item.expired
+    ),
+    None
+  )
+  if selected_alert is not None:
+    selected_alert.hidden = True
+    save_alerts(alerts)
+  else:
+    add_alert("Alert not found, could not be hidden!")
```
SteveMicroNova, line 566: This probably makes sense as an Exception as well

web/src/components/StatusBars/AlertBar.jsx
SteveMicroNova, line 1: The MUI alert component that we use has a "severity" property that has 4 possible values - error, success, warning, and info. I have had no particular use for warning and info thus far, so I flattened it into just success or error. Within that, I generally only needed error so I had it be a `store_true` arg for whether it shows success or fail, I've upgraded that with full granularity to future proof these changes.

Luckily the other two AlertBar style components we use don't have the same controls so I didn't need to upkeep those for full feature parity in this PR

tests/test_rest.py
```diff
@@ -1462,18 +1462,18 @@ def test_api_doc_has_examples(client):
...
         resp_spec = m['responses']['200']['content']['application/json']
-        assert 'example' in resp_spec or 'examples' in resp_spec, f'{path_desc}: At least one exmaple response required'
-        if 'exmaples' in resp_spec:
-          assert len(resp_spec['examples']) > 0, f'{path_desc}: At least one exmaple response required'
+        assert 'example' in resp_spec or 'examples' in resp_spec, f'{path_desc}: At least one example response required'
+        if 'examples' in resp_spec:
+          assert len(resp_spec['examples']) > 0, f'{path_desc}: At least one example response required'
       except KeyError:
-        pass  # reposnse could not be json
+        pass  # response could not be json
```
SteveMicroNova, line 1500: This got hit by a spellcheck, I don't think that anything has been hurt by this (certainly nothing that's covered by tests) but given the consistency of the `exmaple` spelling, particularly in functional code as string comparisons, I wonder if that was intentional

amplipi/rt.py
```diff
@@ -262,8 +264,13 @@ def write_byte_data(self, preamp_addr, reg, data):
       try:
         time.sleep(0.001)  # space out sequential calls to avoid bus errors
         self.bus.write_byte_data(preamp_addr, reg, data)
-      except Exception:
+      except Exception as e:
+        logger.exception(f"Writing preamp failed: {e}")
         time.sleep(0.001)
+        self.bus.close()
+        self.write_byte_data_failures += 1
+        if self.write_byte_data_failures >= 3:
```
linknum23: This has the potential to generate a continuous stream of alerts on failure. How should this handle repeat failures?

SteveMicroNova: Currently I handle repeated errors by giving them an expiration time and a flow further in the PR that says "There can only be one non-expired version of this message at a time", if you try to add the same alert again but there's a fresh alert with the same text already in the file then it will not send

Currently expiration time is set to 1 week

linknum23: That might be a useful comment to add to the docstring of add_alert. I missed that on the first pass.

amplipi/utils.py
```diff
@@ -495,3 +495,80 @@ def clear_custom_configs():
       os.remove(path)
     except Exception as e:
       logger.exception(f"failed to clear device configuration: {e}")
+
+
+# Every alert(s) function was in ctrl.py, but due to many files needing access to the add_alert flow they had to be here in utils
+
+
+def load_alerts() -> List[models.Alert]:
```
linknum23, line 503: You may want to add some doc strings to each of these public functions

amplipi/app.py
linknum23: Having only a single endpoint for this is useful since it reduces the scope on this. I wish we had a place for odds and ends like this that don't focus on the actual functionality of the unit.

tests/test_rest.py
linknum23: This is fine. I just suck at typing, LoL

amplipi/app.py
linknum23: Since it is only exposed via api/info does it make sense to be there?

Maybe `/api/info/hide-alerts` or `/api/info/alerts/hide` ?

linknum23: I'm not sold on any of those options tho.

SteveMicroNova: > I wish we had a place for odds and ends like this that don't focus on the actual functionality of the unit.

It would be trivial to add a `api-utils.py` or something similar that uses the same FastAPI instance without having it in the same file as the primary functionality if that's what you mean

Also for the routing opinions, I like `/api/info/alerts/hide` because I do see a world where we might want a `/api/info/alerts/add` endpoint, the main case being if the frontend finds an error that should become global to all users, or a world where we might want `/api/info/alerts` as a GET request to return just that portion of the info block (this could be fairly useful to raise these issues to home assistant's notification scheme, for example)

SteveMicroNova:
To record a conversation we've had in the office:
We'd like to run this for a few days and see if we trigger this message on anything. It seems fine so far, but directing people to email us when they don't really need to could be needlessly annoying for everyone

SteveMicroNova:
After 24 hours of testing, this is looking good. I'm going to keep my automated testing script going while I'm on vacation next week and I suspect that once I'm back on the 15th this will be good to merge.

SteveMicroNova:
It is now the 15th, and my minute-by-minute script has yet to trip the sensor on this one. We ought to be fine for standard user day-to-day operation without having a big threatening error screen.

amplipi/rt.py
```diff
@@ -269,8 +269,8 @@ def write_byte_data(self, preamp_addr, reg, data):
         time.sleep(0.001)
         self.bus.close()
         self.write_byte_data_failures += 1
-        if self.write_byte_data_failures >= 3:
-          utils.add_alert("Writing data to the I2C bus has failed multiple times, please contact AmpliPi Support at support@micro-nova.com")
+        if self.write_byte_data_failures >= 1:
```
linknum23: I would still leave this at a count of 3 just in case and keep the updated message.

linknum23 (approved):

web/src/App.jsx
```diff
@@ -270,7 +270,6 @@ Page.propTypes = {
 const App = ({ selectedPage }) => {
     const alert = useStatusStore((s) => s.alert);
     return (
-        <div className="app">
```
linknum23: Shouldn't this be "squished" into the previous commit, where the error was created?

Issue #957, opened 2024-10-01
Spotify controlled loopback getting into a bad state


linknum23:
We've had several users report an issue with Spotify streams that stop working unexpectedly due to an ALSA problem.

Here's what a typical logs look like:

```
Sep 23 13:50:53 amplipi-inhouse authbind[776]: ALSA lib pcm_dmix.c:1165:(snd_pcm_dmix_open) unable to open slave
Sep 23 13:50:53 amplipi-inhouse authbind[776]: Audio Sink Error Connection Refused: <AlsaSink> Device lb4c May be Invalid, Busy, or Already in Use, ALSA function 'snd_pcm_open' failed with error 'EINVAL: Invalid argument'
``` 

This smells like a previous stream not getting closed properly.

linknum23:
This may be exacerbated by the #961

PR #1095, opened 2026-06-10
Fix duplicate dmix ipc_key collisions in asound.conf

stamateviorel:

All twelve loopback dmix devices in `asound.conf` share `ipc_key 1028`. dmix instances with the same ipc_key collide on one shared-memory segment, so opening a loopback playback device intermittently fails with `EINVAL` / `snd_pcm_dmix_open: unable to open slave` depending on open order and which zones are active. The player on that source then crash-loops: on our production unit the LMS player flapped 189 times/day until we made the keys unique. The log signature is identical to #957 (`lb4c ... unable to open slave ... EINVAL`), so I believe this is that issue's root cause. The patch gives every loopback dmix a unique ipc_key and pins the 8ch dmixer slave rate to 48000 so its format can't drift from the loopbacks.

Running in production on a stock 0.4.10 unit since 2026-06-08 with zero loopback open failures since (previously several per day).

### Checklist

* [ ] Does your submission pass linting & tests? (config-only change; happy to fix anything CI flags)

### DIFF
--- a/config/asound.conf
+++ b/config/asound.conf
@@ -71,6 +71,7 @@ pcm.dmixer {
         period_size 1024
         buffer_size 4096
         channels    8
+        rate        48000
     }
     bindings {
         0 0
@@ -249,7 +250,7 @@ pcm.lb0c {
 }
 pcm.lb1 {
     type dmix
-    ipc_key         1028
+    ipc_key         1029
     ipc_perm        0666
     slave {
         pcm         "hw:Loopback_1,0"
@@ -266,7 +267,7 @@ pcm.lb1c {
 
 pcm.lb2 {
     type dmix
-    ipc_key         1028
+    ipc_key         1030
     ipc_perm        0666
     slave {
         pcm         "hw:Loopback_2,0"
@@ -282,7 +283,7 @@ pcm.lb2c {
 }
 pcm.lb3 {
     type dmix
-    ipc_key         1028
+    ipc_key         1031
     ipc_perm        0666
     slave {
         pcm         "hw:Loopback_3,0"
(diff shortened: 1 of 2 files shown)


config/asound.conf
```diff
@@ -71,6 +71,7 @@ pcm.dmixer {
         period_size 1024
         buffer_size 4096
         channels    8
+        rate        48000
```
linknum23, line 74: Is setting this fixed rate required? I know this is the default rate of the device but I'm not sure what the implications of fixing this rate rate.

config/asound.conf
```diff
@@ -249,7 +250,7 @@ pcm.lb0c {
 }
 pcm.lb1 {
     type dmix
-    ipc_key         1028
+    ipc_key         1029
```
linknum23, line 253: Nice catch!

config/asound.conf
stamateviorel: Not strictly required, and happy to drop it. It's defensive rather than a fix: without an explicit rate, the first client to open the dmix sets the slave rate for every other loopback, so a stream that opens at 44100 would reconfigure the shared buffer and force resampling on everything else. Pinning it to 48000 (the DAC's native rate, which is also dmix's default here) just makes that explicit so no single client can move it.

That said, it's a separate concern from the ipc_key collision this PR is really about. If you'd rather keep the change minimal, I'll pull the `rate` line out and leave just the unique-key fix.

SteveMicroNova (approved):
This looks good, going to approve this and stamp it into a beta release for field testing

SteveMicroNova:
> This looks good, going to approve this and stamp it into a beta release for field testing

It seems this is either going to require you to resync your fork and rebase this branch, or I'll have to make my own branch where I cherry pick this commit and remake the PR

Would you mind attempting the resync and rebase?

PR #1098, opened 2026-06-10
Auto-recover a hung preamp on persistent I2C write failures

stamateviorel:

When the preamp microcontroller hangs it stops ACKing and every I2C write fails with `OSError 121` (EREMOTEIO). The existing fallback in `_Preamps.write_byte_data` only reopens the SMBus handle, which recovers a transient bus glitch but not a hung preamp — zone control stays dead until someone power-cycles the unit. This escalates the fallback: when the reopened-bus retry also fails, reset the preamps in place, re-assign I2C addresses, reopen the bus, re-flush all cached register values (so zone mute/source/volume state survives the reset), then retry the write. Rate-limited to once per 20 s so a benign one-off glitch never resets audio.

We hit this live on 2026-06-04 (all zone control dead, reads OK / writes failing with Errno 121, only a reboot helped). With this patch deployed the same wedge self-heals in under a second.

### Checklist

* [x] Have you tested your changes and ensured they work? (in production on a real AmpliPi since 2026-06-04)
* [ ] Does your submission pass linting & tests? (`python -m py_compile` clean; happy to fix anything CI flags)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   * Upgraded volume calculations to preserve relative positions when hitting the min or max setting via source volume bar
   * Update our spotify provider `go-librespot` to `0.7.3`
   * Upgrade from Logitech Media Server 8.5.2 to Lyrion Music Server 9.0.3
+  * Added in-place preamp recovery when I2C writes fail persistently (EREMOTEIO), avoiding a full reboot
 
 # 0.4.11
 * System
--- a/amplipi/rt.py
+++ b/amplipi/rt.py
@@ -145,6 +145,11 @@ class _Preamps:
 
   preamps: Dict[int, List[int]]  # Key: i2c address, Val: register values
 
+  # In-place preamp recovery — rate-limited so a benign I2C glitch
+  # never resets audio. See _recover_preamps() / write_byte_data().
+  _RECOVERY_COOLDOWN_S = 20.0
+  _last_recovery = 0.0
+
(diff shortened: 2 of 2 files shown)


linknum23:
We probably need to merge this code with recent changes made with a similar intent.

stamateviorel:
Makes sense — I dug into `ResetOnFailure` and I think these two layer together nicely rather than overlap.

Your branch detects the failure and surfaces it to the user (the failure counter + alert pointing at Settings -> Config -> Hardware Reset), which is exactly the right UX when the hardware genuinely needs attention. This PR sits one level below that: on the failure it automatically pulses `reset_preamps()` and re-flushes the cached register state so zone mute/source/volume survive, then retries — so the common transient wedge self-heals in well under a second without anyone having to touch the unit. Same `reset_preamps()` primitive your manual Hardware Reset uses, just triggered automatically.

The clean combination: auto-recover first; if the auto-recovery itself keeps failing, *that's* when your alert should fire — i.e. "we tried to self-heal and couldn't, this one really does need hardware attention." That's a stronger signal than alerting on the first I2C error, which is often transient.

Happy to rebase this onto `ResetOnFailure` so it merges into the same `write_byte_data` path with no conflict, and wire the escalation into your `add_alert` flow. Want me to put up that combined version?

SteveMicroNova:
Thanks for the work here, stamateviorel!

I've cherrypicked this and brought it in line with a similar thing I was already working on in #1091, so this is now in AmpliPi. Due to this already being merged elsewhere, I'm closing this PR.

PR #1099, opened 2026-06-10
Disable USB autosuspend for the CM6206 USB audio device

stamateviorel:

The CM6206 USB chip provides the digital outputs for sources 1–3; with kernel USB autosuspend active it can power down between uses and glitch on resume. This pins `power/control` to `on` from the same udev rule that already names the card.

### Checklist


SteveMicroNova:
This is another PR I'm happy to merge and interested in testing with a beta release by end of week, but I suspect it'll have the same [issue](https://github.com/micro-nova/AmpliPi/pull/1095#issuecomment-4793527467) as #1095. Would you mind doing the same resync and rebase with this branch and I'll push it through?

SteveMicroNova:
Would you mind rebasing this one more time? The conflict is simple, but as I've discovered I have no rights to correct it.

SteveMicroNova (approved):
This looks good, thank you for your two interventions to make this mergable!

PR #1100, opened 2026-06-19
Fix script permissions by adding missing executable bit

Lohrer:

Fix our ansible runs and regain my sanity.

When 'installing' AmpliPi  by running `scripts/configure.py`, all files in the `scripts` directory are made executable with a `chmod +x`. Since this change shows up in git, every time our auto-deploy to the testing workstations run, it thinks:
1. There is a change to the git checkout, so clean the repo.
2. Since there was a change to the local repo, re-install AmpliPi.

Not only is it confusing to see that AmpliPi "changed" when it didn't, installing AmpliPi is often the longest step in all of the ansible actions, and it has to do it to two workstations. The installation step does not occur if no changes have been made to the repo.

The .service files shouldn't need to  be made executable, but for some reason they're in the scripts directory so they're already getting lumped in with the `+x` so I won't change anything there in this PR. Just noting that it seems odd.

### Checklist

* ~[ ] If applicable, have you updated the documentation/manual?~
* ~[ ] If applicable, have you updated the CHANGELOG?~
* ~[ ] Have you written new tests for your core features/changes, as applicable?~
* ~[ ] If this is a UI change, have you tested it across multiple browser platforms on both desktop and mobile?~


linknum23:
Sorry about this. IIRC I think windows had a hand in changing some of these.

linknum23 (approved):

PR #1102, opened 2026-06-25
Fix duplicate dmix ipc_key collisions in asound.conf

SteveMicroNova:

All twelve loopback dmix devices in `asound.conf` share `ipc_key 1028`. dmix instances with the same ipc_key collide on one shared-memory segment, so opening a loopback playback device intermittently fails with `EINVAL` / `snd_pcm_dmix_open: unable to open slave` depending on open order and which zones are active. The player on that source then crash-loops: on our production unit the LMS player flapped 189 times/day until we made the keys unique. The log signature is identical to #957 (`lb4c ... unable to open slave ... EINVAL`), so I believe this is that issue's root cause. The patch gives every loopback dmix a unique ipc_key and pins the 8ch dmixer slave rate to 48000 so its format can't drift from the loopbacks.

Running in production on a stock 0.4.10 unit since 2026-06-08 with zero loopback open failures since (previously several per day).

### Checklist

* [ ] Does your submission pass linting & tests? (config-only change; happy to fix anything CI flags)

NOTE: This was originally pr #1095, I just screwed up the git history and didn't have the rights to fix it in that PR due to not owning the branch origin. Thank you to stamateviorel for this work!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   * Update our spotify provider `go-librespot` to `0.7.3`
   * Upgrade from Logitech Media Server 8.5.2 to Lyrion Music Server 9.0.3
   * Added in-place preamp recovery when I2C writes fail persistently
+  * Fixed all loopback dmix devices sharing one ipc_key in asound.conf, which made loopback playback devices intermittently fail to open with EINVAL (#957)
 * Web App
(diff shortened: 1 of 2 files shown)


SteveMicroNova:
This may no longer be necessary as stamateviorel has just rebased their PR #1095, once tests are completed this is liable to be deleted

PR #1105, opened 2026-07-15
fix(web): preserve player metadata width

### LINKED ISSUE #1103: Player Page Album Art Size
The width of the metadata section on the player page is decided by the size of the album art. If album art fails to load or isn't provided, this makes it so that the rest of the metadata (song name, album, artist) gets scrunched up.

Set a minimum height and width of about 350 px on desktop and 98 vw on mobile

Evidence of the issue, sometimes characterized by the marquees being activated erroneously as well but not in this case:

Bad:
Image

Good:
Image

Approximation of a solution:
Image

hkJerryLeung:
## What does this change intend to accomplish?

Fixes #1103.

When album art is missing or fails to load, the image has no intrinsic size and the player metadata column collapses. This change reserves a square album-art area so song, album, and artist metadata retain a stable width:

- desktop: 22rem minimum and 352px maximum
- narrow screens: 98vw minimum and maximum
- loaded artwork remains contained without distortion

The metadata Grid now uses a class instead of an inline max-width so the responsive rules live in the stylesheet.

## Validation

- npm ci
- npm run build
- git diff --check

The build passed. npm ci reported the repository's existing dependency audit findings (1 low, 9 moderate, 11 high). Cross-browser manual testing was not available in this environment.

### DIFF
--- a/web/src/pages/Player/Player.jsx
+++ b/web/src/pages/Player/Player.jsx
@@ -98,7 +98,7 @@ const Player = () => {
                         </Chip>
                     </div>
                 </Grid>
-                <Grid item xs={2} sm={4} md={4} style={{maxWidth: "22rem"}}>
+                <Grid item xs={2} sm={4} md={4} className="player-metadata-container">
                     <Box
                         className="album-art-container"
                         sx={{
--- a/web/src/pages/Player/Player.scss
+++ b/web/src/pages/Player/Player.scss
@@ -26,15 +26,23 @@
   padding-right: 2.5vw;
 }
 
+.player-metadata-container {
+  width: 22rem;
+}
+
 .player-album-art {
   align-self: center;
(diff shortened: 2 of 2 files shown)


SteveMicroNova:
I haven't got a good environment to test this at the moment but based on how similar this looks to #1104, I assume the failure points of #1104 still stand

There should be a few chunks of CSS spread out in various places that actually overwrite what you've done here, see my comments on that other PR to see what those CSS chunks do to overwrite the classes you've made


Note that this is based purely on looking at the code and not from actually seeing an issue, so don't feel like you need to dig too deep on this until I have time to test properly. If you'd like to test in lieu of having a unit to test on, we offer scripts such as run_debug_backend and run_debug_frontend in our scripts folder that can be run at the same time to simulate a full AmpliPi unit for testing.

PR #1106, opened 2026-07-20
Raspbian trixie

### LINKED ISSUE #335: Migrate to 64-bit Raspberry Pi OS

### LINKED ISSUE #599: Update updater to flash full images

### LINKED ISSUE #784: Robust TRY_BOOT script
* validate that a recently flashed partition is correct, and set TRY_BOOT correctly
* reboot if failed update

SteveMicroNova:
This is a reworking of AmpliPi to function in a 64 bit raspbian trixie [A:B tryboot](https://www.raspberrypi.com/documentation/computers/config_txt.html) setup with the goal of having a more robust system overall and a cleaner update flow with full image updates

Keep an eye on the docs/a_b_update_design.md file for a good rundown of the high level design and purpose of the new components within this PR

A few things that need to be worked on yet:
- [x] Ensure that LMS config and plugins persist on slot-swap
- [ ] Label the partitions something other than p1-7
- [x] Where should image files be hosted?

### Checklist


SteveMicroNova:
EDIT: This comment has been expanded into a proper markdown file in this pr, see [a_b_design.md](https://github.com/micro-nova/AmpliPi/pull/1106/changes#diff-b35d2ed803dcb7acff59dba9396df3a0c365cb482ccf2a5d30548033e9a0bf26) for the more up-to-date version

This is a WIP that will take a while to finish

This will change the baseline partition scheme of amplipi from basic boot+root to a 7 partition schema:
- `p1`: Autoboot (raspberry pi A:B boot necessity), mounted as `/boot/firmware`
- `p2`: Boot partition A
- `p3`: Boot partition B
- `p4`: Extended partition containing partitions 5, 6, and 7
- `p5`: Root partition A
- `p6`: Root partition B
- `p7`: Persistent data storage for things that need to remain during flash, such as the user's config. Mounted at `/data`

Currently, I have the ability to update from images via the `updater/asgi.py` `/update/flash` endpoint that takes an update package: a folder containing a root image, an optional boot image, and a `manifest.json` file that describes the update package with a bytesize and sha256 checksum of the images.
This endpoint uses that update package to first verify the checksum of the image(s) before flashing, and then flashes the inactive directory. After flashing is complete, it also writes an "update-pending" file to p1 that signals to the update finalization service that things are freshly flashed and need to be checked on reboot.
Once that's all done, it also runs `sudo reboot "0 tryboot"` to move to the other slot


Once in the other boot+root slot, the update verification service checks to see if that `/boot/firmware/update-pending` file exists and if yes, runs a preflight check before allowing amplipi to start up - verifying if all the apt and python packages are properly in place. Once that's done and amplipi runs, it also edits `/boot/firmware/autoboot.txt` to change the newly flashed boot slot to be the new default slot so that the user doesn't revert to the previous slot on next reboot

amplipi/display/check_pass
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
```
SteveMicroNova, line 1: The version of python we run has to be installed via UV on Trixie, so I had to move the invocation paths on a few files

amplipi/display/common.py
```diff
@@ -75,7 +75,7 @@ class DefaultPass:
      the stored default AmpliPi password."""
 
   # Password config location
-  PASS_DIR = os.path.join(os.path.expanduser('~'), '.config', 'amplipi')
```
SteveMicroNova, line 78: The whole `.config` folder and a few services got moved to `/data` to ensure they persist post-update, many lines like this had to be changed as a result

config/autoboot.txt
```diff
@@ -0,0 +1,13 @@
...
+
+[tryboot]
+boot_partition=3
+
+if exist mmc 0:1 /boot_partition.txt then
+  load mmc 0:1 ${loadaddr} /boot_partition.txt
+  env import -t ${loadaddr} ${filesize}
+fi
+
+boot mmc 0:${boot_partition}
```
SteveMicroNova: This is the default autoboot.txt, which is what the update verification service edits when an update is successful. Specifically what is changed is swapping the [all] boot_partition value with the [tryboot] boot_partition value, that's all it takes to swap the default boot slot

docs/imaging_etcher.md
```diff
@@ -5,10 +5,6 @@
   * A micro USB cable
   * Your AmpliPi
 
-## Optional Step: Preserve your config
```
SteveMicroNova, line 8: This file likely needs more reworking as realistically you shouldn't need to flash ever again, though also I'd say that makes more sense to leave until this code is in the hands of users and has a few support cases under its belt to verify if that's necessary

scripts/amplipi-postflash.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Run device-specific customization scripts from p7.
+# These survive OTA updates and are re-applied on each new slot's first boot.
+if [ -d "${SCRIPTS_DIR}" ]; then
+  for script in "${SCRIPTS_DIR}"/*.sh; do
+    [ -f "${script}" ] || continue
+    log "Running ${script}..."
+    bash "${script}" || log "Warning: ${script} exited non-zero"
+  done
+fi
```
SteveMicroNova, line 41: Some users have specific setups on their devices that we don't want to destroy on a full-image update, so I've given them a directory that they can fill with bash scripts to set up their services automatically post-update

I need to test what happens if a user provides scripts that fail for those cases

scripts/bootstrap_pi
```diff
@@ -9,10 +9,10 @@ Bootstrap a Raspberry Pi Compute Module for first-time AmpliPi setup.
   --hostname NAME: The hostname to set on the pi, so once this script
                    successfully completes connect via {hostname}.local.
                    Default is amplipi
-  --cli-only:      Use the lite version of Raspberry Pi OS which does not have
```
SteveMicroNova, line 12: At some point, CLI only became a default case for our units. I decided to flip the default case in this script to reflect that

scripts/bootstrap_pi
SteveMicroNova, line 1: This file either needs full reworking or deletion. The new 7 partition setup is complex enough that updating this is a bit of a lift when we hardly need it and can also just distribute a full 7 partition image for those who need that sort of thing

scripts/configure.py
SteveMicroNova, line 1: Unlike bootstrap.py, I'd say configure.py is still very useful despite not being our update path anymore. manual deploys with the deploy+configure script are likely to be the primary ways we configure the root partitions that later get snapshotted into an image we distribute as an update

amplipi/ctrl.py
```diff
@@ -852,15 +852,15 @@ def set_vol():
               vol, vol_f, vol_f_delta, vol_min, or vol_max.
           """
           # Field precedence: vol (db) > vol_delta > vol (float)
-          # vol (db) is first in precedence yet last in the stack to cover the default case of no volume change
```
SteveMicroNova, line 855: This looks like a rebase error. This branch is long lived and had some merge conflicts on rebase and so may need to be combed for reversions like this before merging

amplipi/rt.py
```diff
@@ -283,6 +283,10 @@ def _recover_preamps(self) -> bool:
     try:
       self.reset_preamps()
       self.set_i2c_addr()
+      try:
+        self.bus.close()
+      except Exception:
+        pass
```
SteveMicroNova, line 289: This is an additional fix for the issue fixed by #1098 which was merged in #1091 

My unit actually got into a bad state during testing, which led to me having the ability to test those changes and add a little more to the fix

My own issue was downstream of some configuration file providing an environment that the I2C didn't like

scripts/bootstrap_pi
```diff
@@ -297,5 +306,32 @@ Bootstrapping successful. Please complete the following steps:
...
+  4. Edit the contents of the partitions as so:
+
+    p1: add autoboot.txt as seen in /config/autoboot.txt
+    p2: adjust cmdline.txt to point to partition 5
+    p3: adjust cmdline.txt to point to partition 6
+
+    p5: Add p7 to fstab as /data
+    p6: Add p7 to fstab as /data
+    p7: create a .config folder
 "
```
SteveMicroNova, line 336: This is useful documentation for someone who might be trying to get their bearings on this PR

SteveMicroNova: Also worth noting that this section isn't useful as a guide to get a unit up and running anymore

As I've continued working I've found more specific details that simply will not be captured here due to being things like "go deep into this directory and change this line in a random config file that I don't remember anymore"

config/autoboot.txt
SteveMicroNova, line 1: I've just considered that this file needs a greater explanation

The [official docs](https://www.raspberrypi.com/documentation/computers/config_txt.html) are a bit sparse (or were when I first read them months ago), so I'll also explain myself:

`tryboot_a_b=1` activates the tryboot mechanism, something supported in the raspbian kernel that allows for partition swapping using the command `sudo reboot "0 tryboot"`. Using specifically that `"0 tryboot"` arg is how you boot into the partition listed under the `[tryboot] boot_partition` section.

That `"0 tryboot"` arg simply loads the boot partition listed under the `[tryboot]` header, it does not toggle anything in this file to make that partition the new default (which is handled by the amplipi-postflash service I've made), so doing a simple `sudo reboot` from the tryboot partition will send you back to whatever partition is listed under the `[all]` header

All the actual lifting is done by the individual boot partitions, which have their own kernels and mappings to their own root partitions

config/autoboot.txt
SteveMicroNova: This specific section may not be useful anymore, I'll have to do some testing without it

This was from a time when I was trying to have the boot_partitions be swapped without having a dedicated service for that purpose

scripts/bootstrap_pi
SteveMicroNova: The answer is reworking. It won't be a simple portion, but creating a clean image that isn't mucked up with any dev happenings is an important goal and this script is likely one of the better ways to achieve that

amplipi/display/check_pass
linknum23: Should we replace this with something more generic  like uv run?

amplipi/updater/static/index.html
```diff
@@ -277,6 +277,27 @@ <h6 class="card-header">
...
+                <small class="update-progress-label d-none" data-label="Downloading boot image" style="width: 0%;">Download Boot</small>
+                <small class="update-progress-label d-none" data-label="Verifying root image" style="width: 0%;">Verify Root</small>
+                <small class="update-progress-label d-none" data-label="Verifying boot image" style="width: 0%;">Verify Boot</small>
+                <small class="update-progress-label d-none" data-label="Flashing root" style="width: 0%;">Flash Root</small>
+                <small class="update-progress-label d-none" data-label="Flashing boot" style="width: 0%;">Flash Boot</small>
+                <small class="update-progress-label d-none" data-label="Downloading manifest" style="width: 0%;">Manifest</small>
+                <small class="update-progress-label d-none" data-label="Downloading release" style="width: 0%;">Download</small>
+                <small class="update-progress-label d-none" data-label="Applying update" style="width: 0%;">Apply</small>
+              </div>
```
SteveMicroNova, line 299: I added a weighted progress bar during these changes to help make the updater feel more responsive and less like a black box of sorts

This is possible now that we're dealing with files of known size, and the system knows how far along the flashing process it is.

Here's an example of what it looks like:

Image

amplipi/updater/asgi.py
```diff
@@ -65,6 +73,83 @@
 sse_messages: queue.Queue = queue.Queue()
 
 
+class SSEChannel:
+  """ Bundles the queue/in-progress-flag/latest-status trio a long-running background job reports
+  its progress through, plus the polling generator that serves it as SSE to a native EventSource.
+  /update/flash and /update/download/images each run their own job in a background thread,
+  separate from the SSE connection itself, so a dropped/reconnecting browser can't interrupt or
+  duplicate the actual work - progress is watched via a paired GET .../progress endpoint instead. """
```
SteveMicroNova, line 81: I needed to ensure that there could only be one update running at a time (a problem with our current updater, but even more important to handle here) and thus I had to make a progress tracking channel.
While doing so, I also had the idea of having a visible progress bar on the frontend that this now also feeds

docs/a_b_design.md
SteveMicroNova, line 1: This is probably going to be the most useful file during PR review, keeping this up to date will prove easier than keeping the actual PR description up to date especially since I can easily link to specific files from this

